Issue #999, opened 2026-08-25
Strategist role: periodic archive-wide review writing HADI hypotheses and futility advisories through existing channels


ozand:
## Context (self-contained)

The loop has tactical decision-making but no strategic layer. Today's roles: the deterministic demand pipeline (priority > defect > goal-gap > hypothesis > decay) turns metrics and scans into work items; the proposer (one stateless 400-token LLM call, #760/#902) refines ONE assigned item into a task; the executor implements it; the curator (#986) distills lessons into the knowledge base daily. Nobody looks at the whole picture: the evolution tree (`state/evolution/tree.json`), the outcome history, the lessons, the curator's knowledge base, and the demand funnel statistics are written but never read by any decision-maker.

Measured consequence (2026-08-19..25): 70% of all proposals chased two goal-gap metrics without moving them; the proposer re-proposed the same themes ~15×/day; nothing exists that could notice a direction is barren, reprioritize across demand kinds, or pose a genuinely new hypothesis. This is the gap the Darwin Gödel Machine fills with archive-conditioned parent selection and the AI Scientist fills with idea generation over prior results: generation conditioned on the archive, not on the last 15 rows.

## Design (binding)

**Role & cadence:** a fourth LLM role, `strategist`, following the knowledge-curator pattern exactly: own systemd oneshot unit + daily timer (offset from the curator, e.g. 03:00 MSK), watermark state under `state/strategist/`, 1–2 bounded LLM calls per run, fail-open (a failed run changes nothing and logs to `errors.jsonl`).

**Model & telemetry:** resolved via `model_registry` with a new `strategist` role (`SELFEVO_STRATEGIST_MODEL` env, preset-configurable, no hardcoded model). Every call recorded through the existing llm_calls recorder with `component: "strategist"`.

**Bounded input (never the whole repo):** (a) a compact digest of the evolution tree (node count, depth, outcome mix, current best path); (b) scorecard metrics with 7-day history; (c) the demand/proposer funnel stats (proposed / integrated / self_dedup / futile per demand id — including the futility sidecar from the companion futility issue); (d) recent `reusable_insight` lines from lessons plus `memory/index.md` and `docs/index.md` (indexes only); (e) the goal charter. Each section separately capped.

**Output — STRICT JSON, advisory-only, through EXISTING channels:**
[...]

PR #1052, opened 2026-08-26
fix(hypotheses): recognize demand-served experiments

ozand:
Follow-up for #999 reviewer finding: normal proposer ledger rows use `serves: demand <hypothesis-id>`. Recognize both demand and hypothesis forms in hypothesis lifecycle/in-flight reconciliation. Added regression test; focused suite passed. No goldens or bridge.py changes.


ozand:
Superseded: current main's _HYPOTHESIS_SERVES_RE already matches both 'hypothesis <ref>' and 'demand <ref>' forms (landed with the #999 durable-hypotheses rework, PR #1054). Nothing left to merge.

Issue #1179, opened 2026-09-02
fix(scorecard): reader_status block, gaps 'unavailable' ≠ 'no gaps', stale_feeds permission vs missing, decide the unprovisioned host_metrics feed


ozand:
Parent ADR: #1173. Depends on #1174 (`Window.status`, `covered_from`). Touches `scorecard.py` and (after #1166) `demand._goal_gap_items`.

## Problem

The scorecard is the runtime's fitness function and the source of `goal-gap` demand — a gap is a work order the proposer spends LLM calls on. Today an unavailable input and a healthy input produce the same output.

- `compute_scorecard`'s outer handler (`scorecard.py:1601-1615`) returns a stub with `"gaps": []` and **no `feeds` key**. `goal_gaps` returns `[]`; `demand._goal_gap_items` reads that as "no measured deficiencies". A completely broken fitness function is indistinguishable from a perfectly healthy one at every consumer.
- `_feeds_section`: an absent feed is `status:"missing", stale:True` whether it was never provisioned or is unreadable (root-owned, #999/#1083 class). `_is_populated` (`:1134-1138`) calls `p.stat()` outside any `try`, so one `OSError` there escapes and zeroes every section via the stub above. `_feed_latest_timestamp` falls back to `st_mtime` for a corrupt JSON feed (`:1107-1110`), so a corrupt feed touched a minute ago reads `"fresh"`.
- **Live**: `host_metrics` has had no writer since **2026-07-12** (`latest_ts: 2026-07-12T22:48:36Z`, `age_seconds: 4411528`). `stale_feeds=1.0` is a V1 rank-6 goal-gap in **all 169** `scorecard/history.jsonl` rows since the 08-28 rotation; `goal-gap-a820ca0c8bb3` (`metric: stale_feeds`) has been tracked in `futility.json` since 2026-08-27 with 1 proposal integrated and `metric_delta: 0.0`. The loop is asked, every cycle, to "Fix or restore stale data feeds: host_metrics" — a feed nobody provisions and the instance repo cannot create.
- `_value_section`: a corrupt `completed.json` ⇒ `completed_declared = 0` ⇒ `min_denominator: 3` suppresses the `confirmed_ratio` gap silently (`:1384`). `_heldout_section`: corrupt `results.json` ⇒ `checked = 0` ⇒ `heldout_gap = None` ⇒ no gap. `_confirmed_cycle_ids` ⇒ `set()` ⇒ every success reported unconfirmed.
[...]

Issue #1175, opened 2026-09-02
fix(runtime): Class-A counters read their true horizon and never persist a count from a partial window (doc-only 24h budget, exhaustion reset, no-op streak, in-flight guard)


ozand:
Parent ADR: #1173. Depends on #1174 (`state_access` module) and on #1166 (pi-2's futility fix — this issue must not touch `demand.py` / `goal_gap_futility.py` until #1166 merges; do the `llm_proposer` and `hypothesis_backlog` parts first if #1166 is still open).

## Problem

Class A decisions ("stop doing X once a count reaches N") count from `state/ledger/cycles.jsonl` only. The ledger rotates at the first append of each UTC day (observed 00:10 UTC; live file afterwards = 3 rows), so every counter resets at midnight and several thresholds are unreachable or wrong for hours.

Live evidence (host, 2026-09-02):

| decision | reader | live evidence |
|---|---|---|
| doc-only 24 h budget (`doc_budget_exceeded = count ≥ 5`, drops every doc-only demand item) | `demand.count_doc_only_integrations_24h`, live file only, docstring "no archive scans" | At 00:30 UTC: doc-only successes in the last 24 h over all files = **5** (= budget, lane should be closed); over the live file = **0** (lane open). `scorecard/latest.json` at 00:14 UTC: `doc_only_integrations_24h: 0`. |
| exhaustion reset on success (`success_ts > exhausted_at`) | `demand._latest_success_ts` over live rows | after rotation no success is visible until the day's first one; `exhausted.json` has 60 `reset` / 7 `exhausted` entries |
| self-dedup / no-op exhaustion (`≥ 2` hides the item) | `_self_dedup_reject_ts_by_demand_id` (ledger half live-only; results half already reads `archive/`) | structural |
| force-proposal after 3 no-op skips | `llm_proposer._consecutive_noop_streak`, live only, **no byte cap** | streak resets at rotation; the escape hatch silently disarms every morning |
| dedup-exhaustion over 3 days | `llm_proposer._dedup_exhausted` | 3-day window from a 1-day file |
| one experiment at a time (`_MAX_HYPOTHESIS_ITEMS = 1`) | `hypothesis_backlog.has_in_flight_experiment`, live only | a `proposed` row that rotated out reads as not-in-flight → a second experiment can be minted (`lifecycle.json`: 52 active) |
[...]

PR #1186, opened 2026-09-02
feat(1174): add bounded state access foundation

### LINKED ISSUE #1174: feat(runtime): state_access — bounded, rotation-aware Window reader for ledger / subagent artifacts / reports / sidecars (ADR #1173 foundation)
Parent ADR: #1173 (bounded state-access contract). This is the **foundation** child: the module, its tests, the live-shape fixture, and the hygiene test. **Behaviour-neutral** — no caller is migrated here; that is #1175/#1176/#1177/#1178/#1179.

## Problem

Every reader of `state/ledger/`, `state/subagents/`, `state/reports/` and the `state/demand/*.json` sidecars re-derives rotation (daily `cycles-YYYY-MM-DD.jsonl.gz`), horizon, ordering and size caps by itself. The ADR audit found eight different answers to "how much rotated ledger do I read" (7 gz / 7 / 2 / 3 / all / all / all-but-500-lines / zero), three copies of `_MAX_LEDGER_BYTES = 2 MiB` with two behaviours plus one uncapped reader, `results/`-only readers that never see `archive/` (3 vs 3004 files on the host), and `max(..., key=st_mtime)` with glob-order tie-breaks (#1168). Comments naming the hazard (`demand.py:1977-1980`, `:165`, `:173-177`) did not propagate to readers 800 lines away.

## Scope

New `nanobot/runtime/state_access.py` (stdlib only, ≤300 LOC per the #942 module rule, added to `runtime_deny.py`), exposing:

```python
@dataclass(frozen=True)
class Window:
    rows: tuple[dict, ...]        # oldest → newest, already phase-filtered
    status: str                   # "complete" | "partial" | "unavailable"
    requested_from: str
    covered_from: str | None
    files_read: int
    files_skipped: int
    bytes_read: int
    notes: tuple[str, ...]        # "gz_corrupt:<name>", "cap_bytes", "dir_missing", "permission", "beyond_retention"

def ledger_window(state_dir, *, since_ts, phases=None, max_bytes=4 * 2**20) -> Window
def artifacts(state_dir, *, newest, max_age_hours=None, statuses=None) -> Window   # results ∪ requests ∪ archive (flat); mtime desc, tie → name desc
def latest_file(directory, pattern, *, max_age_s) -> Latest                        # key=(mtime, name); .path .age_s .stale .status
def sidecar(path, *, default, max_bytes) -> Sidecar                                 # .data .status ∈ present|absent|corrupt|oversize|permission
```

Semantics (from the ADR, D-1): `ledger_window` streams newest → oldest file, orders archives by the date in the name (never glob order), substring-prefilters `'"phase": "<p>"'` before `json.loads` (#1059), skips a corrupt file per-file and names it in `notes`, stops with `status="partial"` at `max_bytes`, marks `beyond_retention` when `since_ts` is older than `cycle_ledger` retention. Never raises.

Also in this PR:
- `tests/fixtures/ledger_live_shape/`: one real, redacted day file from the host (`cycles-2026-09-01.jsonl.gz`, 683 rows; redact `task_title`/`expected_outcome_claim` text but keep every key and every `phase`/`outcome`/`demand_id` prefix), the 3-row post-rotation `cycles.jsonl`, one archived and one live result file, one archived request. The redaction script is committed next to the fixture.
- `tests/test_state_access_hygiene.py` (same shape as `test_import_hygiene.py`): fails when any module outside `state_access.py` contains `"ledger" / "cycles.jsonl"`, `glob("cycles-*`, `"subagents" / "results"`, `glob("evolution-*`, or `max(` … `st_mtime`. Allowlist = `cycle_ledger.py` (writer), `scripts/cleanup_subagent_queue.py` (archiver), and an explicit grandfather list of today's readers that a later PR may only shrink.
- `docs/adr/ADR-002-bounded-state-access.md` = the Decision section of #1173, plus an index row in `docs/adr/README.md`.

## Acceptance Criteria

- [ ] `state_access.py` ≤300 LOC, stdlib only, listed in the runtime deny-set; `ruff` clean.
- [ ] `ledger_window` on the live-shape fixture: `since_ts` 24 h before the newest row returns rows from both the gz archive and the 3-row live file, `status == "complete"`, `covered_from` ≤ `since_ts`, `files_read == 2`.
- [ ] `ledger_window` with a corrupt `.gz` in the middle of the range: rows from the other files, `files_skipped == 1`, `notes` names the file, `status == "partial"`.
- [ ] `ledger_window` with `max_bytes` smaller than the corpus: `status == "partial"`, `notes` contains `cap_bytes`, no exception; memory bounded (assert rows count ≤ what fits).
- [ ] `ledger_window` with an unreadable ledger dir (missing → `dir_missing`; on POSIX chmod 000 → `permission`): `status == "unavailable"`, `rows == ()`.
- [ ] Archive ordering is by the date in the name: a fixture file named `cycles-old.jsonl.gz` (which `cycle_ledger._day_str` rejects) is skipped with a note, not sorted into the window.
- [ ] `artifacts(newest=3)` over a fixture with 1 live + 5 archived results returns the 3 newest across both dirs; `statuses={"failed","blocked","error"}` filters; equal mtimes tie-break on name descending (test stamps identical `os.utime`).
- [ ] `latest_file` with two files stamped to the same mtime returns the lexicographically later name; `.stale` is `True` when age exceeds `max_age_s`; missing dir → `status == "dir_missing"`, no exception. (This is the #1168 acceptance test, written against the new function; #1177 swaps the caller.)
- [ ] `sidecar` distinguishes `absent`, `corrupt`, `oversize`, `permission` (POSIX-only test for the last), and `present`.
- [ ] Prefilter proof: a test with 10,000 non-matching rows and 3 matching rows asserts `json.loads` is called ≤ 3 + (number of lines that contain the substring) times (monkeypatch counter).
- [ ] `tests/test_state_access_hygiene.py` passes on the grandfather list and FAILS when a new `"ledger" / "cycles.jsonl"` literal is added to a scratch module in the test.
- [ ] Deploy-gate pattern pinned: a test asserts the exact string `deploy_release.sh` greps (`'"phase": "outcome"'`) matches ≥1 line of the fixture `cycles.jsonl` (the #1153 class).
- [ ] `docs/adr/ADR-002-bounded-state-access.md` exists; `docs/adr/README.md` index updated.
- [ ] No production caller changed; full `python -m pytest tests/` green locally and in CI; deploy to the host is a no-op behaviour change (verify one cycle completes and `scorecard/latest.json` is unchanged in shape).
- [ ] **Do not edit `nanobot/runtime/demand.py` or `nanobot/runtime/goal_gap_futility.py`** — pi-2 owns them under #1166.

## Non-goals

No caller migration, no index/cache of the ledger, no change to `cycle_ledger` rotation or the archiver, no retries/alerting.

ozand:
## Summary

Add the bounded, fail-open `state_access` foundation from ADR #1173. This foundation defines the access contract but intentionally migrates no production caller; caller migrations are tracked by #1175–#1179.

## Contract settled here

- `Window.status` is exactly `complete`, `partial`, or `unavailable`; `rows == ()` is reserved for unavailable/empty data and is never itself used as the error signal.
- `Sidecar.status` is exactly `present`, `absent`, `corrupt`, `oversize`, or `permission`.
- `Latest.status` distinguishes `present`, `empty`, `dir_missing`, and `permission`; `stale` is computed from the caller-supplied age bound.
- `ledger_window` reads the active ledger plus valid dated gzip archives newest-first, stops before archives older than `since_ts`, enforces an explicit uncompressed byte cap, prefilters phases before JSON parsing, and reports skips/caps in `notes`.
- `artifacts` unifies live `results/`, `requests/`, and flat `archive/` directories with bounded candidate selection and deterministic `(mtime, name)` ordering.
- All public readers catch failures and return a status-bearing result; they never raise and never turn unavailable into a successful empty read.

## Architecture / safety

- `nanobot/runtime/state_access.py`: **294 LOC**, stdlib-only, no provider/LLM imports or calls.
- Added to `nanobot/runtime/runtime_deny.py`.
- Explicit caps are justified in comments: 4 MiB uncompressed ledger read, 256 artifact candidates, and the existing 90-day ledger retention boundary.
- No production caller was changed. Expected future callers are the overlapping readers listed in ADR #1173, migrated independently by #1175–#1179; this foundation intentionally changes behavior nowhere.

## Failing-test-first evidence

The new tests were first exercised against the pre-change tree. The relevant RED failures were:

```text
ERROR: not found: .../tests/test_state_access.py::test_ledger_window_reads_archive_and_active_in_order
(no match in any [<Module test_state_access.py>])

ERROR: not found: .../tests/test_state_access.py::test_ledger_window_cap_is_partial
(no match in any [<Module test_state_access.py>])
```

This is the expected pre-change failure because `state_access.py` and its tests did not exist on the pre-change tree. After implementation, all acceptance tests pass; fixture-specific failing assertions were also verified during development and corrected before merge.

## Test / fixture coverage

- `tests/test_state_access.py` covers archive + active ordering, corrupt archive partial status, byte cap, missing directory, invalid archive names, unified artifacts, status filtering, mtime/name tie-break, stale latest file, sidecar absent/corrupt/oversize, phase prefilter, and the deploy gate `"phase": "outcome"` contract.
- `tests/test_state_access_hygiene.py` covers the explicit grandfather list and scratch-pattern detection.
- `tests/fixtures/ledger_live_shape/` contains a redacted 683-row dated gzip shape plus a 3-row active ledger; `tests/subagents/` contains live/archived result and request shapes. The fixture preserves observed field/phase shapes while replacing task prose with `REDACTED`.
- A sanitized read-only SSH inspection confirmed the host shape before fixture creation: 53 ledger directory entries with dated gzip archives, live `subagents/results` and `requests`, and archived subagent artifacts. No secrets or payload text were copied.

## Verification

- Focused foundation + hygiene tests: **14 passed**.
- Ruff: passed for changed Python files.
- Branch was pushed from `T:/Code/eeebot-wt-1174`; status clean and `git log origin/feat/1174-state-access..HEAD` empty.
- No deployment performed. Rollout remains operator-owned.

Closes #1174 after merge. Subsequent caller migrations remain separately governed; do not infer migration completion from this foundation PR.


ozand:
Reviewed the module rather than the diff summary. The contract is the right one and the redacted live-shape fixture is exactly the discipline this repo has been missing — a mock built from the code under test proves nothing, and this one is not. Three findings, one of which I think should be settled before merge because everything in #1175–#1179 will be written against this dataclass.

### 1. `covered_from` cannot express a hole at the recent end (the one I care about)

`sources` is `[cycles.jsonl] + archives newest-first`, and each file is consumed in file order, which is oldest-line-first. The byte cap breaks *inside* whichever file exhausts it.

So under cap pressure the read is: all of today, all of yesterday… then the **oldest part** of the file where the budget ran out, with that file's newest rows dropped. The missing span is in the *middle* of the window, not at its old end.

`covered_from` is then `min(covered)` — the earliest ts observed — so the Window reports coverage starting at that file's morning while silently missing that file's evening. A consumer that trusts `covered_from` gets a false coverage claim, which is the precise failure ADR #1173 exists to eliminate.

Mitigation already present: `capped` sets `status = "partial"` and appends `cap_bytes`, so a consumer that honours `status` is safe. That is why this is a design question and not a bug report. But `partial` alone does not say *where* the hole is, and five follow-on issues are about to start reading this. Options, cheapest first:
- add `covered_to` alongside `covered_from` (my preference — one field, and it makes the hole expressible);
- or read each capped file from its tail instead of its head, so the truncation lands at the old end where `covered_from` already describes it;
- or document in the dataclass docstring that `partial` + `cap_bytes` means the hole may be interior and `covered_from` is a lower bound only.

Any of the three is fine. Silence is not, because the next five PRs will encode whichever assumption they guess.

### 2. Non-permission `OSError` is reported as `permission`

Five places: lines 80-81, 140-141, 162-163, 235-236, 276-277. Each catches `PermissionError` first and then `OSError` — so by construction the second handler only ever sees errors that are *not* permission problems, and labels them `permission` anyway. EIO, ENOTDIR, a truncated read all surface as a permission diagnostic.

`status` stays correct (`unavailable`), so this is not a correctness bug for gating. It is a truthfulness bug in the field a human reads to find out what broke, inside the module whose stated purpose is "returns status instead of collapsing unavailable input into empty data". It also drives control flow at line 185: `if error == "permission" and files_read == 0` returns early, so a corrupt active ledger takes the permission branch.

Suggest a distinct note (`io_error`, or the errno name) for the second handler.

### 3. Nit: `_` is a live variable

Line 97: `[path for _, path in archives if _ + timedelta(days=1) >= since]`. `_` is bound and then used in the condition. It works, and the horizon arithmetic is correct — I checked both boundaries — but `_` conventionally means discarded, so this reads as a bug on first pass. Name it `day`.

### Verified, not just read

- Horizon arithmetic in `_ledger_sources` is right in both directions: for `since = 2026-08-28T07:08Z`, `cycles-2026-08-28.jsonl.gz` is kept (day + 1d ≥ since); for `since = 2026-08-29T10:00Z` it is correctly dropped, since every row in it precedes the window.
- Newest-first source ordering means cap pressure sheds the *oldest* data across files, which is the right priority.
- 294 LOC, stdlib-only, registered in `runtime_deny.py`, no production caller migrated — all as the body claims.

Nothing here blocks CI. Finding 1 is worth ten minutes now rather than five migrations later.

Issue #1183, opened 2026-09-02
Reflector v2 mint path is permanently off: a 512 KB size guard discards a file the next line reads only 50 lines of


ozand:
## Problem

The reflector → v2 lesson path is switched off permanently, silently, and by a guard that protects against nothing.

`knowledge_curator.promote_reflector_recommendations_to_v2()` opens with:

```python
stat = path.stat()
if stat.st_size > 512 * 1024 or time.time() - stat.st_mtime > 90 * 86400:
    return 0
rows = [json.loads(line) for line in path.read_text(encoding="utf-8").splitlines()[-50:] if line.strip()]
```

Live on the host:

```
699393 bytes  /var/lib/eeepc-agent/self-evolving-agent/state/reflector/reflections.jsonl
```

699,393 > 524,288, so the function returns 0 before parsing anything. Verified against the deployed tree (`current/nanobot/runtime/knowledge_curator.py:922-925`, release `20260902T010947Z-canonical-a48975c1`).

## Why the guard is wrong, not merely unlucky

Three things make this worse than a threshold that needs raising.

1. **The guard rejects the file to avoid reading it, and the very next line reads only `[-50:]`.** The work is already bounded to the newest 50 rows. Whatever the cap was defending — memory on a 2 GB host — is defended by the tail slice regardless of file size.
2. **`reflections.jsonl` is append-only with no rotation or chunking.** It crosses 512 KB once and never comes back. This is not an outage that resolves; it is a one-way shutdown. The store passed the line around 2026-08-29, which is exactly the `first_seen` date on all four v2 reflector cards and the last day anything was minted.
3. **Failure is indistinguishable from "nothing to promote".** `return 0` is the same value a healthy empty run returns. No log line, no ledger row, no counter.

Measured on #1171: **386 validly-structured mint attempts** (`approach_hint` / `error_pattern` with non-empty detail) sit in the log, none of them rejected on merit. They are discarded before parsing.

## Consequence beyond this function

#1106 fixed the *content* of minted reflector lessons and its measured knowledge lift was zero. This is why: there was nothing being minted for it to improve. A quality fix applied to an event that does not occur is unobservable by construction, and #1106's live criterion was uncheckable for a reason that had nothing to do with #1106.

## Same class, third instance tonight

- #1166 — goal-gap `attempt_count` read only the un-rotated ledger, reset every midnight, returned 0.
- #1178 — strategist scorecard history is dead because the file is 516 KB against a 256 KB cliff.
[...]

Issue #1178, opened 2026-09-02
fix(runtime): side-role size cliffs — strategist history input dead at 256 KB, curator reflector promotion dead at 512 KiB, reflections.jsonl has no rotation, read-blank-then-write-back sidecars


ozand:
Parent ADR: #1173. Depends on #1174 only for the tail helper; the rotation and the write-guards are independent and can ship first. The `demand._reflection_items` part waits for #1166 (pi-2 owns `demand.py`).

## Problem

Side-role readers return the empty value when a file exceeds a size cliff, and two of those cliffs have already been crossed on the host. Separately, five sidecars are read-blank-then-written-back, so one transient read error erases their history.

Live (`eeepc-lan`, 2026-09-02):

| reader | cliff | live size | effect |
|---|---|---|---|
| `strategist._read_lines` (`strategist.py:85`, magic `256_000`) feeding `_scorecard_input` `history_7d` | 256,000 B | `scorecard/history.jsonl` = **513,471 B** (rotates at 512 KiB, `scorecard.py:308`) | `history_7d` is `[]` right now and for ~half of every rotation period; the strategist has run 7 times with no trend data and no record of the drop. Never reads `scorecard/archive/*.jsonl.gz`. |
| `knowledge_curator.promote_reflector_recommendations_to_v2` (`knowledge_curator.py:923`, magic `512 * 1024`) | 524,288 B | `reflector/reflections.jsonl` = **697,881 B** | permanent `return 0` — reflector recommendations never become v2 lesson cards |
| `knowledge_lift.compute_knowledge_digest` (`knowledge_lift.py:169`, `MAX_FILE_BYTES = 50_000`) | 50,000 B | same file | reflections never enter the digest; `knowledge_unchanged` watermark ignores reflection churn |
| `reflector._journal_tail`, `knowledge_curator.iter_lessons`, `demand._reflection_items`, `reflector.mark_reflection_consumed` | **none** | same file | whole 698 KB parsed per call; no rotation exists for `reflections.jsonl` |
| `skill_eval_harness._load_rows` (2 MB) → `_rewrite_rows`; `knowledge_lift._read_eval_rows` (2 MB) → `_atomic_write_eval_rows`; `validator_harness._prune_last_runs` (8 MiB → **`_atomic_write(path, "")`**) | oversize → `[]` → **file rewritten from `[]`** | not yet hit | history truncated on the first oversize read |
[...]

Issue #1166, opened 2026-09-01
goal-gap futility can never fire: attempt_count only sees the un-rotated ledger, so it resets every midnight


ozand:
## Problem

The goal-gap futility suppression (`nanobot/runtime/goal_gap_futility.py`, shipped in #1069) counts a gap's attempts from ledger rows that only ever cover the **current, un-rotated** `cycles.jsonl`. The ledger rotates daily into `cycles-YYYY-MM-DD.jsonl.gz`, so the counting window resets every midnight and `attempt_count` can never accumulate toward the threshold of 10.

The feature has been deployed since 2026-08-28 and has produced **zero** `goal_gap_futile` ledger rows and **zero** suppressions.

## Evidence

`state/demand/futility.json` on the live host (read 2026-09-01 22:35 UTC) — the tracking itself is alive and updating, but every counter is stuck at zero:

| gap_id | metric | attempt_count | first_seen |
|---|---|---|---|
| goal-gap-5d4d5a9dc822 | repeat_failure_rate | 0 | 2026-08-26 |
| goal-gap-acb65b1911ab | confirmed_integration_ratio | 0 | 2026-08-26 |
| goal-gap-2d9ab3aa9d09 | confirmed_ratio | 0 | 2026-08-27 |
| goal-gap-a820ca0c8bb3 | stale_feeds | 0 | 2026-08-27 |
| goal-gap-c09521b7459a | heldout_gap | 0 | 2026-08-28 |

Recomputing `_integrated_count`'s own `proposed ∩ outcome:success` intersection over the **full** ledger (live file plus all 51 rotated archives, 3148 relevant rows, 689 success cycles) gives the real numbers:

| gap_id | proposed cycles | ∩ success = true attempt_count |
|---|---|---|
| goal-gap-5d4d5a9dc822 | 0 | 0 |
| goal-gap-acb65b1911ab | 0 | 0 |
| goal-gap-2d9ab3aa9d09 | 19 | **8** |
| goal-gap-a820ca0c8bb3 | 1 | **1** |
| goal-gap-c09521b7459a | 3 | **1** |

`goal-gap-2d9ab3aa9d09` sits at 8 of the 10 needed. The live sidecar reports 0 for it and will keep reporting ~0 indefinitely, because each day's rotation throws the history away.

## Cause

Two live-file-only readers feed the path:

[...]

PR #1192, opened 2026-09-02
fix(runtime): preserve memory and skills under prompt cap

### LINKED ISSUE #1191: System prompt silently drops memory and skills when bootstrap exceeds the 24k cap — live since 2026-09-01 16:34 UTC
## Problem

`ContextBuilder.build_system_prompt` caps the system prompt with a flat character slice. Bootstrap files are assembled first and are documented as never truncated, so once they alone exceed the cap, **memory and the skills catalogue are silently removed from every prompt** and the last bootstrap file is severed mid-token.

This is happening right now on the live host, release `5b76f524`. Measured by calling the deployed `ContextBuilder` read-only against the live workspace:

```
CAP                      24000
identity chars            1446
AGENTS.md chars          22972
identity + AGENTS.md     24418   <-- over the cap before anything else is appended
final system prompt len  24056
TRUNCATED?                True

contains "# Memory"       False
contains "Active Skills"  False
contains "Skills"         False

last 120 chars:
'\n   `python3 -m unittest tests.test_<your_target> -v` (or\n   `py\n\n[truncated: system prompt capped by memory discipline]'
```

Every cycle currently runs with no memory block, no active skills and no skills catalogue.

## Mechanism

`nanobot/agent/context.py:84-86`:

```python
system_prompt = "\n\n---\n\n".join(parts)
if len(system_prompt) > self.MAX_SYSTEM_PROMPT_CHARS:
    trimmed = system_prompt[: self.MAX_SYSTEM_PROMPT_CHARS]
    system_prompt = trimmed + "\n\n[truncated: system prompt capped by memory discipline]"
```

Part order is identity → bootstrap → active skills → skills summary → memory.

The `#939 Part E` docstring says the ordering exists "so skills are not truncated when the memory block is huge", and places memory last for that reason. The pressure arrived from the other end: bootstrap is first and never trimmed, so it consumed the whole budget and evicted exactly what the ordering was meant to protect. The mitigation guards the wrong end of the list.

Two properties make it invisible:
- the notice is appended to the *end* of a string nobody inspects, not logged or counted;
- a prompt missing its memory section is structurally valid, so nothing downstream errors.

## When it started

`AGENTS.md` char counts across consecutive commits; budget available to it is `24000 − 1446 identity − 7 separator = 22,547`:

| commit | UTC | chars | |
|---|---|---|---|
| `74135026` | 2026-08-31 01:54 | 22,256 | under by 291 |
| `4dfc6ee4` | 2026-09-01 16:34 | 23,389 | **over by 842** |

Dark since **2026-09-01 16:34 UTC**. The commit that did it is a routine `+19`-line append, indistinguishable from the nineteen before it.

## Secondary: three of four bootstrap files do not exist

`BOOTSTRAP_FILES = ["AGENTS.md", "SOUL.md", "USER.md", "TOOLS.md"]` (`context.py:19`). On the host workspace only `AGENTS.md` exists; `SOUL.md`, `USER.md` and `TOOLS.md` are absent, and are not tracked in the product repo either. Three quarters of the list is dead configuration, skipped silently — the "reads a path that does not exist" shape of #1182 and #1153.

## Acceptance criteria

- [ ] Exceeding the cap is **reported**, not merely marked: a log line or ledger row naming which sections were dropped and by how many characters. The current trailing notice ran for twelve hours unnoticed.
- [ ] Memory and the skills catalogue survive bootstrap pressure — either by budgeting per section rather than slicing the concatenation, or by truncating bootstrap itself and saying so. Choose one and state why; do not add a second silent cap (see #1183, #1178, #1166 for how that ends).
- [ ] Truncation never cuts mid-token. If a section must be trimmed, trim at a line boundary.
- [ ] Test: with a bootstrap file larger than the whole budget, assert `# Memory` and the skills catalogue are still present in the output (or that the build refuses loudly). It must FAIL against the current tree — quote the observed failure in the PR body.
- [ ] Test: the drop is observable — assert the report names the dropped sections.
- [ ] Resolve `SOUL.md` / `USER.md` / `TOOLS.md`: create them or remove them from `BOOTSTRAP_FILES`. Missing must not look the same as empty.
- [ ] Full pytest green (shared runtime module — run the whole suite, not the touched files); CI green, Linux matrix authoritative.
- [ ] **Live verification** after deploy: call `build_system_prompt` on the host workspace read-only and quote `contains "# Memory"` and `contains "Skills"` as `True`, alongside the total length and the cap.

## Context

Found while measuring #1188 (reflection lane appending ~20 lines/day to `AGENTS.md`). #1188 is the upstream policy question — whether the loop should be writing that file at all. This issue is the runtime defect that turns that growth into silent capability loss, and it is worth fixing independently: any bootstrap file large enough would do the same.

Full measurement: https://github.com/ozand/eeebot/issues/1188#issuecomment-5504204519

ozand:
## Summary

Closes #1191.

`ContextBuilder` now trims oversized bootstrap content at complete line boundaries before allowing bootstrap pressure to evict active skills, the skills catalogue, or memory. Prompt-cap loss is reported through a structured warning naming each affected section and dropped character count. The obsolete trailing truncation marker is removed.

`BOOTSTRAP_FILES` is now explicitly `['AGENTS.md']`; `SOUL.md`, `USER.md`, and `TOOLS.md` were not tracked product bootstrap files and are no longer silently probed.

## Design choice

The pressure observed in #1191 comes from the bootstrap end of the prompt: `AGENTS.md` alone exceeds the cap before memory and skills are appended. We defend the sections after bootstrap by trimming bootstrap first at line boundaries. If a protected section itself exceeds the cap, that section is also trimmed at line boundaries and the loss is reported. No second silent cap or opaque trailing marker is used.

## Regression evidence

The new tests were run against the pre-change `origin/main` before implementation and failed as required:

- `test_oversized_bootstrap_preserves_memory_and_skills_and_reports_drop`: failed because `nanobot.agent.context` had no `logger` and the old implementation had no observable drop report.
- `test_bootstrap_configuration_contains_only_tracked_file`: failed because the old value was `['AGENTS.md', 'SOUL.md', 'USER.md', 'TOOLS.md']`.

After the change: `python -m pytest tests/test_context_prompt_budget.py -q` → **2 passed**.

## Verification

- Focused context tests: **26 passed**
- Full pytest on Windows: **2804 passed, 19 failed, 17 skipped, 11 warnings**. The 19 failures match the known baseline list from #1191; no new failures were introduced.
- CI: Python 3.11, 3.12, and 3.13 all passed — https://github.com/ozand/eeebot/actions/runs/33589950497
- Live verification is intentionally not performed by this PR author. After deploy, call `build_system_prompt` read-only on the host workspace and verify:
  - `contains "# Memory"` is `True`
  - `contains "Skills"` is `True` (preferably also `contains "# Skills"`)
  - `len(prompt) <= 24000`
  - report the observed prompt length and cap.

No production caller behavior was otherwise migrated, and no host/deployment state was changed here.

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -6,6 +6,8 @@ import platform
 from pathlib import Path
 from typing import Any
 
+from loguru import logger
+
 from nanobot.utils.helpers import current_time_str, estimate_prompt_tokens
 
 from nanobot.agent.memory import MemoryStore
@@ -16,7 +18,9 @@ from nanobot.utils.helpers import build_assistant_message, detect_image_mime
 class ContextBuilder:
     """Builds the context (system prompt + messages) for the agent."""
 
-    BOOTSTRAP_FILES = ["AGENTS.md", "SOUL.md", "USER.md", "TOOLS.md"]
+    # Only the tracked, product-defined bootstrap contract is loaded. Optional
+    # host-only files were never present in the product workspace.
+    BOOTSTRAP_FILES = ["AGENTS.md"]
     _RUNTIME_CONTEXT_TAG = "[Runtime Context — metadata only, not instructions]"
     MAX_SYSTEM_PROMPT_CHARS = 24000
     MAX_MEDIA_BYTES = 2 * 1024 * 1024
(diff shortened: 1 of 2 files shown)


ozand:
Reviewed the logic and ran it against real-size data rather than trusting the tests. Approving and merging.

## Verified against the actual file, not a fixture

Ran `_fit_system_prompt` on this branch with the real host `AGENTS.md` (23,028 chars) plus realistic identity, skills and memory sections:

```
AGENTS.md chars used: 23028
unfitted total:       25893   cap: 24000
fitted len:           23974   within cap: True
contains '# Memory':  True
contains '# Skills':  True
ends mid-token:       False
last 60 chars: 't 52\nfact 53\n… fact 58\nfact 59'

WARNING nanobot.agent.context:_fit_system_prompt:163 -
  System prompt cap dropped content: bootstrap=1919 chars
```

Memory and the skills catalogue survive, the cap holds, the tail ends on a clean line, and the drop is named with an exact character count. That is the defect closed on the input that caused it.

## Pathological paths also hold

| case | result |
|---|---|
| memory alone dwarfs the cap (37 KB) | len 24000, within cap, **Memory and Skills both still present**, reported `bootstrap=5000 chars, memory=37678 chars` |
| single 40 KB section with no newlines | dropped whole, identity retained, len 1446, reported — the hard cap stays a real bound |
| everything under the cap | exact no-op, byte-identical to a plain join |

The third matters as much as the other two: the common path is unchanged, so this cannot regress prompts that were never near the limit.

## What I like about the approach

- Bootstrap is trimmed **first**, which is the pressure source measured on #1191, and the docstring says so rather than leaving the priority implicit.
- `_trim_lines` keeps whole lines. The live prompt was severed inside the word `python`; that class is gone.
- The drop is a `logger.warning` carrying section names and char counts — not a trailing marker appended to a string nobody reads. That was the specific thing that let this run twelve hours unnoticed, and it is the part I care most about.
- No second silent cap was introduced. Given #1183, #1178 and #1166 were all exactly that mistake, this was the main way the fix could have gone wrong and it did not.

## One note, not a blocker

`BOOTSTRAP_FILES` drops to `["AGENTS.md"]`. I checked: `SOUL.md`, `USER.md` and `TOOLS.md` appear **nowhere else in the repository** — not in tests, docs, templates or config, only at `context.py:19`. So nothing in-tree breaks and the removal is correct on the evidence.

The residue is that `ContextBuilder` is shared with the operator-facing chat agent, so a workspace that *does* carry a `SOUL.md` will now have it ignored silently — where previously an absent one was skipped silently. One silent case traded for another, narrower one. Not worth holding the merge for, since this fix is live-urgent and no such workspace is known, but if the chat-agent side ever grows those files the list is where to look.

## Process

Full suite 2804 passed / 19 failed / 17 skipped, and the 19 match the known Windows baseline exactly. CI green on 3.11 / 3.12 / 3.13. Pushed after every commit — that clause was tightened tonight after an unpushed commit had to be rescued from a dead session, and it was followed.

Merging and deploying. Live verification after deploy: call `build_system_prompt` on the host workspace read-only and quote `contains "# Memory"` and `contains "Skills"` as `True` alongside the length and cap. I will post that on #1191.

Issue #1162, opened 2026-09-01
deploy health gate scans 3 hours of pre-deploy journal: --since gets a UTC string but journalctl parses it as host-local (MSK)


ozand:
## Problem

The post-deploy health gate in `host/eeepc/scripts/deploy_release.sh` bounds its journal queries with

```sh
FLIP_JOURNAL_TS=$(date -u +"%Y-%m-%d %H:%M:%S")
...
sudo journalctl -u eeepc-self-evolving-subagent-bridge.service --utc --since "$FLIP_JOURNAL_TS" ...
```

`date -u` produces a UTC wall-clock string, but `journalctl` parses a bare timestamp in the **host's local time**, and the host runs MSK (UTC+3). `--utc` only changes how journalctl *formats* its output; it does not change how `--since` is parsed.

Confirmed on the host, using the exact value the gate passed during the deploy of `20260901T212713Z-canonical-5ebebd36`:

```
$ date -u
Tue Sep  1 09:50:10 PM UTC 2026
$ sudo journalctl -u eeepc-self-evolving-subagent-bridge.service --utc --since '2026-09-01 21:35:36' --no-pager | head -1
Sep 01 18:36:49 eeepc systemd[1]: Starting eeepc-self-evolving-subagent-bridge.service ...
```

The window opens at 18:35:36 UTC instead of 21:35:36 UTC — exactly three hours early.

## Impact

Both FAIL branches of the gate (the traceback check and the `main process exited` check) scan three hours of **pre-deploy** history. A crash or traceback that happened before the deploy started — and that the operator may already have fixed by deploying — will be attributed to the new release and roll it back, then restart the bridge on the old one.

This is the fifth defect in this gate, after #1153, #1155, #1157 and #1159. It has not fired yet only because the host happened to have a clean three-hour window during the verification deploy.

## Fix

Pass an unambiguous instant rather than a bare local-time string. `journalctl --since` accepts `@<epoch-seconds>`, which has no timezone ambiguity:

```sh
FLIP_JOURNAL_TS="@$(date -u +%s)"
```

Any equivalent form that removes the ambiguity is acceptable; the requirement is that the window starts when the gate arms, on any host timezone.

## Acceptance criteria

- [ ] The gate's journal window starts at the moment the gate arms, regardless of the host's timezone offset.
[...]

PR #1195, opened 2026-09-02
fix(state): break _latest_json_file mtime ties by filename

### LINKED ISSUE #1168: flaky: _latest_json_file breaks mtime ties by glob order, so 'newest report' can resolve to the oldest
## Problem

`test_commands.py::test_runtime_state_prefers_newest_evolution_report` passes or fails depending on how fast the filesystem clock ticks during the run. It passes in isolation and passed on `origin/main` in a two-file run; it fails when a slower test module runs before it in the same session.

```
>       assert runtime["cycle_id"] == "new"
E       AssertionError: assert 'old' == 'new'
```

Not platform-neutral: CI (Linux, all three Python versions) is green, and the failure reproduces on Windows.

## Cause

The test writes two report files back to back:

```python
(reports_dir / "evolution-20260401.json").write_text(json.dumps({"cycle_id": "old"}), ...)
(reports_dir / "evolution-20260412.json").write_text(json.dumps({"cycle_id": "new"}), ...)
```

and `load_runtime_state` picks the winner in `nanobot/runtime/state.py:49`:

```python
def _latest_json_file(directory: Path, pattern: str) -> Path | None:
    ...
    matches = directory.glob(pattern)
    return max(matches, key=lambda p: p.stat().st_mtime if p.exists() else 0, default=None)
```

When both writes land inside one mtime tick the two keys are equal, and `max` returns the **first** maximum it encounters — glob order, which puts `evolution-20260401.json` (the older date) first. Linux mtime resolution is fine enough that the two writes almost never tie; a coarser Windows filesystem timestamp ties routinely, and whether it ties depends on what ran before.

## Discussion

The flake points at something real in the selector rather than only in the test. These filenames carry their date, and the field the caller wants is "the newest report" — but the tie-break is glob order, which is unrelated to recency in either sense. A tie today resolves to the lexicographically-first name, i.e. the *oldest* report.

Two directions, either acceptable:

- Break the tie deterministically on the filename, so equal mtimes resolve to the later-named file: `max(matches, key=lambda p: (mtime, p.name))`. Smallest change, keeps mtime as the primary signal.
- Order by the date encoded in the name for `evolution-*.json`, treating mtime as the fallback. Closer to the intent, larger change, and it only works for patterns that carry a date.

Not urgent for production — on the host the reports are written minutes apart — but the current behaviour picks the oldest report whenever two are written in the same tick, which is not what the function name promises.

## Acceptance criteria

- [ ] Two report files written in the same mtime tick resolve deterministically to the newer one; a test proves it by stamping both files with an identical `os.utime` rather than relying on write speed, and that test must FAIL against current code.
- [ ] The existing `test_runtime_state_prefers_newest_evolution_report` no longer depends on how long the preceding tests took (running it after a slow module must not change the result).

## Context

Surfaced while merging #1162 (0cd94452): the Windows full-suite count went 18 → 19 failures on that branch. Confirmed unrelated to the change — the same two-file pair passes on clean `main` and fails on the branch purely because the added `test_deploy_release.py` cases shift timing, and CI was green on all three versions. Recorded here so the next person who sees 19 does not re-derive it.

ozand:
Closes #1168.

## What was wrong

```python
return max(matches, key=lambda p: p.stat().st_mtime if p.exists() else 0, default=None)
```

`max` returns the **first** maximal element. When two files share an mtime the winner was therefore decided by `glob` order — by the filesystem, not by anything meaningful. Two reports written back to back land on the same mtime whenever the clock has not ticked between the writes, so "newest evolution report" could resolve to the oldest one.

## Change

The sort key becomes `(mtime, name)`. The tiebreak is deterministic, and for the date-stamped patterns this function is actually called with (`evolution-YYYYMMDD.json`) lexical order is chronological order, so it is correct rather than merely stable. Call sites that pass an exact filename (`latest.json`, `report.index.json`, `backlog.json`) match at most one file and are unaffected.

## Evidence

**The new tie test fails deterministically against the pre-change tree.** It forces the tie with `os.utime` instead of hoping the clock cooperates:

```
FAILED tests/test_commands.py::test_latest_json_file_breaks_mtime_ties_by_name
E  assert WindowsPath('.../evolution-20260401.json') == WindowsPath('.../evolution-20260412.json')
```

It returns the **older** file — the bug, made deterministic. It asserts both write orders so it cannot pass by luck. A second test pins the opposite direction: a lexically later name must still lose to a strictly newer mtime, so the tiebreak cannot override real ordering. That one passes both before and after by design — it guards against over-correction, it is not a bug repro, and I am not claiming it as one.

**The original intermittent test.** `test_runtime_state_prefers_newest_evolution_report` reproduces only when a slower module runs first in the same session, per the issue. Running `tests/test_autoevolve.py tests/test_commands.py` together, ten runs each side:

| | failures of the original test |
|---|---|
| pre-change | **4 / 10** |
| with this change | **0 / 10** |

Consistent with the root cause being removed, but ten runs of an intermittent failure is supporting evidence, not proof. The deterministic test above is the actual proof; this table is why I believe it addresses the reported flake rather than a neighbouring one.

Failure counts in that paired run were stable at 5 with the fix and varied 6–7 without it, the delta being exactly this test.

**Full suite** on Windows: matches the known baseline, no new failures in touched files. Linux CI is authoritative.

## Note

The fix is one line of key material plus a docstring. The docstring is longer than the change because the failure mode — `max` silently preferring iteration order on ties — is not visible from reading the expression, and this is the second time it has cost debugging effort: it produced a false alarm during #1162 review, where the failure looked like it belonged to that PR and did not.

### DIFF
--- a/nanobot/runtime/state.py
+++ b/nanobot/runtime/state.py
@@ -47,10 +47,25 @@ def _safe_read_json(path: Path | None) -> Any:
 
 
 def _latest_json_file(directory: Path, pattern: str) -> Path | None:
+    """Newest match by mtime, with filename as a deterministic tiebreak.
+
+    #1168: `max` returns the FIRST maximal element, so when two files share an
+    mtime the winner was decided by `glob` order — i.e. by the filesystem. Two
+    files written back to back land on the same mtime whenever the clock has
+    not ticked between them (coarse timestamps, or simply a fast write), and
+    the "newest report" then resolved to whichever the directory happened to
+    yield first. Falling back to the name is both deterministic and correct
+    for the date-stamped patterns used here (`evolution-YYYYMMDD.json`), where
+    lexical order is chronological order.
+    """
     if not directory.exists():
         return None
     matches = directory.glob(pattern)
-    return max(matches, key=lambda p: p.stat().st_mtime if p.exists() else 0, default=None)
+    return max(
+        matches,
+        key=lambda p: (p.stat().st_mtime if p.exists() else 0, p.name),
+        default=None,
+    )
 
 
 def _workspace_looks_like_eeepc_live_runtime(workspace: Path) -> bool:
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -1,4 +1,5 @@
 import json
+import os
 import re
 from pathlib import Path
 from unittest.mock import AsyncMock, MagicMock, patch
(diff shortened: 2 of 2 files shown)


ozand:
Full-suite result on Windows, single clean run:

```
18 failed, 2822 passed, 17 skipped in 1010.91s
```

**One fewer failure than the 19-failure baseline**, and `test_runtime_state_prefers_newest_evolution_report` does not appear in the failure list at all. The baseline included it; without the tie the count drops to 18 and the remaining failures are the unchanged set — 10 `autoevolve*`, 3 `bridge_locking` (flock), 2 `selfevo_export_security`, 1 `bridge_cycle_tags`, 2 `autoevolve_state`. No new failures in any touched file.

That is a third, independent line of evidence alongside the deterministic tie test and the 4/10 → 0/10 paired-module runs: the flake that defined the baseline is no longer in it.

PR #1210, opened 2026-09-02
fix(runtime): read rotated reflection archives for hints (#1207)

### LINKED ISSUE #1207: fix(runtime): build_reflection_hints reads only the live reflections.jsonl — #1178 rotation empties its window to ~2 rows for the first hours after each rotation
## Problem

`nanobot/runtime/reflection_context.py:122` (`build_reflection_hints`) reads the reflector journal from **one file only**, `state/reflector/reflections.jsonl`, taking a 128 KB tail and filtering to a 7-day TTL. It fails open to `[]`. Two callers put the result in front of an LLM:

- `nanobot/runtime/llm_proposer.py:2559`: every proposal's subagent request payload, `lessons_context.reflection_hints` (#1008 steering section for the executor);
- `nanobot/runtime/knowledge_lift.py:413`: the knowledge-context A/B harness.

#1178 (merged as #1204, `6bd50f4d`) added rotation to that file: at 512 KiB the live journal is gzipped to `reflector/archive/reflections-YYYY-MM-DD.jsonl.gz` and truncated. #1178 made `_journal_tail`, the skipped set, `mark_reflection_consumed`, the curator's `iter_lessons`/promotion top-up, `demand._reflection_items` and the knowledge-lift digest rotation-aware. This reader was not on the list.

The rotation fired today at 2026-09-02 09:43 UTC (12:43 MSK): live file 745,306 B → 2,447 B; archive `reflections-2026-09-02.jsonl.gz` 162,714 B, **667 rows**. At that instant this reader's window went from its full 7-day set to roughly two rows, and an empty result is indistinguishable from "no reflection matched this task".

## Live measurement (host, 2026-09-02 12:46 UTC / 15:46 MSK, release `20260902T092749Z-canonical-44605ba2`)

Live journal 3 h after rotation: **20 rows, 21,364 B** (the reflector appends about 7 rows/hour).

Deployed `build_reflection_hints(state_dir, title, target)` against the live state versus the same function against a temporary state dir holding archive + live concatenated (769,000 B), for the six most recent `llm-proposed-improvement` requests:

```
request-cycle-b303f507d6c3  live 3 | archive+live 3
request-cycle-ca65df314e3d  live 3 | archive+live 3
request-cycle-83c17976b9dc  live 3 | archive+live 3
request-cycle-62a4ad2dc7bd  live 3 | archive+live 3
request-cycle-64aa36de552c  live 3 | archive+live 3
request-cycle-9ec5ff111e8e  live 3 | archive+live 3
```

So it is **not biting at 3 h post-rotation**: 20 rows already saturate the 3-hint cap for current task shapes. The window it bit is the first hour or two after truncation. Request payloads by file mtime (`task_id == llm-proposed-improvement`, `subagents/requests` + `subagents/archive`):

| window (UTC) | requests | with hints | total hints |
|---|---|---|---|
| 2026-09-01 all day | 60 | 59 | 173 |
| 2026-09-02 00:00 → 09:43 (pre-rotation) | 36 | 35 | 105 |
| 2026-09-02 09:43 → 12:46 (post-rotation) | 14 | 11 | 32 |

3 of 14 post-rotation proposals shipped without hints against 1 of 36 before. Small, and it heals within hours. It recurs at every rotation: at ~160 rows/day and ~1 KB/row the live file reaches 512 KiB about every three days.

## Why it is the #1173 class

Cap in front of an already-bounded read, plus the variant the operator named on #1173 at 10:05 UTC today: **introducing rotation silently narrows every bounded reader of that file, and a tail reader is the easiest to miss because it already looks bounded and safe.** The fix from #1183 (curator mint path) would have switched itself off the same way had #1178's backfill not covered it; this reader has no backfill.

## What to do

Read the journal through `reflector.reflection_files` / `reflector.iter_reflection_rows` (newest two archives + live), or an equivalent that opens the archive only when the live tail is short, keeping the current byte and TTL bounds. No behaviour change when the live file is large.

## Acceptance criteria

- [ ] `build_reflection_hints` returns hints from `reflector/archive/reflections-*.jsonl.gz` when the live file holds no match; bounded to the same 128 KB-equivalent / 7-day window; still never raises.
- [ ] A test with a fixture of one rotated archive containing a matching row and a 2-row live file with no match **fails on the current tree** (quote the failure in the PR) and passes with the fix.
- [ ] `compute_knowledge_digest` (`knowledge_lift.py:165`) either keeps hashing the live tail (state that it flips on every reflector append, so the `knowledge_unchanged` gate is bounded only by the weekly cap) or is made rotation-consistent; either is fine, but say which.
- [ ] Live proof after the next rotation (quote UTC and MSK): live-file byte size right after truncation, and the first post-rotation `llm-proposed-improvement` request payload showing non-empty `lessons_context.reflection_hints` when archive+live holds a match for its task.
- [ ] Grep quoted in the PR for every remaining reader of `reflections.jsonl` outside `reflector.py`, with each one classified (rotation-aware / not a window read / N.A.).

## Not to build

- A second cap or a "journal recently rotated" flag. The reader should see the same rows before and after rotation.
- Any change to rotation size or the reflector writer.

Parent: #1173. Related: #1178 / #1204 (rotation), #1183 (same shape, curator mint path), #1008 / #1089 (this reader's origin).

ozand:
## Summary

`build_reflection_hints` now reads the existing rotation-aware ordered journal paths from `reflector.reflection_files`, including the newest rotated gzip archives and the live journal. It preserves the existing 128 KiB byte-tail budget, 7-day TTL, matching/ranking, three-hint bound, and fail-open behavior. The reflector rotation threshold and writer are unchanged.

The implementation establishes file order before applying the byte bound: it reads newest-to-oldest across the ordered archive/live paths, spends the existing byte budget, removes a partial leading line when a file is sliced, then restores chronological order before the existing newest-first candidate scan.

## Inputs and design

- Reused `nanobot.runtime.reflector.reflection_files`, which returns the newest rotated archives oldest-first followed by the live journal.
- `gzip.open(..., "rb")` reads archive bytes; `open(..., "rb")` reads the live file.
- The existing `_MAX_TAIL_BYTES = 128_000` remains the only byte bound in this reader; no new cap or rotation flag was added.
- The existing 7-day timestamp filter and 3-hint output cap remain unchanged.
- `compute_knowledge_digest` in `nanobot/runtime/knowledge_lift.py` remains live-tail-based. This change intentionally does not alter its digest contract; its bounded digest changes on live reflector appends, and the weekly knowledge-evaluation cap remains the control on that churn.

## Regression evidence

The new test was run against `origin/main` before the implementation:

```text
FAILED tests/test_reflection_context.py::test_reads_matching_hint_from_rotated_archive
E       AssertionError: assert 'preserve archive import resolution' in []
```

After the implementation:

```text
8 passed in 0.76s
35 passed in 2.81s  # reflection_context + knowledge_curator + knowledge_lift
```

## Validation

- Focused reflection-context tests: `8 passed`.
- Related rotation-reader tests: `35 passed` across `tests/test_reflection_context.py`, `tests/test_knowledge_curator.py`, and `tests/test_knowledge_lift.py`.
- Full pytest: `2883 passed, 18 failed, 17 skipped, 11 warnings` on Windows. This matches the known Windows baseline category (18 platform failures); the failures are unrelated to this two-file change. Representative baseline failures include local temporary-repository pushes using `master` (`src refspec master does not match any`) and Windows-only bridge lock/tag expectations (`fcntl` unavailable). The run completed in 17m24s.
- `ruff check nanobot/runtime/reflection_context.py tests/test_reflection_context.py`: passed.
- `git diff --check`: passed.

## Reader audit

Grep of runtime readers outside `reflector.py`:

- `nanobot/runtime/reflection_context.py`: **rotation-aware in this change** via `reflector.reflection_files`.
- `nanobot/runtime/demand.py`: **rotation-aware**, via `reflector.reflection_files` / `reflector.iter_reflection_rows`.
- `nanobot/runtime/knowledge_curator.py`: **rotation-aware** for curation reads via `_reflection_lines` / `_reflection_tail_lines`; its path discovery strings and legacy fallback paths are not journal-window reads.
- `nanobot/runtime/knowledge_lift.py`: `compute_knowledge_digest` is **not a journal-row window read**; it intentionally hashes the bounded live tail only, as stated above.

The remaining `reflections.jsonl` references in `reflector.py` are the journal writer/path and its own rotation-aware readers, excluded from this audit as requested.

Please verify immediately after the next rotation, not hours later:

1. Record UTC and MSK timestamps and the live journal byte size immediately after truncation.
2. Count rows available in the newest rotated archive plus the live file.
3. Inspect the first post-rotation `llm-proposed-improvement` request payload whose task matches an archive row. Its `lessons_context.reflection_hints` must be non-empty, demonstrating that the request sees archive+live rather than only the newly refilled live file.
4. Record rows seen versus rows available. The proof is that a matching archive row is visible immediately after rotation; the prior failure observed only 20 of 687 rows (3%) and could return zero hints for matching tasks.

This defect self-heals as the live journal refills, so a sample taken hours after rotation may look healthy. That is the trap, not evidence that the defect was absent.

## Changed files

- `nanobot/runtime/reflection_context.py`
- `tests/test_reflection_context.py`

Closes #1207

### DIFF
--- a/nanobot/runtime/reflection_context.py
+++ b/nanobot/runtime/reflection_context.py
@@ -1,6 +1,7 @@
 """Bounded, steering-only hints from the reflector journal (#1008, #1089)."""
 from __future__ import annotations
 
+import gzip
 import json
 import re
 from datetime import datetime, timedelta, timezone
@@ -119,9 +120,40 @@ def build_reflection_hints(
         task_words = _words(f"{task_title} {target_path}")
         if not task_words:
             return []
-        path = Path(state_dir) / "reflector" / "reflections.jsonl"
+        from nanobot.runtime import reflector
+
+        # Establish the journal order first (oldest archive to newest live
+        # file), then take one bounded tail from that ordered stream. Reading
+        # newest-to-oldest while spending the existing byte budget preserves
+        # the same newest-record semantics after a rotation without reading an
+        # entire archive into memory.
+        paths = reflector.reflection_files(state_dir)
+        tail_chunks: list[list[str]] = []
+        remaining = _MAX_TAIL_BYTES
+        for path in reversed(paths):
+            if remaining <= 0:
+                break
+            try:
+                opener = gzip.open if path.name.endswith(".gz") else open
+                with opener(path, "rb") as fh:  # type: ignore[call-arg]
+                    fh.seek(0, 2)
+                    size = fh.tell()
+                    take = min(size, remaining)
+                    fh.seek(max(0, size - take))
+                    data = fh.read(take)
+                lines = data.decode("utf-8", errors="replace").splitlines()
+                if size > take and lines:
+                    lines = lines[1:]
+                tail_chunks.append(lines)
+                remaining -= take
+            except Exception:
+                continue
+        tail_lines: list[str] = []
+        for chunk in reversed(tail_chunks):
+            tail_lines.extend(chunk)
+
         candidates: list[tuple[int, str, str]] = []
-        for line in reversed(_tail_lines(path)):
+        for line in reversed(tail_lines):
             try:
                 row = json.loads(line)
             except Exception:
--- a/tests/test_reflection_context.py
+++ b/tests/test_reflection_context.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import gzip
 import json
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
@@ -23,6 +24,30 @@ def test_selects_recent_relevant_bounded_hints(tmp_path: Path) -> None:
     hints = build_reflection_hints(tmp_path, "fix timeout bridge", now=datetime.now(timezone.utc))
     assert hints == ["new timeout approach", "other timeout error"]
     assert len(hints) <= 3 and all(len(x) <= 200 for x in hints)
+
+
+def test_reads_matching_hint_from_rotated_archive(tmp_path: Path) -> None:
+    now = datetime.now(timezone.utc)
+    ts = now.isoformat().replace("+00:00", "Z")
+    archive_row = _row("preserve archive import resolution", kind="approach_hint")
+    archive_row.update({"ts": ts, "task_title": "Fix import resolution", "target_path": "scripts/verify_imports.py"})
+    archive = tmp_path / "reflector/archive/reflections-2026-09-02.jsonl.gz"
+    archive.parent.mkdir(parents=True, exist_ok=True)
+    with gzip.open(archive, "wt", encoding="utf-8") as fh:
+        fh.write(json.dumps(archive_row) + "\n")
+    _write(
+        tmp_path / "reflector/reflections.jsonl",
+        [_row("unrelated live hint", kind="approach_hint"), _row("another live hint", kind="error_pattern")],
+    )
+
+    hints = build_reflection_hints(
+        tmp_path,
+        "Fix import resolution",
+        "scripts/verify_imports.py",
+        now=now,
+    )
+
+    assert "preserve archive import resolution" in hints
 
 
 def test_missing_corrupt_and_expired_are_empty(tmp_path: Path) -> None:


ozand:
## Validation update

CI completed successfully on Python 3.11, 3.12, and 3.13. The local full suite completed with `2883 passed, 18 failed, 17 skipped, 11 warnings`; the 18 failures are the documented Windows baseline (temporary test repositories pushing `master`, unavailable Windows `fcntl`, and Windows tag/lock expectations). No new failures are attributable to this two-file change.

Opus review verdict: **OK**, no findings. The reviewer specifically checked gzip seeking, ordering-before-bound, fail-open handling, and scope.

The branch is fully pushed: `git log origin/fix/1207-reflection-archive-hints..HEAD` is empty.

ozand:
Checked the three things that could have gone wrong here and none did.

**Reused the existing helper.** `reflector.reflection_files` rather than a second rotation walker. That was the instruction and it matters more than it sounds: #1178 already carries three of these, and a fourth would be the thing that diverges next time rotation changes.

**Order before bound, and the only bound is the pre-existing one.** `reversed(paths)` newest-first, spend `_MAX_TAIL_BYTES`, drop the partial leading line, restore chronological order. `grep` for new cap constants in the diff returns exactly one line — `remaining = _MAX_TAIL_BYTES` — reusing the 128 KiB budget that was already there. No new cliff, which is the mistake three separate fixes made this week.

**The regression test is real.** `assert 'preserve archive import resolution' in []` against `origin/main` is the defect's own signature, and it needs the rotated-archive fixture to exist at all.

Also good: naming `compute_knowledge_digest` as deliberately unchanged, with the reason. Saying what a change does *not* cover is worth as much as saying what it does — that is how the seventh reader of this file got missed in the first place.

Two files, 59 insertions, CI green on 3.11/3.12/3.13. Merging.

Live verification is mine and its shape is on #1207: rows-seen against rows-available, sampled immediately after a rotation. Recording the trap again so it is on this PR too — the defect self-heals as the journal refills toward 512 KiB, so a sample taken hours later shows it working regardless. The pre-fix baseline to beat is **20 of 687 rows, 3%**.

Issue #1211, opened 2026-09-02
36% of all cycles end in suppression, each after paying for the proposal it rejects; the worst gap shows 79 suppressed attempts as attempt_count 0


ozand:
## Problem

**35.9% of every cycle this system has ever run ended in suppression.** The suppression is correct. The cost is that it is decided after the cycle has already spent an LLM call generating the proposal it then rejects.

Measured over the live ledger plus all 52 rotated archives:

```
terminal outcomes, all time:  1700
   success            714
   skipped-duplicate  611   (35.9%)
   partial            298
   failed              77

skipped-duplicate by reason:
   recent_duplicate_failure   266
   existence_index_duplicate  182
   already_done               163

last 24 h:  15 of 96 terminal outcomes (15.6%)
```

## The mechanism, traced end to end on one demand

`reflection-a62c486e7a1f` served **eight cycles in eight hours**, all targeting `scripts/verify_imports.py`:

```
05:13  proposed  Add --ignore-modules option        ->  05:31  failed             gate_failed
08:00  proposed  Add --json output flag             ->  08:06  partial            executor_reported_skip
09:20  proposed  Add AST detection for try-except   ->  09:23  skipped-duplicate  recent_duplicate_failure
10:48  proposed  Add targeted path arguments        ->  10:52  skipped-duplicate  recent_duplicate_failure
11:11  proposed  Add module resolution cache        ->  11:16  skipped-duplicate  recent_duplicate_failure
11:40  proposed  Add --fail-fast flag               ->  11:44  skipped-duplicate  recent_duplicate_failure
12:16  proposed  Add --summary flag                 ->  12:20  skipped-duplicate  recent_duplicate_failure
12:54  proposed  Add --json flag                    ->  12:58  skipped-duplicate  recent_duplicate_failure
```

The first attempt failed. `_recent_failure_match` then correctly suppressed everything that followed. **Six cycles produced nothing, and each one first paid an LLM call to invent a fresh feature for the same script.** The demand kept being served; the proposer kept answering it with something new; the gate kept refusing it for the right reason.

Note what this is *not*: the demand does not survive its own success. Checked across all 389 demand ids — **zero** re-proposed after one of their cycles succeeded. The lifecycle is correct on that axis. The gap is that a demand whose attempt *failed* keeps being served at full price.

## Concentration, and a second finding inside it

Suppressed cycles are not spread thin:

```
goal-gap-5d4d5a9dc822   79 suppressed cycles
[...]

PR #1213, opened 2026-09-02
fix(runtime): cool recent duplicate-failure demands before LLM selection (#1211)

ozand:
## Summary

Adds a demand-level cooldown at `_select_assigned_demand`, before `build_context` and proposal generation. A demand is temporarily ineligible when its latest linked terminal outcome is `skipped-duplicate` with reason `recent_duplicate_failure` and its timestamp is within the existing `SUBAGENT_BRIDGE_FAILURE_SUPPRESS_HOURS` window (default 24 hours). When every current demand is cooled, `maybe_propose` records `proposer_reject: all_cooled` and returns without an LLM call.

This is cost avoidance, not a suppression-policy change. The existing bridge-level `_recent_failure_match` remains unchanged as the correctness backstop.

## Scope boundaries

- Addresses only `recent_duplicate_failure` (baseline: 266 of 611 suppressed duplicate outcomes).
- Does not address `already_done` (163): that is completion bookkeeping and may mark `[Done]`.
- Does not address `existence_index_duplicate` (182): that is separate existence-index semantic suppression and remains unchanged.
- Does not weaken or modify the bridge matcher or add a new size cap.

## Design and reuse

- Demand ids are linked from recorded `proposed.demand_id` rows to their cycle's recorded `outcome`; no historical id is re-derived.
- The latest outcome per demand id wins. A newer non-`recent_duplicate_failure` outcome clears the cooldown, so a transient failure does not permanently starve a demand.
- The existing suppression window is reused through the same environment variable and default (24h); the ledger read horizon is at least the existing three-day horizon and expands for a longer configured window.
- Existing `demand/rotation.json` timestamps are retained for cooled items. Once eligible again, a demand resumes its prior rotation position rather than jumping ahead as never-served.
- Ledger read errors and malformed timestamps fail open: the demand remains selectable.

## Regression evidence

The new tests were run against `origin/main` before implementation. The primary test failed as expected:

```text
FAILED tests/test_llm_proposer_rotation.py::TestSelectAssignedDemand::test_recent_duplicate_failure_cools_demand_before_selection
E       AssertionError: assert [{'affected_p...defect', ...}] == []
E       Left contains one more item: {'affected_path': '', 'evidence': '', 'id': 'defect-cooled', 'kind': 'defect', ...}
```

The final focused suite passes:

```text
227 passed  # llm proposer + rotation tests
402 passed  # proposer rotation + proposer + demand tests
```

## Validation

- Full local pytest: `2890 passed, 18 failed, 17 skipped, 11 warnings` on Windows. The 18 failures are the established Windows baseline: local temporary repositories push `master` without a created commit, Windows lacks `fcntl` for bridge lock tests, and Windows tag-permission behavior differs. No failures are in the changed files or related proposer/demand tests.
- `ruff check nanobot/runtime/llm_proposer.py tests/test_llm_proposer_rotation.py`: passed.
- `git diff --check`: passed.
- Opus review: **OK**, no remaining findings after fixing selection fallback, rotation-history retention, and all-cooled observability.

Baseline to watch: `recent_duplicate_failure` currently accounts for 266 of 611 `skipped-duplicate` terminal outcomes, with 15 of 96 terminal outcomes suppressed in the last 24 hours.

After deployment, measure the same counts over a comparable window and inspect the demand-linked sequence. Success is not merely a lower total `skipped-duplicate` count, because `already_done` and `existence_index_duplicate` are intentionally untouched. The proving signal is:

1. A demand with a recent `recent_duplicate_failure` terminal outcome is absent from the selected/assigned demand set during the 24-hour window, and no proposal-generation LLM call is recorded for that demand.
2. The same demand becomes eligible after the configured window (or after a newer non-failure outcome), and can then be selected normally.
3. Compare `recent_duplicate_failure` counts and proposal-call volume, not the aggregate duplicate count. A reduction in the 266 count and fewer proposal calls for those cooled demand ids is evidence; a drop in all duplicate reasons is not proof.
4. Confirm `already_done` and `existence_index_duplicate` counts remain separately reported rather than being reclassified.

The cooldown has a starvation risk: a transient gate failure can justify retrying sooner than 24 hours. To distinguish correct suppression from starvation, inspect the per-demand `proposed`/`outcome` chain, `demand/rotation.json`, the `all_cooled`/idle records, and whether the demand reappears after the 24-hour boundary or a newer non-failure outcome. A demand that remains absent after that boundary indicates starvation or stale evidence rather than correct cooldown.

### DIFF
--- a/nanobot/runtime/llm_proposer.py
+++ b/nanobot/runtime/llm_proposer.py
@@ -1134,6 +1134,62 @@ def _write_rotation(state_dir: Path, data: dict[str, Any]) -> None:
     os.replace(tmp_path, path)
 
 
+def _failure_suppress_hours() -> float:
+    try:
+        return float(os.environ.get("SUBAGENT_BRIDGE_FAILURE_SUPPRESS_HOURS", "24").strip() or "24")
+    except ValueError:
+        return 24.0
+
+
+def _recent_duplicate_failure_demand_ids(state_dir: Path) -> set[str]:
+    """Return demand ids with a recent ``recent_duplicate_failure`` outcome.
+
+    This is selection-time cost avoidance only. It deliberately reads the
+    recorded demand id from the proposed/outcome cycle chain and ignores the
+    other duplicate reasons: ``already_done`` is completion bookkeeping and
+    ``existence_index_duplicate`` is a separate semantic suppression. Any
+    unreadable ledger is treated as no evidence so demand remains selectable.
+    """
+    try:
+        suppress_hours = _failure_suppress_hours()
+        rows = _load_ledger_rows(
+            state_dir,
+            days=max(_LEDGER_HORIZON_DAYS, int(suppress_hours / 24) + 1),
+        )
+        demand_by_cycle = {
+            str(row.get("cycle_id") or "").strip(): str(row.get("demand_id") or "").strip()
+            for row in rows
+            if row.get("phase") == "proposed"
+            and str(row.get("cycle_id") or "").strip()
+            and str(row.get("demand_id") or "").strip()
+        }
+        cutoff = datetime.now(timezone.utc) - timedelta(hours=suppress_hours)
+        latest_by_demand: dict[str, tuple[datetime, dict[str, Any]]] = {}
+        for row in rows:
+            if row.get("phase") != "outcome":
+                continue
+            demand_id = str(row.get("demand_id") or "").strip() or demand_by_cycle.get(
+                str(row.get("cycle_id") or "").strip(), ""
+            )
+            if not demand_id:
+                continue
+            ts = _parse_ts(row.get("ts"))
+            if ts is None:
+                continue
+            previous = latest_by_demand.get(demand_id)
+            if previous is None or ts > previous[0]:
+                latest_by_demand[demand_id] = (ts, row)
+        return {
+            demand_id
+            for demand_id, (ts, row) in latest_by_demand.items()
+            if ts >= cutoff
+            and str(row.get("outcome") or "").strip().lower() == "skipped-duplicate"
+            and str(row.get("reason") or "").strip() == "recent_duplicate_failure"
+        }
+    except Exception:
+        return set()
+
+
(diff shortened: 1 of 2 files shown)


ozand:
## Validation update

- Full local pytest: `2890 passed, 18 failed, 17 skipped, 11 warnings`; the 18 failures are the established Windows baseline (temporary test-repository `master` push setup and Windows-specific `fcntl`/tag behavior). No changed-file or focused proposer/demand tests failed.
- Focused proposer/rotation/demand tests: `402 passed`.
- Opus review: approved after fixes for eligible fallback selection, rotation timestamp retention, and all-cooled observability.

Branch synchronization is verified: `git log origin/investigate/1211-suppression-cost..HEAD` is empty.

ozand:
Merging. Three things I checked rather than took from the summary.

**The scope boundary is real, not rhetorical.** `already_done` (163) and `existence_index_duplicate` (182) are named with their counts and explicitly excluded, and the verification section says a drop in the aggregate is *not* proof because those two are untouched. That is the difference between a fix and a fix that will later be credited for someone else's numbers.

**Starvation has a discriminator, not a disclaimer.** "A demand that remains absent after that boundary indicates starvation or stale evidence rather than correct cooldown" is an actual test a human can run. Most risk sections say the risk exists; this one says how to tell which side of it you are on.

**Retaining `demand/rotation.json` timestamps for cooled items** is the detail I would have missed. Without it a cooled demand returns looking never-served and jumps the queue — the fix would have created a new unfairness while removing a cost.

Also right: `proposer_reject: all_cooled` rather than a silent no-op. An idle cycle that produces no record is indistinguishable from a cycle that failed to run, which is the defect class this repo has spent the day removing.

CI green on 3.11/3.12/3.13, full suite 2890 passed / 18 failed matching baseline, the new test fails against `origin/main` with the assertion quoted.

Deploying. Live verification is mine and I will use your four criteria as written — in particular #3, comparing the 266 rather than the 611.

Issue #1209, opened 2026-09-02
Curator writes lessons.yaml into the working tree and the next cycle's reset --hard erases it — measured lifetime 2m45s


ozand:
## Problem

The reflector mint path works. Its output survives about three minutes.

`promote_reflector_recommendations_to_v2` writes `lessons/lessons.yaml` directly into the instance working tree, uncommitted. Every cycle begins with a `reset --hard HEAD`, which discards it.

Verified on the host:

```
$ git -C <instance> status --short
(empty)

$ git -C <instance> diff --stat HEAD -- lessons/lessons.yaml
(empty)

HEAD: 77843 bytes    disk: 77843 bytes
```

The working tree is byte-identical to HEAD. Anything the curator wrote is gone.

The reflog shows the mechanism as a fixed rhythm — `reset: moving to HEAD` opens every cycle:

```
487a3f9f HEAD@{09-02 16:24:04}: reset: moving to HEAD
487a3f9f HEAD@{09-02 15:58:12}: reset: moving to HEAD
5466e911 HEAD@{09-02 15:42:21}: reset: moving to HEAD
5466e911 HEAD@{09-02 15:33:43}: reset: moving to HEAD
5466e911 HEAD@{09-02 15:20:23}: reset: moving to HEAD
```

Measured window for one instance: the curator wrote at 06:18:56 MSK, `lessons.yaml` reached 78,129 bytes, and the next cycle's reset erased it at 06:21:45.97 — **2 minutes 45 seconds of life**.

## Why this went unnoticed, and why it matters more than the original defect

The four surviving v2 reflector cards, all dated `first_seen: '2026-08-29'`, are there **by accident**: a cycle's auto-commit happened to sweep them up before a reset caught them. Their existence is what made the path look like it worked at all.

This is the same signature as everything in ADR #1173, one layer further out: **success and failure produce the same observable.** `lessons.yaml` at 77,843 bytes is what "nothing was minted" looks like and also what "something was minted and destroyed" looks like. The curator's own `decisions.jsonl` records the promotion as successful, because from its side it was.

It also invalidates a measurement method I have been using all day. I verified #1183 by copying live state into a sandbox and calling the function — it promoted 2 lessons, so I reported the mechanism as working. That was true and irrelevant: a sandbox has no cycle to reset it. **The durable proof for any side-role write is `git show origin/main:<path>` — never the working tree, never a "wrote successfully" log line, never a sandbox reproduction.**

[...]

PR #1217, opened 2026-09-02
fix(runtime): stage loose lesson migration through bridge pickup (#1214)

### LINKED ISSUE #1214: migrate_loose_lessons has never run: 37 loose notes unmigrated, and a direct workspace write would live 2m45s
## Problem

`migrate_loose_lessons` has never run on the host. 37 loose `lessons/*.md` notes sit unmigrated on `origin/main`, and `lessons/archive/loose` does not exist — `lessons/archive/` holds only `lessons-2026-08-25.yaml.gz`.

This is #986's one acceptance item that was never executed, and it is the last piece of that issue that is code rather than an event.

## The route, decided

`migrate_loose_lessons` is documented operator-only because it **writes the workspace checkout directly** — and after #1209 we know exactly what that means: the next cycle's `_restore_to_main` runs `git reset --hard && git clean -fd`, so a direct write has a lifetime measured in minutes. The 2m45s figure on #1209 was measured, not estimated.

Two options were open:

1. Stop the bridge timer, run it by hand, commit and push the result.
2. Route it through the staging path that #1209 built.

**Take (2).** Reasons, in order:

- (1) requires stopping the loop, and an operator step that must be repeated is a step that will eventually be skipped. This migration is one-time, but the same function is the template for any future bulk workspace edit.
- (2) reuses a path that now exists, is tested, and whose success record means "on `origin/main`" rather than "written somewhere". #1209 made `promoted` conditional on the push returning 0.
- Consistency matters more than convenience here: after #1209 there is exactly one durable way for a side role to change the checkout. A second way, used once, is how the next person learns the wrong pattern.

## Scope

- [ ] `migrate_loose_lessons` stages its result rather than writing the checkout. Reuse the staging protocol and the manifest shape #1209 established — do not invent a second one.
- [ ] The pickup applies it at the cycle-start boundary and pushes, exactly as the fact and lesson-card paths now do. A rejected push must drop the commit and retain staging, same as #1209.
- [ ] The 37 notes end up wherever the function was always meant to put them (`lessons/archive/loose` per the current implementation — confirm from the code rather than from this sentence).
- [ ] A migration that cannot be staged is reported, not silently skipped. `return 0` and "nothing to migrate" must remain distinguishable.

## Acceptance criteria

- [ ] Test: a workspace with loose notes produces a staged manifest entry, and a simulated cycle-start reset does not lose it. Must FAIL against the pre-change tree — quote the observed failure.
- [ ] Test: the pickup applies the staged migration and the notes are present afterwards; a second run is idempotent and does not re-stage.
- [ ] `classify_change_tier` and the doc-only budget behaviour for these paths are unchanged, or the change is stated deliberately. 37 files landing at once should not silently blow a daily budget — check what it does and say so.
- [ ] Full pytest green; CI green, Linux authoritative. Windows baseline **18 failed / 2890 passed**.
- [ ] **Live verification**: `git show origin/main:lessons/archive/loose/<one known note>` succeeds after deploy and a cycle, and `git ls-tree origin/main lessons/ | grep -c '\.md$'` drops from 37. A working-tree reading does not count.

## Non-goals

No change to what a lesson *is* or to the v2 schema. No change to the reset — it is the cycle's isolation guarantee. Do not fold in the 20 orphaned curator facts; those are tagged `rescue/curator-*` and are a separate one-time recovery with a different decision attached.

## Context

From the #986 analysis: https://github.com/ozand/eeebot/issues/986#issuecomment-5511499363 — "the one #986 acceptance item never executed". The durability mechanism it depends on is #1209 / PR #1212, merged and deployed.

ozand:
## Summary

Route `migrate_loose_lessons` through the existing #1209 curator staging manifest and cycle-start pickup instead of writing directly into the shared checkout.

The migration preserves the legacy deterministic behavior:

- loose notes with identical normalized bodies produce one `memory/facts/<slug>.md` fact and one index entry;
- every source `lessons/*.md` note is staged as a move to `lessons/archive/loose/<name>.md`;
- the bridge applies the fact/index changes and source removals in one commit, pushes `origin/main`, and records promotion only after the push;
- staged data remains available across the cycle-start reset and a repeated migration after pickup is a no-op.

## Design

- Reuses the existing #1209 `_stage_promotions`, `manifest.json`, payload files, `_pickup_staged_promotions`, push, retention, and decision-record protocol.
- Adds only the manifest metadata needed for this migration: `kind: loose_lesson` and a validated `source_path`.
- The pickup accepts only exact one-level `.md` source/destination paths under `lessons/` and `lessons/archive/loose/`; traversal components are rejected.
- Curator-fact rescue (`rescue/curator-*`) is untouched and out of scope.

## Regression evidence

The new reset-survival/idempotency test was run against `origin/main` first and failed before the fix:

```text
FAILED tests/test_issue_1214_loose_lesson_staging.py::test_staged_loose_migration_survives_cycle_reset_and_is_idempotent
E       KeyError: 'ok'
```

The failure is expected: origin/main's `migrate_loose_lessons` returned the legacy result shape and directly wrote the checkout, so the staged migration contract was absent.

Final focused validation:

```text
25 passed  # staging, curator, and #1209 durability tests
3 passed   # #1214 migration tests, including traversal rejection
```

## Doc-only budget check

The migration's archive paths are under `lessons/`, so `classify_change_tier` continues to classify a batch of 37 such paths as `doc-only`. The classifier was not changed. The existing post-merge 24-hour budget guard therefore remains the governing behavior; no new cap or budget exception was added. The 37-file migration is intentionally not silently reclassified as code-bearing.

## Validation

- Full local pytest: `2900 passed, 18 failed, 17 skipped, 11 warnings` on Windows. The 18 failures are the established Windows baseline: temporary test repositories attempt to push a `master` ref without a commit, and Windows-specific `fcntl`/tag behavior differs. No failures are in the changed migration/staging tests.
- `git diff --check`: passed.
- Opus review: **OK**, including path traversal, reset survival, idempotency, source deletion, duplicate-body semantics, and scope.

After deployment and a bridge cycle, verify against the remote repository, not the working tree:

1. `git show origin/main:lessons/archive/loose/<known-note>.md` succeeds.
2. `git ls-tree origin/main lessons/ | grep -c '\.md$'` decreases from the baseline 37.
3. Confirm the resulting pickup commit is on `origin/main`, and inspect `curator/decisions.jsonl` for `staged` followed by `promoted` only after the push.
4. Confirm no `rescue/curator-*` curator facts were included.

The working-tree presence of an archive file is not evidence; the remote `git show` is the pass condition.

Closes #1214

### DIFF
--- a/nanobot/runtime/bridge.py
+++ b/nanobot/runtime/bridge.py
@@ -1814,6 +1814,7 @@ def _pickup_staged_promotions(repo_root: 'Path', state_dir: 'Path') -> int:
         changed_files: list[str] = []
         applied_lesson_ids: list[str] = []
         fact_ids: list[str] = []
+        loose_lesson_ids: list[str] = []
         # Filter unsupported entries before any validation (#1094 tier-2).
         # overlap_flag=True means zero keyword overlap between fact and support_claim;
         # these entries are kept in staging for audit but never committed to main.
@@ -1840,7 +1841,15 @@ def _pickup_staged_promotions(repo_root: 'Path', state_dir: 'Path') -> int:
             if str(entry.get('kind') or '') == LESSONS_KIND:
                 path_ok = rel == LESSONS_REL
             elif str(entry.get('kind') or '') == 'loose_lesson':
-                path_ok = rel.startswith('lessons/archive/loose/') and Path(rel).name == rel.rsplit('/', 1)[-1]
+                source_rel = str(entry.get('source_path') or '').replace('\\', '/')
+                path_ok = (
+                    rel.startswith('lessons/archive/loose/')
+                    and Path(rel).name == rel.rsplit('/', 1)[-1]
+                    and Path(rel).name.endswith('.md')
+                    and source_rel.startswith('lessons/')
+                    and Path(source_rel).name == source_rel.rsplit('/', 1)[-1]
+                    and Path(source_rel).name.endswith('.md')
+                )
             else:
                 path_ok = _fact_path(rel) is not None
             if not rel or not path_ok or not slug or Path(slug).name != slug:
@@ -1854,6 +1863,9 @@ def _pickup_staged_promotions(repo_root: 'Path', state_dir: 'Path') -> int:
                 index_rel = str(entry.get('index_rel') or '').replace('\\', '/')
                 if index_rel:
                     snapshot_paths.add(repo_root / index_rel)
+            source_rel = str(entry.get('source_path') or '').replace('\\', '/')
+            if source_rel:
+                snapshot_paths.add(repo_root / source_rel)
         snapshots = {path: path.read_bytes() if path.exists() else None for path in snapshot_paths}
 
         def _rollback_pickup() -> None:
@@ -1896,11 +1908,20 @@ def _pickup_staged_promotions(repo_root: 'Path', state_dir: 'Path') -> int:
                     changed_files.append(rel)
                     fact_ids.append(str(entry.get('lesson_id') or rel))
                 continue
-            fact_ids.append(str(entry.get('lesson_id') or rel))
+            if str(entry.get('kind') or '') == 'loose_lesson':
+                loose_lesson_ids.append(str(entry.get('lesson_id') or rel))
+            else:
+                fact_ids.append(str(entry.get('lesson_id') or rel))
             target = repo_root / rel
             target.parent.mkdir(parents=True, exist_ok=True)
             target.write_bytes(payload_path.read_bytes())
             changed_files.append(rel)
+            source_rel = str(entry.get('source_path') or '').replace('\\', '/')
+            if source_rel:
+                source = repo_root / source_rel
+                if source.exists():
+                    source.unlink()
+                    changed_files.append(source_rel)
             if action == 'create' and index_line and index_rel:
                 index_path = repo_root / index_rel
                 index_path.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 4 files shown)


ozand:
## Validation update

- Full local pytest: `2900 passed, 18 failed, 17 skipped, 11 warnings`; the 18 failures match the established Windows baseline (temporary repository `master` push setup and Windows-specific `fcntl`/tag behavior). No migration/staging tests failed.
- Focused #1214/#1209/curator tests: `25 passed`; #1214 regression tests: `3 passed`.
- The path-validation follow-up `c4cd4c7f` is pushed and included; branch synchronization is clean.
- Opus review: approved, including staging durability, path traversal, idempotency, source deletion, and legacy duplicate-body semantics.

ozand:
Reviewed and verified. The staging route is right and the test is the strong kind — it asserts `git status --porcelain == ""` so the checkout is provably untouched, performs a real `reset --hard` + `clean -fd` cycle boundary, and then proves the result with `git show origin/main:...` rather than from the working tree. That is the criterion this whole class of issue needs, and it is the one the pre-#1209 code passed while losing the data.

I pushed one commit onto this branch: `eaad3bab`, docstring only.

The module docstring still carried the pre-change contract:

> ``migrate_loose_lessons`` is an operator-only utility — run it only while the bridge timer is stopped; it writes directly into the workspace checkout and the operator must commit and push the result, or the next cycle start discards it.

Two clauses are merely stale after this PR — it no longer writes the checkout, and the operator no longer commits and pushes. The third is actively harmful: the staged batch is applied by the bridge's **cycle-start pickup**, so an operator who follows the documented procedure — stop the timer, then run the utility — leaves the migration staged and never applied. The instruction would have blocked the delivery this PR makes, and it sits in the one place an operator reads before running the thing.

Checked for the same wording elsewhere in `docs/`, `AGENTS.md` and the markdown tree: this docstring was the only occurrence.

Focused suite on the patched branch: `36 passed` across `test_issue_1214_loose_lesson_staging.py`, `test_knowledge_curator.py`, `test_issue_1209_durable_curator_writes.py`, `test_curator_staging_pickup.py`.

Two notes for the record, neither blocking:

- The red state reported for the new test is `KeyError: 'ok'`, because the old return dict had no `ok` key. That is a weaker red than the test deserves — the assertions that actually encode the defect (`status --porcelain`, the archive path, the `origin/main` reads) sit below the line that raises. Worth remembering when the next regression test is written: the first failing assertion should be the one that names the bug.
- `migrate_loose_lessons` has no automated caller; it is operator-triggered. So merging this changes nothing on the host by itself — the 37 loose notes stay unmigrated until someone runs it, now with the timer left running.

Issue #1176, opened 2026-09-02
fix(runtime): archive-aware subagent artifact readers — recent-failure suppression is dead, previous-attempts/defect/decay evidence see ≤1 hour of history


ozand:
Parent ADR: #1173. Depends on #1174 (`state_access.artifacts`). The `demand._result_file_defects` part waits for #1166 (pi-2 owns `demand.py`); everything in `bridge.py`, `existence_index.py`, `usage_evidence.py` can go first.

## Problem

`state/subagents/results/` is drained into `state/subagents/archive/` by the hourly `cleanup_subagent_queue.py --max-queue 9`. Live: `results/` 3 files (ages 18–80 min), `archive/` 3004. Five readers look only at `results/` and reason about horizons of 24–48 hours over ≤1 hour of data:

| reader | declared horizon | live evidence (2026-09-02) |
|---|---|---|
| `bridge._recent_failure_match` (`FAILURE_SUPPRESS_HOURS = 24`, `max_scan = 10`) → skip cycle as `recent_duplicate_failure` | 24 h | **260 fires all-time, 0 in 10 of the last 11 days** (08-22 → 09-01) while `skipped-duplicate` rose to 12/day, all from the existence index. In the 24 h window `archive/` holds 8 `error` + 5 `blocked` results; `results/` holds 3 in-flight files. |
| `bridge._get_previous_attempts` (`max_attempts = 3`) → `## Previous attempts for this task` prompt section | all prior attempts | same corpus — the section sees ≤1 hour of history |
| `bridge._recent_activity_context` (`[:5]`) → `## Recent activity (do not repeat)` | recent | same |
| `demand._result_file_defects` (`_DEFECT_WINDOW_HOURS = 48`, `_MAX_RESULT_FILES = 50`, statuses `failed/blocked/error`) → `defect` demand items | 48 h | `archive/` files modified within 48 h: **456** (error 31, blocked 17, bounded_stop 27 …); `results/`: 3. Sibling code in the same module (`demand.py:2280`, `:2515`) already uses `_RESULT_DIRS = ("results", "archive")`. |
| `usage_evidence._touched_from_results` (`_MAX_RESULT_FILES = 50`) → `last_touched` → `stale_artifacts` → **decay lane (deletion proposals)** | 50 newest results | ≤3 files; the 50-cap is dead weight; mtime sort has a bare `except: pass` that degrades to glob order |

[...]

PR #1218, opened 2026-09-02
fix(existence_index): a refused proposal's title is not evidence the thing exists (#1215)

### LINKED ISSUE #1215: The existence index treats refused proposals as proof the thing exists: 75 of 83 duplicate-suppressions matched an attempt that produced nothing
## Problem

The existence index treats the **title of a refused proposal** as evidence that the thing exists. The next attempt at the same subject is then refused as a duplicate of it. The first refusal seeds every later one, and nothing ever retires the entry.

`existence_index` ingests `subagents/results/*.json` titles as `ledger_title` documents with **no outcome filter** (`existence_index.py:345-380`) and never deactivates them. A `ledger_title` document asserts that an *attempt happened*, not that an *artifact exists*.

Measured on the live ledger plus all rotated archives:

```
existence_index_duplicate rows:      182
   matched an existing script:        99   (correct — see non-goals)
   matched a past proposal title:     83

of those 83, what happened to the proposal they matched:
   itself skipped as existence_index_duplicate   44
   itself skipped as recent_duplicate_failure    30
   actually integrated                            8
   partial                                        1

since 2026-08-26:  17 of 19 matched a proposal that was itself eid-suppressed
last 14 days:      50 eid rows, all 50 test-for proposals
same target_path as the suppressed one:  76 of 83
```

**75 of 83 times the matched attempt produced nothing.** The index refused a proposal because an earlier proposal for the same thing had been refused.

## Refused as a duplicate of nothing

25 of the 83 refused a proposal whose target does not exist on `origin/main` and whose matched attempt was never integrated. Verified directly:

```
$ git cat-file -e origin/main:tests/test_verify_and_proof.py   ->  ABSENT
$ git log origin/main --oneline -- tests/test_verify_and_proof.py | wc -l   ->  0
```

Never created, never touched by any commit — and the proposal to create it has been suppressed **4 times** since 2026-08-16 as a duplicate.

The stuck list since 2026-08-16, as suppressions / integrations on that target:

```
tests/test_summarize_failure_reasons.py   7 / 0
tests/test_lessons_integrity.py           7 / 2
tests/test_prevent_repeat_failures.py     6 / 0
tests/test_analyze_repeat_failures.py     5 / 1
tests/test_check_confirmation_ratio.py    4 / 1
tests/test_verify_and_proof.py            4 / 0
```

Where the file does exist with 0 integrations on target, it arrived as a side file of some unrelated cycle — the "duplicate" verdict was still reasoned from a refused title, not from that file.

## Why it does not self-clear

The suppression is deliberately **not** completion (`bridge.py:2496`), so the demand stays live and returns. `skipped-duplicate` is in `_NOOP_OUTCOMES` (`demand.py:246`), so two in 24 h exhaust a demand id — but goal-gap ids are re-presented after their TTL and the chain re-fires:

```
demand ids carrying an eid suppression:                45
  of those, proposed again afterwards:                 43
  later cycles suppressed again:                      100
  goal-gap-5d4d5a9dc822 alone:                         49
```

So a subject that is refused once becomes permanently unreachable, while the demand that wants it keeps consuming cycles.

## The fix, as far as the evidence supports it

- [ ] A `ledger_title` hit counts as a duplicate **only** when the attempt it names actually integrated — its cycle has an `outcome: success` in the ledger — or when the target artifact exists on `origin/main`. A refused title is not evidence of existence.
- [ ] Consider retiring `ledger_title` documents whose attempt did not integrate, rather than leaving them to accumulate. Today nothing is ever deactivated.
- [ ] A test-for demand whose target file already exists on `origin/main` can be skipped at selection time, before the LLM call. This is the cheap half and needs no index. Note it does **not** fix the chain — an earlier check with the same rule refuses the same non-duplicates sooner.

## Acceptance criteria

- [ ] Failing fixture, from the live case: a `ledger_title` document for a refused attempt at `tests/test_verify_and_proof.py`, target absent on `origin/main`, and a new proposal for the same subject. It must currently be suppressed and must not be after the fix. Quote the pre-change failure.
- [ ] The script-match half is unchanged: a proposal duplicating an **existing script** is still refused. Test both directions or the fix can pass by never suppressing anything.
- [ ] An integrated attempt still suppresses a repeat. That is the case the rule is for.
- [ ] Full pytest green; CI green. Windows baseline **18 failed / 2890 passed**.
- [ ] **Live verification**: after deploy, one of the stuck targets above reaches a non-suppressed outcome, or the `existence_index_duplicate` count for test-for proposals drops from its 14-day rate of ~4/day. Quote the count and the window.

## Non-goals

The script-match half (99 rows, mostly July–August) is **correct and unavoidable** at proposal time — the proposer already receives `existence_index.related_scripts` in its prompt (`llm_proposer.py:938-980`) and proposes a near-duplicate anyway; the post-check is the backstop. Do not touch it.

No change to `_recent_failure_match` or to the #1211 cooldown; `existence_index_duplicate` stays out of failure history, correctly.

`already_done` is **not** in scope and should be removed from #1211's accounting: all 163 rows fall between 2026-07-12 and 2026-07-15, zero since, and the branch became unreachable on 2026-07-15 when demand-serving proposals began skipping the heuristic. Current cost: 0 cycles/day.

## Context

From the #1211 analysis: https://github.com/ozand/eeebot/issues/1211

ozand:
Closes #1215.

## Defect

`existence_index` minted a `ledger_title` document from every `subagents/results/*.json` title with no outcome filter and never deactivated one. A `ledger_title` asserts that an *attempt happened*, not that an *artifact exists*, so a proposal was refused as a duplicate of an earlier proposal that was itself refused. Live: 75 of 83 `ledger_title` suppressions matched an attempt that produced nothing; `tests/test_verify_and_proof.py` is absent on `origin/main`, touched by zero commits, and was refused 4 times since 2026-08-16 as a duplicate of it.

## Fix (existence_index.py, `_reindex_ledger_titles`)

- A `ledger_title` is indexed **only** when the attempt behind it is evidence of an artifact:
  - its result carries `rollback.integrated: true` (written in the same bridge step as the ledger's `outcome: success` row), **or**
  - the result predates the rollback record and the ledger has `outcome: success` for its `cycle_id` — new `cycle_ledger.successful_cycle_ids()`, read across every `cycles-*.jsonl.gz` archive plus the active file (rotation would otherwise narrow this reader to today), **or**
  - the attempt's own `target_path` exists on `origin/main` (`git ls-tree -r --name-only origin/main`, one lazy subprocess per reindex and only when a non-integrated attempt names a target; falls back to the working tree when git is unavailable, e.g. the test tmp dirs).
- **Retirement — yes, done.** Every active `ledger_title` not re-verified as evidence in the pass is deactivated via the existing `_deactivate_missing` mechanism the `script` corpus uses for deleted files. Why: the live index already holds the poisoned documents and their result JSONs have migrated out of `results/`; without retirement the fix would apply to new refusals only and the 83 existing seeds would keep refusing. An attempt that can no longer be found in `results/` or `archive/` is treated as unknown, not as proof — an integrated artifact stays covered by the `script` corpus because the file exists. A retired document is reactivated if its attempt later integrates (tested).
- Results are read from `results/` **and** `archive/result-*.json`, newest-first, under the **existing** `_MAX_LEDGER_RESULTS` bound (no new cap). Results migrate to `archive/` within the hour (#1176), so the live dir alone no longer holds the history the retirement rule needs to classify. Same `request_id` in both dirs is classified once (newest copy wins). `request-*.json` files in the flat archive are skipped.
- New reindex counts: `ledger_titles_not_integrated`, `ledger_titles_deactivated`.

## Untouched (non-goals)

`find_similar`, `find_duplicate_script`, the script-match half, `_recent_failure_match`, the #1211 cooldown, `bridge.py`. `git diff origin/main...HEAD` touches three files: `existence_index.py`, `cycle_ledger.py`, `tests/test_existence_index.py`.

## Failing fixture (pre-change, quoted)

`TestRefusedTitleIsNotExistence::test_refused_attempt_title_does_not_suppress_repeat` — a `ledger_title` for a refused attempt at `tests/test_verify_and_proof.py` (`result_status: blocked`, `rollback.integrated: false`, `reason: existence_index_duplicate`), target absent, new proposal for the same subject:

```
>       assert matched is None, (
            f"refused attempt {matched!r} was treated as proof the target exists"
        )
E       AssertionError: refused attempt 'req-refused' was treated as proof the target exists
E       assert 'req-refused' is None

tests\test_existence_index.py:602: AssertionError
9 failed, 5 passed, 36 deselected in 2.77s
```

The 5 that passed pre-change are the other-direction guards (integrated still suppresses, existing script still refused).

## Both directions tested

| direction | test |
|---|---|
| refused title, target absent → **not** suppressed | `test_refused_attempt_title_does_not_suppress_repeat`, `test_refused_attempt_is_not_indexed_and_is_counted` |
| integrated attempt → **still** suppressed | `test_integrated_attempt_still_suppresses_repeat` |
| refused title but target now exists → still suppressed | `test_refused_attempt_whose_target_now_exists_still_suppresses` |
| rollback-less result + ledger success / non-success | `test_result_without_rollback_uses_ledger_outcome_success`, `test_result_without_rollback_and_non_success_ledger_row_is_not_evidence` |
| ledger success only in a rotated `.gz` | `test_ledger_success_in_rotated_archive_is_read` |
| pre-existing poisoned document retired on first reindex | `test_preexisting_refused_title_document_is_retired`, `test_title_document_with_no_surviving_result_is_retired`, `test_retired_document_is_reactivated_if_the_attempt_later_integrates` |
| archive read; request files ignored; dedupe by request_id | `test_integrated_result_in_archive_still_suppresses`, `test_same_request_in_results_and_archive_is_classified_once` |
| script-match half unchanged | `test_existing_script_still_refused_with_refused_title_present`, `test_existing_ordinary_script_still_refused` (+ the existing #757/#798 tests) |

Three existing fixtures that seeded a bare `{"request_id","backlog_title"}` result now mark the attempt `rollback.integrated: true` — they assert that a prior title suppresses, which under the new rule requires the attempt to have integrated.

## Full pytest (Windows, `-n auto`)

| | failed | passed |
|---|---|---|
| origin/main `ee8925a4` | 18 | 2897 |
| this branch | 18 | 2911 |

Identical failing set (autoevolve symlink/tar, bridge flock, `master` refspec — Windows environment classes). +14 = the new tests.

## Live verification (yours, post-deploy)

First reindex after deploy should log a non-zero `ledger_titles_deactivated`. Then: one of the stuck targets (`test_verify_and_proof`, `test_summarize_failure_reasons`, `test_prevent_repeat_failures`, `test_analyze_repeat_failures`) reaches a non-suppressed outcome, or `existence_index_duplicate` for test-for proposals drops from ~4/day over the 14-day window.

### DIFF
--- a/nanobot/runtime/cycle_ledger.py
+++ b/nanobot/runtime/cycle_ledger.py
@@ -307,6 +307,46 @@ def read_events(state_dir: Path) -> list[dict]:
     return rows
 
 
+def successful_cycle_ids(state_dir: Path) -> set[str]:
+    """Return the ``cycle_id`` of every terminal row with ``outcome: success``,
+    read ACROSS the rotation: every ``cycles-YYYY-MM-DD.jsonl.gz`` archive
+    plus the active file. Best-effort — never raises; an unreadable archive
+    or malformed line is skipped (#1215).
+
+    Unlike :func:`read_events`, this answers a question that is NOT
+    same-day: "did the cycle behind this result artifact ever integrate?"
+    Result artifacts outlive the day they were written (they migrate to
+    ``subagents/archive/`` and stay readable for hundreds of cycles), so a
+    reader that opened only the active file would call every integrated
+    attempt older than today "never integrated" (#1178/#1207: rotation
+    narrows every reader that only opens the live file).
+    """
+    ids: set[str] = set()
+
+    def _collect(lines) -> None:
+        for line in lines:
+            with contextlib.suppress(Exception):
+                row = json.loads(line)
+                if (
+                    isinstance(row, dict)
+                    and row.get("phase") == "outcome"
+                    and row.get("outcome") == "success"
+                    and row.get("cycle_id")
+                ):
+                    ids.add(str(row["cycle_id"]))
+
+    ledger_dir = _ledger_dir(state_dir)
+    with contextlib.suppress(Exception):
+        for gz_path in sorted(ledger_dir.glob("cycles-*.jsonl.gz")):
+            with contextlib.suppress(Exception):
+                with gzip.open(gz_path, "rt", encoding="utf-8") as fh:
+                    _collect(fh)
+    with contextlib.suppress(Exception):
+        with open(ledger_dir / _LEDGER_FILENAME, encoding="utf-8") as fh:
+            _collect(fh)
+    return ids
+
+
 def record_explore_started(state_dir: Path, cycle_id: str, candidates_count: int, declared_measurement: str) -> None:
     append_event(state_dir, {
         'phase': 'explore_started',
--- a/nanobot/runtime/existence_index.py
+++ b/nanobot/runtime/existence_index.py
@@ -42,12 +42,19 @@ Indexed corpus (built by :func:`reindex`)
   into spaces, plus the first line of the module docstring (``ast``,
   best-effort; falls back to the first ``#`` comment line if the file
   doesn't parse). ``path`` = the path relative to ``selfevo_repo``.
-- **ledger_titles**: titles of past subagent attempts, read from
-  ``<state_dir>/subagents/results/*.json`` (bounded to the 500
-  most-recently-modified files — this is the durable, title-bearing record;
+- **ledger_titles**: titles of past subagent attempts that PRODUCED
+  something, read from ``<state_dir>/subagents/results/*.json`` plus the
+  ``result-*.json`` files of ``<state_dir>/subagents/archive/`` (results
+  migrate there within the hour, #1176; bounded to the 500 most-recently-
+  modified files across both — this is the durable, title-bearing record;
   ``cycles.jsonl`` itself never carries ``task_title``, only cycle/phase
-  bookkeeping, so the results directory is the actual source of past
-  proposal titles). ``path`` = the request id.
+  bookkeeping). ``path`` = the request id. Since #1215 a title is indexed
+  only when its attempt integrated (``rollback.integrated``, or the ledger's
+  ``outcome: success`` for results predating the rollback record) or its
+  ``target_path`` exists on ``origin/main`` — a refused proposal's title
+  asserts an attempt, not an artifact, and indexing it let the first refusal
+  of a subject suppress every later one. Documents that no longer qualify
+  are deactivated on reindex, like deleted scripts.
 - **hypotheses**: titles from ``<state_dir>/hypotheses/backlog.json``
   (``entries[].task_title``) and ``<state_dir>/research/hypotheses.json``
   (``[].candidates[].title``). Both optional; each fails open independently.
@@ -102,6 +109,7 @@ import json
 import os
 import re
 import sqlite3
+import subprocess
 from pathlib import Path
 from typing import Any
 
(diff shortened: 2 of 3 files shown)


ozand:
Verified independently and merging.

**Red confirmed on `origin/main`** with your test file applied to the unfixed tree — 8 failures, including the headline one with the assertion you quoted:

```
FAILED TestRefusedTitleIsNotExistence::test_refused_attempt_title_does_not_suppress_repeat
FAILED TestRefusedTitleIsNotExistence::test_refused_attempt_is_not_indexed_and_is_counted
FAILED TestRefusedTitleIsNotExistence::test_preexisting_refused_title_document_is_retired
FAILED TestRefusedTitleIsNotExistence::test_title_document_with_no_surviving_result_is_retired
FAILED TestRefusedTitleIsNotExistence::test_retired_document_is_reactivated_if_the_attempt_later_integrates
FAILED TestRefusedTitleIsNotExistence::test_integrated_result_in_archive_still_suppresses
FAILED TestRefusedTitleIsNotExistence::test_same_request_in_results_and_archive_is_classified_once
FAILED TestRefusedTitleIsNotExistence::test_result_without_rollback_and_non_success_ledger_row_is_not_evidence
8 failed, 42 passed
```

Green on your head: `50 passed`.

**Full suite, both sides, same machine, `-p no:randomly`:**

| | failed | passed |
|---|---|---|
| `origin/main` | 18 | 2900 |
| `dc0d2779` | 18 | 2911 |

Failure sets are **identical** — `comm` against the baseline list returns nothing in either direction. Zero regressions, zero incidental fixes, +11 tests.

Worth recording: an earlier head produced 19. You found and fixed it yourself in `dc0d2779` before I got the list, which is the outcome I want from the full-suite-on-central-modules rule. `test_bridge_cycle_tags.py::test_fail_open_tag_failure_never_breaks_a_green_cycle` is in the baseline 18 on `origin/main` and is not yours.

**The retirement decision was the right call.** Measured on the live index tonight while auditing something else:

```
documents total = 8727   active = 8727
  hypothesis      6930   ledger_title  1451   script  346
```

Nothing in the whole index had ever been deactivated. Your reasoning — *"without retirement the fix would apply to new refusals only and the 83 existing seeds would keep refusing"* — is exactly right, and after this merge two of three corpora have a retirement path.

The third does not, and it is the largest: `hyp-research-*` is 6,214 documents, 71% of the index, of which **6,114 (98.4%) correspond to nothing in any file on the host**. That is #1219, not this PR — I confirmed `_deactivate_missing` is called for `script` (line 347) and now `ledger_title` (line 501) and never for `hypothesis`, so your scope statement was accurate. Nothing owed here.

Merging.

Issue #1171, opened 2026-09-02
reflector lesson minting is dormant: 249 reflection-derived proposals in 8 days, 4 lessons total and all from one day


ozand:
## Problem

The reflector lesson-minting path has produced nothing since 2026-08-29. The entire v2 lesson population is four cards, all minted on that one day, all carrying identical filler.

Meanwhile the reflection *demand* lane is busy: 249 `proposed` ledger rows derived from reflection over the last 8 days, the most recent at 2026-09-01 23:17 UTC. So the loop reflects constantly and mints almost never.

## Evidence

From `/var/lib/eeepc-agent/self-evolving-agent/eeebot-self-evolving/lessons/lessons.yaml` on the host:

```
$ grep -cE '^\s*-?\s*id:' lessons.yaml
108
$ grep -oE "first_seen: '[0-9-]+'" lessons.yaml | sort -u
first_seen: '2026-08-29'
$ grep -A1 'tags:' lessons.yaml | grep -oE '^\s*- [a-z_]+' | sort | uniq -c
      4   - reflector
```

`first_seen` and `tags` exist only on v2 records, so those two outputs describe the whole v2 population: **four records, one date**. The other 104 entries are legacy rows without those fields.

All four carry the same solution string:

```
id: LESS-REF-c871bf9abe41        solution: Apply the reflected approach hint.
id: LESS-REF-c871bf9abe41-8608   solution: Apply the reflected approach hint.
id: LESS-REF-2782e96c24a8        solution: Apply the reflected approach hint.
id: LESS-REF-5b349fbfddd0        solution: Apply the reflected approach hint.
```

The store itself is alive — `lessons.yaml` was last written 2026-09-01 19:05:42 UTC — so this is not a stuck writer.

Reflection-derived demand over the same window, counted from the live ledger plus the 8 newest `.gz` archives (3791 rows):

| phase | rows mentioning reflection |
|---|---|
| proposed | 249 |
| proposer_skip | 34 |
| proposer_reject | 10 |

Newest: `2026-09-01T23:17:24Z`.

## Why this matters more than it looks

#1106 fixed the *content* of minted reflector lessons — filling the v2 solution from the finding text and rejecting template filler. That fix merged as 46f0c2ff and deployed. Its live criterion was "a reflector lesson minted after 46f0c2ff carrying a real recommendation, or one of the four filler cards upgraded in place".

[...]

PR #1220, opened 2026-09-02
feat(runtime): expose curator mint stage outcomes (#1216)

### LINKED ISSUE #1216: Curator mint path has no signal: a dry run and a broken run both print {"ok": true}
## Problem

The v2 reflector mint path stopped producing on 2026-08-29 and ran dry for four days. Nothing reported it. The only evidence that it had stopped was the `first_seen` date sitting on the four surviving cards, and that was read by a person going looking for it — after the cause (#1183) had already been found by other means.

This is not a restatement of #1183. #1183 was one cause; this is about why any cause on this path is invisible.

Two properties combine:

1. **A failure and a healthy quiet day emit the same thing.** `run_curation` returns `{"ok": true, "processed": N, "writes": M}` and `promote_reflector_recommendations_to_v2` returns an integer. Zero minted is the correct output when there is nothing worth minting, and it is also the output when the reader is switched off, the store is unreadable, or every candidate is rejected. `{"ok": true}` is printed in all of those cases.
2. **The cadence is daily**, so even a perfect detector would have a 24-hour floor:

```
eeebot-knowledge-curator.timer -> eeebot-knowledge-curator.service   (daily, 00:00 MSK)
last run: Wed 2026-09-02 06:18:56 MSK
{"ok": true, "processed": 40, "writes": 3, "staged": [...3 facts...]}
```

Measured on the live host: `decisions.jsonl` holds 565 rows and **zero** with `decision: "staged"` — the marker `_stage_lesson_cards` writes. So the mint path has not staged a single card since #1209 moved it to staging, and no artifact anywhere says so.

Meanwhile the material is not the problem. Against the current live store:

```
rows in file=65  tail considered=50  parsed=50  unparseable=0
recommendation kinds: {'approach_hint': 23, 'instruction_change': 1}
PROMOTABLE items (kind ok + detail non-empty): 23
```

## Why it matters beyond this path

#1106 fixed the *content* of minted reflector lessons and measured zero knowledge lift. The reason was that nothing was being minted — a quality fix aimed at an event that does not occur. That misattribution cost a full issue cycle, and it was possible only because "the event stopped happening" had no observable.

## Acceptance criteria

- [ ] Each curator run emits a per-stage outcome, not one aggregate `ok`. At minimum: rows read from the reflector store, candidates that passed the kind/detail filter, cards rejected by `validate_lesson_for_mint`, cards folded as duplicates, cards staged. Zero staged with a non-zero candidate count must be visibly different from zero staged with a zero candidate count.
- [ ] The counts land somewhere durable and queryable after the process exits — `decisions.jsonl`, the ledger, or a sidecar — not only on stdout into the journal.
- [ ] A "last successful mint" timestamp is recorded and readable, so *when it stopped* is answerable without inferring it from `first_seen` on surviving cards.
- [ ] Unreadable/absent store stays distinguishable from an empty one at this level too, consistent with what #1183 established inside the function.
- [ ] Decide and state whether the daily cadence is right for this path. If it stays daily, say so with the reason and accept the 24h floor explicitly; if it moves, name the new trigger. Do not change the cadence silently as a side effect of adding signal.
- [ ] Full pytest green; CI green (Linux matrix is authoritative).
- [ ] Live verification: quote one real post-deploy curator run's emitted counts from the host, including a run where nothing was staged, and show that the reason is legible from the record alone.

## Notes

Do not add a threshold or an alert that fires on "zero staged" — on a genuinely quiet day zero is correct, and an alarm that cries on correct behaviour gets muted and then the path is invisible again for a different reason. The requirement is that the record distinguish the cases, not that something shouts.

`knowledge_curator.py` is a central runtime module: run the full pytest suite, not the touched files. Work in your own worktree, never `T:\Code\eeebot`.

Related: #1183 (one cause), #1209 (staging durability, closed), #1106 (the misattributed zero-lift result), and the #1173 ADR, which this belongs under.

ozand:
## Summary

Re-scope #1216 on top of merged PR #1221 (#1171). The reflector mint keeps #1221's recurrence/pooling implementation, integer return value, durable `load_reflector_pool(state_dir)["last_run"]` counts, `last_staged_at`, and existing per-item counters.

This residual adds:

- durable reflector store state (`absent`, `present`, `stale`, or `unreadable`) in the existing reflector sidecar's `last_run`;
- the same store state on the emitted `curator-reflector:` line;
- the four #1221 sidecar fields that were not previously printed: `rows_after_cursor`, `unparseable`, `evicted`, and `at`;
- `run_curation`'s separate fact-curation stage under `stages.curation`, with `status`, `processed`, `writes`, and `staged` on empty, success, and error paths;
- durable `state/curator/status.json` stage records and `last_success_ts` for the whole curator run.

`last_success_ts` remains distinct from the reflector sidecar's `at`/`last_staged_at`: it records that the combined curator pass completed successfully, including a pass with no fact entries, while the reflector sidecar timestamps and describes only the mint stage.

## Explicitly dropped after #1221

- the reflector function return-type change (it still returns `int`);
- `mint_succeeded`;
- `last_successful_mint_ts`;
- duplicate candidate/rejected/folded counters and other reflector counts already supplied by #1221.

## Design decisions

- No alert or zero-staged threshold was added.
- Reflector recurrence/pool behavior, caps, rotation, digest, and daily cadence remain unchanged.
- Host rollout and live verification remain operator-owned.

## Validation

- Focused residual and related curator suites: `81 passed` during the final local run before the last test-only assertion refinement; the current signal file was then rechecked separately with `6 passed`.
- Full local pytest: `2938 passed, 18 failed, 17 skipped, 11 warnings` on Windows. The 18 failures are the established Windows baseline: autoevolve ×12, bridge_locking ×3, selfevo_export ×2, and `test_fail_open_tag_failure_never_breaks_a_green_cycle`.
- CI on the pushed head `68ddd4cd`: Python 3.11, 3.12, and 3.13 all green.

## Changed files

- `nanobot/runtime/knowledge_curator.py`
- `tests/test_issue_1216_curator_signal.py`

After deployment, inspect one real curator run and its durable sidecars. Confirm the reflector line/sidecar can distinguish `absent`, `present`, `stale`, and `unreadable`; confirm `stages.curation` reports the fact stage; and confirm `last_success_ts`, `last_run.at`, and `last_staged_at` are readable with their separate scopes.

Closes #1216

### DIFF
--- a/nanobot/runtime/knowledge_curator.py
+++ b/nanobot/runtime/knowledge_curator.py
@@ -1112,6 +1112,55 @@ def _prune_pool(clusters: list[dict[str, Any]], newest_ts: str) -> int:
     return before - len(clusters)
 
 
+def _empty_reflector_run(store: str, pool_size: int = 0) -> dict[str, Any]:
+    """Build a durable zero-count reflector outcome for unavailable input."""
+    return {
+        "store": store,
+        "rows_read": 0,
+        "rows_after_cursor": 0,
+        "rows_processed": 0,
+        "items": 0,
+        "folded": 0,
+        "repaired": 0,
+        "pooled_new": 0,
+        "pooled_recurrence": 0,
+        "same_day_only_waiting": 0,
+        "graduated": 0,
+        "deferred_by_cap": 0,
+        "near_misses": 0,
+        "rejected": 0,
+        "staged": 0,
+        "evicted": 0,
+        "unparseable": 0,
+        "pool_size": pool_size,
+        "at": _now(),
+    }
+
+
+def _persist_reflector_run(state_dir: Path, pool: dict[str, Any], stats: dict[str, Any], cursor: str = "") -> None:
+    """Persist and print one reflector outcome, including early-return states."""
+    pool["last_run"] = stats
+    try:
+        _atomic_json(_reflector_pool_path(state_dir), pool)
+    except OSError as exc:
+        import logging as _logging
+
+        _logging.getLogger(__name__).warning(
+            "promote_reflector_recommendations_to_v2: cannot write %s: %s",
+            _reflector_pool_path(state_dir), exc,
+        )
+    print(
+        "curator-reflector: "
+        + " ".join(f"{k}={stats[k]}" for k in (
+            "store", "rows_read", "rows_after_cursor", "rows_processed", "unparseable",
+            "items", "folded", "repaired", "pooled_new", "pooled_recurrence",
+            "same_day_only_waiting", "graduated", "deferred_by_cap", "near_misses",
+            "rejected", "staged", "evicted", "pool_size", "at",
+        ))
+        + f" cursor={cursor or '-'}"
+    )
+
+
 def promote_reflector_recommendations_to_v2(
     workspace: Path, state_dir: Path, *, max_items: int = 5,
 ) -> int:
@@ -1143,6 +1192,7 @@ def promote_reflector_recommendations_to_v2(
         state_dir / "state" / "reflector" / "reflections.jsonl",
         state_dir / "reflections.jsonl",
     ]
+    pool = load_reflector_pool(state_dir)
     path = next((p for p in candidates if p.is_file()), None)
     if not path:
         # Fallback glob for nested state paths
@@ -1150,8 +1200,9 @@ def promote_reflector_recommendations_to_v2(
         if matches:
             path = matches[0]
         else:
+            stats = _empty_reflector_run("absent", len(pool["clusters"]))
+            _persist_reflector_run(state_dir, pool, stats)
             return 0
-    pool = load_reflector_pool(state_dir)
     try:
         stat = path.stat()
         # The staleness guard stays. A reflector log untouched for 90 days
@@ -1161,7 +1212,14 @@ def promote_reflector_recommendations_to_v2(
         # 512 KiB size cap #1183 removed, which an append-only file could only
         # trip once and which silently switched this path off for four days.
         if time.time() - stat.st_mtime > _REFLECTOR_MAX_AGE_SECONDS:
+            stats = _empty_reflector_run("stale", len(pool["clusters"]))
+            _persist_reflector_run(state_dir, pool, stats, str(pool.get("cursor") or ""))
             return 0
+        # Probe the live file itself: _reflection_lines intentionally skips
+        # unreadable archive members, but the live store must remain
+        # distinguishable from a readable empty store.
+        with path.open("rb"):
+            pass
         rows, counts = _reflector_rows_after(path, str(pool.get("cursor") or ""), _REFLECTOR_MAX_ROWS_PER_RUN)
     except OSError:
         # Unreadable is not the same as "nothing to promote" (#1183). Both
@@ -1173,6 +1231,8 @@ def promote_reflector_recommendations_to_v2(
             "promote_reflector_recommendations_to_v2: cannot read %s; "
             "no promotions attempted (unreadable, not empty)", path,
         )
+        stats = _empty_reflector_run("unreadable", len(pool["clusters"]))
+        _persist_reflector_run(state_dir, pool, stats, str(pool.get("cursor") or ""))
         return 0
     if counts["unparseable"]:
         import logging as _logging
@@ -1183,6 +1243,7 @@ def promote_reflector_recommendations_to_v2(
         )
     stats: dict[str, Any] = dict(counts)
     stats.update({
+        "store": "present",
         "rows_processed": len(rows), "items": 0, "folded": 0, "repaired": 0,
         "pooled_new": 0, "pooled_recurrence": 0, "same_day_only_waiting": 0,
         "graduated": 0, "deferred_by_cap": 0, "near_misses": 0, "rejected": 0,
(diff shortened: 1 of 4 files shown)


ozand:
## Validation update

- CI: Python 3.11, 3.12, and 3.13 all green.
- Full local pytest: `2903 passed, 18 failed, 17 skipped, 11 warnings`; the 18 failures match the established Windows baseline and are unrelated to this change. No changed signal/curator tests failed.
- Focused signal and related curator/rotation durability tests: `58 passed`.
- Ruff was run on the changed runtime/test paths and passed; `git diff --check` passed.
- Opus review: OK, with no findings.
- Branch synchronization is clean: no commits remain ahead of `origin/feat/1216-curator-mint-signal`.

ozand:
## Validation update

The branch was rebased onto the current `origin/main` after #1218 landed and force-with-lease pushed. The PR head is now synchronized with the rebased branch.

- Focused signal/curator/rotation durability tests: `58 passed`.
- Full local pytest: `2903 passed, 18 failed, 17 skipped, 11 warnings`; the 18 failures remain the established Windows baseline and are unrelated to this change.
- `git diff --check`: passed.
- Opus review: OK.

The PR body claims were checked against `git diff origin/main...HEAD` for stage counts, store states, status sidecar, cadence/no-alert scope, and the new regression tests; each claimed change is present.

ozand:
Reviewed. The stage counts are right — the candidate/rejected/folded/staged split is exactly what the issue asked for, the absent/present/stale/unreadable distinction is clean, and you correctly did not add an alert on zero. One defect blocks merge, and it is this issue's own failure mode reproduced inside the fix.

```python
result["rows_read"] = len(lines)
result["mint_succeeded"] = True        # set before a single card is minted
```
```python
if reflector_result.get("mint_succeeded"):
    status["last_successful_mint_ts"] = _now()
```

`last_successful_mint_ts` therefore advances on every run where the store was **readable**, minted or not. The acceptance criterion for that field is *"so that when it stopped is answerable without inferring it from `first_seen` on surviving cards"* — a timestamp that advances when nothing is minted reports "now" forever and can never answer it.

It survives the #1183 case by luck: that path returns early, so the flag stays False. It fails for four of the five ways minting yields nothing — every candidate rejected by `validate_lesson_for_mint`; every candidate folded as a duplicate; zero candidates in the window (the reach ceiling, #1171); `max_items` reached with no successful merge. In each, the store reads fine, the flag is True, the timestamp says now, and no card exists.

**The test locks it in rather than guarding it.** :

```python
assert empty["mint_succeeded"] is True     # empty store, nothing minted
```

and line 64 asserts only that `status["last_successful_mint_ts"]` is truthy — never that it must not advance.

Requested: keep both concepts rather than collapsing them. `mint_succeeded` stays as "the stage ran without error" (real, worth reporting, current placement correct — consider `stage_ok` so the name stops implying a mint). Derive `last_successful_mint_ts` from `staged > 0`. Test: seed a known old timestamp, run over a readable store that mints nothing, assert the timestamp is **unchanged**; then the positive direction. Mutate the fix and quote the failing output.

Separately, on the red: `TypeError: 'int' object is not subscriptable` is an honest consequence of the return-type change, not a weak fixture, and I am not asking you to contort it. But when the API change is what fails first, say so in the PR body and quote the assertion that would have failed on the old *semantics* — otherwise "it failed before the fix" carries no information about the defect.

ozand:
Correction to the previous comment — one file reference was eaten by my own shell before it reached GitHub. The sentence "**The test locks it in rather than guarding it.**" should name its evidence:

`tests/test_issue_1216_curator_signal.py:52`

```python
assert empty["mint_succeeded"] is True     # empty store, nothing minted
```

That is the line asserting a successful mint on a store from which nothing was minted, and line 64 asserts only that `status["last_successful_mint_ts"]` is truthy. Everything else in that comment stands as written.

ozand:
## Re-scope and validation update

#1221 is merged into `origin/main`; this branch was rebased onto it and now retains only the residual #1216 scope.

- `promote_reflector_recommendations_to_v2` still returns `int`; #1221 remains the source of reflector counts, recurrence/pooling, `last_run`, and `last_staged_at`.
- Added durable and printed reflector store state: `absent`, `present`, `stale`, `unreadable`.
- Added `rows_after_cursor`, `unparseable`, `evicted`, and `at` to the existing `curator-reflector:` line.
- Added durable `stages.curation` with `status`, `processed`, `writes`, and `staged` for empty, success, and error fact-curation paths.
- Kept `last_success_ts` because it describes the combined curator pass, whereas `last_run.at` and `last_staged_at` describe only the reflector mint stage.
- Dropped `mint_succeeded`, `last_successful_mint_ts`, the return-type change, and all duplicate per-item counters from this branch.

Validation:

- Focused residual/curator suites: `79 passed`; final signal-file refinement: `6 passed`.
- Full local pytest: `2938 passed, 18 failed, 17 skipped, 11 warnings`. The 18 failures are the stated Windows baseline: autoevolve x12, bridge_locking x3, selfevo_export x2, and `test_fail_open_tag_failure_never_breaks_a_green_cycle`.
- CI for pushed head `68ddd4cd`: Python 3.11, 3.12, and 3.13 green.

Host rollout/live verification remains operator-owned.

ozand:
Rebase and re-scope done exactly as asked. Verified by behaviour, not by diff. Merging.

**The core requirement, demonstrated:**

```
absent store   staged=0  store='absent'    (19 durable fields)
empty store    staged=0  store='present'   (19 durable fields)
```

Both return zero — the caller counts promotions, so that cannot change — and the durable record separates them. That is the acceptance criterion of #1216 in one line: *unavailable is distinguishable from empty*, and it now is at the point where a reader looks, not only at the point where a human reads a log.

**`_empty_reflector_run` / `_persist_reflector_run` are the part I did not specify and should have.** Before this, an early return on an absent, stale or unreadable store wrote **nothing** to the sidecar — the run left no record at all, which is precisely the failure mode #1216 exists to close, sitting inside the mint's own error paths. Now every exit writes a zero-count record carrying `store`. Good catch; it is a better answer than the one in my brief.

**You also took the free finding and extended it.** All four sidecar-only fields are now in the printed line — `rows_after_cursor`, `unparseable`, `evicted`, `at` — plus `store`. `evicted` is the field UM nominated as the tell for its own least-certain claim (the 400-entry pool bound, which I measured hitting on run one), and until now a journal reader could not see it.

**Keeping `last_success_ts` was the right call and your reason is the right reason:** it covers the whole curator pass, while the sidecar's `at` and `last_staged_at` cover only the reflector-mint stage. I asked you to judge it rather than keep it from sunk cost, and you judged it on scope.

Blast radius on my machine: **94 passed** across `test_issue_1216_*`, `test_issue_1171_*`, `test_knowledge_curator`, `test_lesson_v2`, `test_issue_1209_*`, `test_curator_staging_pickup`, `test_issue_1138_*` (both), `test_side_role_cliffs`, `test_issue_1214_*`. Your full-suite itemisation — autoevolve ×12, bridge_locking ×3, selfevo_export ×2, `test_fail_open_tag_failure_never_breaks_a_green_cycle` — matches my independently measured `origin/main` baseline exactly, at 2938 passed.

One thing on record, and it is mine: this PR spent a cycle building `mint_succeeded` and `last_successful_mint_ts` that #1221 was concurrently making redundant, and I sent you a defect report on code that was about to be deleted. The overlap was mine to catch before dispatching two agents into the same function. You rebased across a 435-line rewrite of it without complaint and the result is smaller and better scoped than what you started with.

PR #1221, opened 2026-09-02
feat(curator): reflector recommendations earn a card on recurrence, not on sight (#1171)

### LINKED ISSUE #1171: reflector lesson minting is dormant: 249 reflection-derived proposals in 8 days, 4 lessons total and all from one day

ozand:
Closes #1171. Design agreed on the issue (comment 5516742216 and the reply); this PR is that design.

## What changes

`promote_reflector_recommendations_to_v2` (`knowledge_curator.py`) — **a reflector recommendation earns a card on recurrence, folds on agreement, otherwise waits.**

- **Cursor replaces the window.** A timestamp watermark in `state/curator/reflector_pool.json`; each run reads every row after it across the newest two archives plus the live journal (`_reflector_rows_after`), at most `_REFLECTOR_MAX_ROWS_PER_RUN = 600` rows, and carries the remainder over. `_REFLECTOR_TAIL_ROWS` and `_reflection_tail_lines` are gone.
- **Fold first, uncapped.** An item folds into an existing card whose solution says the same thing (keyword Jaccard ≥ `_REFLECTOR_FOLD_THRESHOLD = 0.35`) or whose problem matches (`find_duplicate` rule, hash or ≥ 0.8). Folds advance `seen_count`/`last_seen`/`evidence` and never grow the base. A card whose solution is filler is matched through its `problem` (`_fold_words`) — that is where the 2026-08-29 mint stored the recommendation — and upgraded (#1106).
- **Pool, then graduate.** Otherwise the item joins the pool cluster it matches or starts one. A cluster graduates to a new card when it spans **≥ 2 distinct cycles AND ≥ 2 distinct days** (`_REFLECTOR_MIN_CYCLES`, `_REFLECTOR_MIN_DAYS`); graduation is evaluated over the whole pool after each run, most-recurrent first, so a cluster deferred by the cap or waiting on a second day graduates later without another matching item. Pool bounded to 400 clusters / 14 days (`_prune_pool`).
- **Cap is a valve.** `max_items = 5` new cards per run (was 2, and it counted folds).
- **Card fields.** `problem` = the item's `evidence` (the observation), `solution` = `detail`, `title` = `detail[:200]`, `seen_count` = distinct cycles, `evidence` = cycle ids (cap 8), `distinct_days` recorded per card.
- **Repair in place: yes.** When a re-read item is a card's own origin (its cycle already in the card's `evidence`, solution match ≥ 0.8), the card's narrative `problem` is replaced by the observation and `seen_count` is **not** incremented. On the live store this repairs `LESS-REF-4f03664182da-0d35` and `LESS-REF-16436c7d5b3a-4c67` (problem becomes "Tool execution after write_file 8v84c3XW6 produced TypeError…" / "Analysis during test writing noted: 'The script has a duplic…") and upgrades all four 08-29 filler cards, on the first backfill run, via the normal pickup.
- **Same rule at pickup.** `_merge_card_into` (used by `apply_staged_lesson_cards`) folds through `_fold_target`, so the bridge agrees with the curator's decision.
- **Signal.** `last_run` in the sidecar and one `curator-reflector:` stdout line carry `rows_read, rows_processed, items, folded, repaired, pooled_new, pooled_recurrence, same_day_only_waiting, graduated, deferred_by_cap, near_misses, rejected, staged, pool_size, cursor`; `last_staged_at` is kept. Near-misses = items whose best match fell in `[_REFLECTOR_NEAR_MISS_FLOOR = 0.25, 0.35)`. Feeds #1216; does not decide its cadence question.
- `lesson_v2.keyword_set` / `set_jaccard`: normalize a text once per run; `keyword_jaccard` is unchanged in behaviour.

## Pre-change failure (origin/main `e6c8e9ef`)

Three-assertion repro (`um-scratch/1171/test_repro_1171.py`, old API only), verbatim:

```
E       AssertionError: a single sighting minted a card
E       assert 1 == 0
E        +  where 1 = promote_reflector_recommendations_to_v2(...)
E       AssertionError: assert 'One narrativ...commendations' == 'LiteLLM retu...r three calls'
E         - LiteLLM returned 502 for three calls
E         + One narrative shared by both recommendations
E       AssertionError: assert ['Configure a...tay covered.'] == ['Configure a...tay covered.']
E         At index 1 diff: 'Configure a fallback model group ...' != 'Register newly added AGENTS.md sections ...'
E         Left contains 2 more items
3 failed in 1.73s
```

One sighting minted; the card's problem was the cycle narrative; two same-row recommendations produced four staged cards (each folded into the other's narrative). Same repro on this branch: `3 passed`.

## The live store through this code (read-only copy, 738 rows / 502 items, store from `origin/main` of the instance repo, 110 entries / 6 v2)

| run | ms (here) | rows | items | folded | repaired | pooled_new | same-day waiting | graduated | deferred | near-misses | staged |
|---|---|---|---|---|---|---|---|---|---|---|---|
| 1 | 412 | 600 | 432 | 5 | 4 | 400 | 9 | 5 | 1 | 85 | 10 |
| 2 | 240 | 138 | 70 | 2 | 2 | 63 | 11 | 4 | 0 | 11 | 6 |
| 3 | 120 | 0 | 0 | 0 | 0 | 0 | 9 | 0 | 0 | 0 | 0 |

Result: 6 v2 cards → 15 (9 graduated, 6 repaired in place, 0 rejected). Top cards: LiteLLM fallback (seen 4, 2 days), commit-early-in-turn-budget (3, 2 days), document out-of-scope pre-existing failures (3, 3 days), pytest-not-unittest, loop-breaker pivot, chained git exec. **9 clusters wait as same-day-only** — the 7-cycle LiteLLM echo of 2026-09-01 among them; with days not gated they would have been 21 cards (the design comment's number). 96 of 502 items were near-misses in `[0.25, 0.35)` — 19%; that is the shoulder you asked to see, and it says the threshold is not obviously too high or too low without a second week. Pool reached its 400 bound during run 2 (singletons from 08-27/28 evicted by size before by age).

**Cost per run.** Backfill run: 0.41 s here (was 2.5 s before caching the per-entry problem keys — `find_duplicate` re-normalized every entry per item, 86% of the run); the Atom N550 is ~10–15× slower → **~4–6 s once, split over two nightly runs by the 600-row cap**. Steady state (~120 rows, ~70 items/day against ~120 cards + 400 pool): 0.24 s here → **~3 s/day on the host**. Memory: sidecar ~250 KB at the bound. No LLM, no embeddings.

## Tests

`tests/test_issue_1171_reflector_mint_recurrence.py` — 20 tests: first sighting waits; two cycles/two days graduate with the observation as problem; same-day recurrence waits and is counted; same-row siblings stay separate lessons; fold is uncapped (`max_items=0`); narrative problem repaired from the origin item without re-counting; 08-29 filler card reached through its problem and repaired; cursor advances / rows not reprocessed; recurrence across archive + live; rows-per-run cap carries over; new-card cap defers and the next run finishes without new rows; near-misses counted; run counts distinguish candidates from idle; pool bounded and stale clusters evicted; error/unparseable/other-kind rows skipped; corrupt sidecar restarts; pickup folds into a card that arrived meanwhile; threshold constant and calibration note on record.

Adapted (fixtures that minted from one sighting now recur it across two cycles on two days; the durability/id assertions are unchanged): `test_issue_1209_durable_curator_writes.py`, `test_issue_1138_lesson_ids.py`, `test_issue_1138_lesson_ids_guard.py`, `test_lesson_v2.py` (padding rows now timestamped so the cursor can advance through them), `test_side_role_cliffs.py` (archive read now observed in the pool).

## Review pass (cavecrew-reviewer, applied)

- Rows sharing the boundary timestamp are never split across runs (the cursor skips `ts <= cursor`, so a tie at the 600-row cap would have lost the rows past it). Real journal timestamps are microsecond-unique; closed anyway, tested.
- A graduation that folds into a card minted earlier in the same run refreshes that card's fold caches (within-run staleness only — the caches are per-run locals).
- The sidecar write is fail-open (cards are already staged; a lost sidecar re-processes the rows, idempotently), and `run_curation` now wraps the mint so an exception there cannot take the fact-curation pass down — a pre-existing exposure, logged as a warning rather than swallowed silently.

## Full pytest (Windows, `-n auto`, this branch `83ea0305`)

**18 failed, 2920 passed, 17 skipped.** The 18: `test_autoevolve*` (12, symlink/tar), `test_bridge_locking` (3, flock), `test_selfevo_export_security` (2, `master` refspec), `test_bridge_cycle_tags::test_fail_open_tag_failure_never_breaks_a_green_cycle` (fails on origin/main, per your note). None touch the curator or lesson modules. Absolute counts as asked; reconcile against your baseline.

## Least-certain claim

**The pool bound (400) bites before the age bound (14 d) at the live rate**: ~67 singletons/day fill it in six days, so a recommendation first seen on day 1 and repeated on day 8 is evicted before it can recur. Every recurrence in the calibration week spanned ≤ 4 days, so it held there, but that is one week. `evicted` is in the run counts; if it is non-zero while `same_day_only_waiting`/`graduated` stay flat, raise `_REFLECTOR_POOL_MAX` — cost is linear in it (~0.5 ms per 100 clusters per item here).

Also on record: 0.35 / K=2 from one week of one loop (comment block above the constants); `error_pattern` is never a recommendation kind in the reflector's own prompt (it is a *finding* kind), so `_REFLECTOR_KINDS` effectively means `approach_hint` — left as is, not widened.

## Live verification (yours)

After deploy and the first 00:00 MSK run: `state/curator/reflector_pool.json` `last_run` with `items > 0`, `graduated + folded > 0`, `cursor` at the newest row; after the next bridge pickup, `git show origin/main:lessons/lessons.yaml` carries new `LESS-REF-*` cards with `distinct_days: 2`, `seen_count ≥ 2`, `problem` an observation, and the two 09-02 cards with their problems replaced. Do not hand-write a card.

### DIFF
--- a/nanobot/runtime/knowledge_curator.py
+++ b/nanobot/runtime/knowledge_curator.py
@@ -52,7 +52,9 @@ from nanobot.runtime.lesson_v2 import (
     fill_related_links,
     find_duplicate,
     inline_related_slugs,
+    keyword_set,
     related_hint,
+    set_jaccard,
     validate_lesson_for_mint,
 )
 from nanobot.runtime.model_registry import resolve_model
@@ -336,21 +338,6 @@ def _reflection_lines(live: Path):
                 yield from fh
         except OSError:
             continue
-
-
-def _reflection_tail_lines(live: Path, max_lines: int) -> list[str]:
-    """The last ``max_lines`` journal lines across the newest archive and the
-    live file — a freshly rotated live journal is nearly empty (#1178)."""
-    lines = _bounded_tail_lines(live, max_lines)
-    if len(lines) < max_lines:
-        for archive in reversed(_reflection_archives(live, newest=1)):
-            try:
-                with gzip.open(archive, "rt", encoding="utf-8") as fh:
-                    older = fh.read().splitlines()
-            except OSError:
-                continue
-            lines = older[-(max_lines - len(lines)):] + lines if older else lines
-    return lines
 
 
 def _bounded_tail_lines(path: Path, max_lines: int) -> list[str]:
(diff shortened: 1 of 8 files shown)


ozand:
Verified independently and merging. Every condition I set is implemented, and I checked the behaviour rather than the diff.

**Both directions, run against the deployed function with realistic reflector data:**

```
origin/main   single sighting -> staged = 1        <- mints on sight, the defect
PR #1221      single sighting -> staged = 0
              curator-reflector: rows_read=1 items=1 pooled_new=1 pool_size=1 staged=0

PR #1221      run 1 (first sighting)      -> staged = 0
              run 2 (recurrence, +2 days) -> staged = 1
                card id       : LESS-REF-...-fa10
                distinct_days : 2
                problem       : Executor re-implemented a helper that already existed in
                                scripts/verify_and_proof.py.
                evidence      : ['cycle-day1aaaaaaa', 'cycle-day2bbbbbbb']
```

The `problem` field is the thing I most wanted and it is right. Confirmed against the live store that reflector `evidence` is a **prose string** in all 46 current items, not an id list — so this yields a real problem statement instead of the cycle narrative the pre-#1171 mint wrote. (My first probe used a list there and produced nonsense; the fixture was wrong, not the code.)

**The conditions I set, each located in the diff:**

| condition | where |
|---|---|
| incident echo ≠ recurrence, made visible not gated | `distinct_days` (1077, 1444-1445), `same_day_only_waiting` counter |
| 0.35 as a named constant with the calibration recorded | `_REFLECTOR_FOLD_THRESHOLD` (995), comment at 986-995 naming one week / one loop |
| near-miss band counted | `_REFLECTOR_NEAR_MISS_FLOOR` (991), `near_misses` (1282) |
| problem = item evidence | verified above |
| repair of the two cards already on origin/main | `repaired` (1266) |
| cursor replaces window | sidecar `cursor`, advanced per run |
| fold at pickup, one implementation | `_merge_card_into` extended, `distinct_days` merged at 1444 |

Blast radius on my machine: **88 passed** across `test_issue_1171_*`, `test_knowledge_curator`, `test_lesson_v2`, `test_issue_1209_*`, `test_curator_staging_pickup`, `test_issue_1138_*` (both), `test_side_role_cliffs`, `test_issue_1214_*`. Your full-suite itemisation matches my independently measured `origin/main` baseline exactly — autoevolve ×12, bridge_locking ×3, selfevo_export ×2, `test_fail_open_tag_failure_never_breaks_a_green_cycle` — 18, with +20 passing.

Two things I want on the record.

**The 0.41 s number came from profiling, not estimation** — you found `find_duplicate` re-normalizing every entry per item at 86% of the run and removed it. That is the difference between a cost claim and a cost measurement, on a host where it matters.

**Your least-certain claim is the right one and is now instrumented:** the 400-entry pool fills in ~6 days at 67 singletons/day, so the bound bites before the 14-day age does. `evicted` in the run counts is the tell, and if it climbs the pool is too small rather than the recurrence rule being wrong. Recorded so nobody misreads eviction as noise.

**Also noted, not changed, and worth its own issue eventually:** `error_pattern` is never emitted as a recommendation kind by the reflector's own prompt, so `_REFLECTOR_KINDS` is effectively `{approach_hint}`. That matches what I measured in the live tail — 23 `approach_hint`, 1 `instruction_change`, zero `error_pattern`.

One process note that is mine, not yours: this PR and #1220 were dispatched into the same file in parallel, which is precisely what my own rule forbids. #1220 will rebase onto this and be re-scoped to the residual.

Merging.

Issue #1222, opened 2026-09-02
Orphaned state reads beyond research/: credits, outbox, control_plane, goals/registry, reports/evolution-*, durable.json, completed/ probe (found by #1219's writer registry)


ozand:
## Found while building #1219's state-path writer registry

#1219 adds `nanobot/runtime/state_paths.py` (`STATE_PATH_WRITERS`) and a test that requires every state subpath read by `nanobot/` to declare a writer that still exists. Mapping the 24 read segments to writers found more reads with no writer in this repository than the two `research/` files. Measured read-only on the host 2026-09-03 02:10 UTC (`ls --time-style=long-iso`):

| state path | readers | writer in `nanobot/` | live mtime | note |
|---|---|---|---|---|
| `credits/latest.json` | `state.py` | none | **2026-08-22 02:00** | same instant as `research/*` — same deleted writer (#923) |
| `outbox/{latest,report.index}.json` | `bridge.py:2043`, `state.py` | none | **2026-08-22 02:00** | same |
| `control_plane/current_summary.json` | `autoevolve.py:569` | none | **2026-08-22 02:00** | same; `stale_execution_repair.json` (08-28) has no in-repo writer either |
| `goals/registry.json`, `goals/active.json`, `goals/cycle_archive.json` | `bridge.py`, `backlog_snapshot.py`, `llm_proposer.py`, `state.py`, `archive.load` | none (`archive.record_stepping_stone` writes the archive only via callers that no longer exist?) | **2026-08-22 02:00** | `state.py:930` still says "rewritten every cycle by the current loop"; `derived_priorities.json` (goal_review) is live, `goal_text.json` is operator canon |
| `reports/evolution-*.json` | `autoevolve.py:572`, `state.py:1073` | none | last **2026-08-21 23:00** | `proof-*.json` in the same dir is live (49 files) and written outside `nanobot/` |
| `hypotheses/durable.json` | `hypothesis_backlog.py`, `demand.py` | none | 2026-09-02 03:00 (daily) | a writer exists **outside this repo**; identify it and register it, or the registry cannot vouch for it |
| `completed/completed.json` | `doctor.py:137` (integrity probe) | none | directory does not exist | probe of a path nothing has written; `demand/completed.json` is the live file |
| `self_evolution/` | `autoevolve.py`, `state.py:1296` | `autoevolve.write_guarded_evolution_state`, `write_noop_export_status` | directory empty since 2026-06-20 | writers exist but never run on the host |

[...]

PR #1223, opened 2026-09-02
fix(existence_index): retire the hypothesis corpus; readers stop reading state/research; state-path writer registry (#1219)

### LINKED ISSUE #1219: state/research/ has had no writer since its producer was deleted, and existence_index turns its frozen titles into suppression evidence
## Problem

`state/research/` has had no writer since 2026-08-22 and is read by five live code paths. One of them converts its contents into duplicate-suppression evidence.

```
$ ls -la /var/lib/eeepc-agent/self-evolving-agent/state/research/
drwxr-x---   2 eeepc-agent eeepc-agent  4096 Jun 22 02:06 .
-rw-r--r--   1 eeepc-agent eeepc-agent  1855 Aug 22 02:00 feed.json
-rw-r--r--   1 eeepc-agent eeepc-agent 41902 Aug 22 02:00 hypotheses.json
```

Both files are 12 days old. There is no code left that writes either one — `grep` for a writer across `nanobot/` returns nothing.

## Why the writer is gone

The producer was `cycle_planning._write_research_feed`. `cycle_planning.py` was **deleted** — `runtime_deny.py:212` records it: *"#924: the former rung 3, cycle_planning.py, was dropped"*. The deterministic planner itself was retired earlier in #739.

The readers were never revisited. This is the decommission failure mode: a component's *responsibilities* were not mapped when its *file* was removed, so every consumer kept consuming.

## The five readers, and which ones matter

| reader | what it does with it | live? |
|---|---|---|
| `existence_index.py:406-431` | indexes up to **50** frozen research candidate titles into the `hypothesis` corpus as `hyp-research-<cycle_id>-<idx>` | **yes — this is the harmful one** |
| `hypothesis_backlog.py:152-181` (`_research_candidates`) | offers the same titles to the LLM proposer as `serves: hypothesis <id>` targets, tagged `source: "research"` | yes |
| `demand.py:1647` | reads `research/hypotheses.json` | yes |
| `bridge.py:4747-4850` | auto-seeds 2 backlog priorities from `feed.json` when the backlog is empty | **no** — 0 firings in 14 days |
| `bridge.py:4777` | `hypotheses.json` fallback for the same seeding | no |

`feed.json` holds 2 entries and, per the journal, has seeded nothing in 14 days. `hypotheses.json` holds **50 snapshots** and its consumers run every cycle.

## The sharp consequence

`existence_index` indexes those titles as documents in the `hypothesis` corpus. A research *candidate title* is a statement of what someone thought should be built. Indexing it as an existence document makes it evidence the thing already exists, so a new proposal can be refused as a duplicate of an intention from a frozen file that was never implemented and now never will be.

**This is #1215's defect in a second corpus.** #1215/#1218 addresses `ledger_title` — a refused attempt's title treated as proof. This is the same error with a different source, and #1218 does not cover it: the corpus name is `"hypothesis"`, the retirement rule added there keys on `ledger_title`, and the entries are re-upserted from the frozen file on every reindex so they can never age out.

## Not a simple deletion

`hypothesis_backlog.py` reads this file **deliberately** — #751 revived the hypothesis→priority chain precisely because "with the planner off, nothing on the live path read either file". That was correct when written; the file was still being produced then. The producer's later deletion turned a deliberate revival into a permanent stale feed. Deleting the readers would re-break what #751 fixed; leaving them serves 12-day-old data forever. Neither is obviously right, which is why this is an issue and not a patch.

Note also `hypotheses/backlog.json` is **fine** — still written every cycle by `cycle_persist._build_hypothesis_backlog_snapshot`, mtime 0.0 days. Only the `research/` pair is orphaned. Do not conflate them.

## Acceptance criteria

- [ ] Decide, and record the decision with its reason: does `state/research/` get a writer again, or do its readers stop reading it? Do not split the difference silently.
- [ ] If the readers stop: `existence_index` must not index `hyp-research-*` documents, and the ones already in the index must be **deactivated**, not merely left inert — they are active suppression documents today. `hypothesis_backlog._research_candidates` and `demand.py:1647` must be updated in the same change, with #751's intent addressed explicitly rather than reverted by accident.
- [ ] If it gets a writer: name the component that owns the duty, and state why that duty survived `cycle_planning`'s deletion when its implementation did not.
- [ ] Either way, an orphaned state file must become detectable: a reader of a path with no writer should be visible without someone grepping for it. Coordinate with #1216 rather than inventing a second mechanism.
- [ ] Measure before/after: how many proposals were refused against a `hyp-research-*` document. If the answer is zero, say so — that changes the priority and is a legitimate outcome.
- [ ] Full pytest green; CI green (Linux matrix authoritative).

## Notes

Work in your own worktree, never `T:\Code\eeebot`. `existence_index.py` and `demand.py` are central runtime modules — full pytest suite, not the touched files. Current `origin/main` baseline is **18 failed, 2900 passed** on Windows; `test_bridge_cycle_tags.py::test_fail_open_tag_failure_never_breaks_a_green_cycle` is among the 18 and is not yours.

Belongs under the #1173 ADR (fail-open readers make 'unavailable' indistinguishable from 'empty') — this is the variant where the input is neither unavailable nor empty but *frozen*, which reads as healthy to every check.

ozand:
Closes #1219. Design agreed on the issue (comment 5517570157 and the reply); follow-up findings filed as #1222.

## Decision, recorded

**`state/research/` does not get a writer back; its readers stop.** The files were outputs of the deterministic planner (`cycle_planning._write_research_feed`), retired in #739 and deleted in #916/#923 (#924). The duty — hypotheses reach the proposer as `serves:` candidates — is owned by `backlog_snapshot.write_backlog_snapshot` (`hypotheses/backlog.json`, every cycle). **#751's intent is served from `backlog.json`, deliberately**: #751's own text names that file as the reader's source; `research/` was the tertiary source in every reader, held 2 distinct titles both already `stale` in the lifecycle, and contributed zero live candidates. Pinned by `TestResearchFeedIsNotASource` and the demand test.

Readers removed: `existence_index` (with the corpus, below), `hypothesis_backlog._research_candidates`, the `demand.py` research block, `bridge._auto_seed_backlog_from_research` (+ its only helper `_active_backlog_is_empty`). The seeder fired **twice in the repository's history — `2760a525` 2026-06-24 and `1520d8e7` 2026-06-22 — and never in the 70 days since**; its input froze with two entries. Nothing reads `state/research/` any more (`test_research_is_no_longer_read_anywhere`).

## The `hypothesis` corpus is retired, whole

A hypothesis is a statement of something not yet done; its title is never evidence an artifact exists — the backlog half exactly as much as the frozen research half. The corpus also had no consumer: a `hypothesis` hit is never `duplicate_suspect`, `related_scripts` filters it out, and the proposer gets hypotheses directly from `backlog.json`. Its only effect was crowding the BM25 top-k the gate reads. On the 400 newest live proposals against the live index copy: 36% of gate top-5 slots, a real script duplicate pushed out of the top-5 in **10** proposals, **536** proposer reuse hints lost; with only the 721 backlog docs kept: 11 and 458.

**Refusals against a `hyp-research-*` document: 0 of 186** dedup rows, ledger plus every archive — the issue's "if zero, say so" case. The corpus did not over-suppress; it defeated the gate and starved the proposer.

**Non-goal, on record** (in `_CORPORA`'s comment): "has this already been PROPOSED?" is a legitimate question and a different one from "does this EXIST?". If it is wanted it needs its own mechanism with its own semantics — not the existence index with a hypothesis corpus.

## Retirement is now the index's contract

Every corpus builder returns `(counts, evidence)` — the paths it re-derived from a living source this pass — and `reindex` deactivates every other active document of that kind for every kind in `_CORPORA = (script, ledger_title, hypothesis)`, including a kind with no builder (evidence = ∅). Retirement used to live inside each builder, so a builder without one had no retirement path and a source whose writer vanished left its documents active forever. Counts: `<kind>_deactivated` for all three; `hypotheses_indexed/unchanged` are gone.

`_deactivate_missing` is batched: `docs_fts` has no index on `path`, so the per-row delete scanned the table each time — retiring 6,935 rows took **11 s** on a desktop (minutes on the Atom, inside the bridge's gate). Now one statement per fully-retired kind, IN-list chunks of 400 otherwise: **1.3 s** here, ~15 s once on the host.

## The acceptance number — live index copy, new `reindex`, instance repo at HEAD

| | before | after first reindex | second reindex |
|---|---|---|---|
| `hyp-research-*` active | **6,214** | **0** | 0 |
| `hypothesis` active | 6,935 | 0 | 0 |
| `script` active | 346 | 347 (clone HEAD has one more script) | 347 |
| `docs_fts` rows | 7,761 | 347 | 347 |
| `hypotheses_deactivated` | | 6,935 | 0 |

The copy has no `subagents/` results, so its 480 ledger titles retired too (`ledger_titles_deactivated: 480`); that is an artifact of the copy, not a prediction for the host, where results are present. Rows stay in `documents` with `active = 0`. Host confirmation post-deploy is yours.

## Pre-change failure (origin/main `877f86e2`), verbatim

Old-API repro (`um-scratch/1219/test_repro_1219.py`):

```
E       AssertionError: a research candidate title became an active existence document
E       assert 1 == 0
E       AssertionError: 6 hypothesis documents stayed active with no source; reindex counts={'scripts_indexed': 0, ...
E       assert 6 == 0
2 failed in 1.51s
```

Same repro on this branch: `2 passed`.

## Detection: a reader with no writer fails a test

`nanobot/runtime/state_paths.py` — `STATE_PATH_WRITERS: {segment: (writer refs)}`; `tests/test_state_path_writers.py` — 12 tests:

- **Load-bearing half:** every declared writer resolves to a real module attribute (`importlib` + `getattr`) or an existing repo file; a non-empty string is not enough. `nanobot.runtime.cycle_planning:_write_research_feed` fails it today (tested) — #924's case, independent of whether the reader scan sees anything.
- Coverage half: the scan of `state_dir / "x"` / `state_root / 'x'` / `STATE_DIR / 'x'` (24 segments today) must be ⊆ the registry, else the failure names the segment and every reader (tested on a synthetic tree in #1219's shape). The scan is a floor, as you found; the registry may list more.
- `orphan:#<issue>` is a **declared, tracked** orphan, not an exemption: the test requires the issue to be listed in `ORPHAN_ISSUES` with its reason, rejects an issue nothing references, and hot directories (`ledger`, `reflector`, `subagents`, `curator`) are asserted to have real writers.

Relation to #1216: a different layer (static, test-time, "does a writer exist"), not a runtime signal. `scorecard._FEEDS` and its predicate are untouched.

### What the registry found — #1222

Mapping the 24 segments to writers found more writer-less reads than `research/`, all measured read-only on the host:

| read path | live mtime | |
|---|---|---|
| `credits/latest.json`, `outbox/{latest,report.index}.json`, `control_plane/current_summary.json`, `goals/{registry,active,cycle_archive}.json` | **2026-08-22 02:00** — the same instant as `research/*` | same deleted writer |
| `reports/evolution-*.json` | last 2026-08-21 23:00 | writer gone; `proof-*.json` alongside is live, written outside `nanobot/` |
| `hypotheses/durable.json` | 2026-09-02 03:00, daily | a writer exists **outside this repository** — not the strategist code here, which appends to `backlog.json` |
| `completed/completed.json` (doctor probe) | directory does not exist | nothing has ever written it |

Registered as `orphan:#1222` and not widened into this change; the PR touches none of those readers. `cycle_persist.py` is also gone (#923) — the docstrings that credited it now name `backlog_snapshot.write_backlog_snapshot`.

## Full pytest (Windows, `-n auto`, this branch `7e584bad`)

**18 failed, 2951 passed, 17 skipped.** The 18 by file: `test_autoevolve` 5, `test_autoevolve_commit` 3, `test_autoevolve_guards` 1, `test_autoevolve_remote_target` 1, `test_autoevolve_state` 2, `test_bridge_locking` 3, `test_selfevo_export_security` 2, `test_bridge_cycle_tags` 1 — the itemised baseline set. None touch the changed modules. Absolute counts as asked.

## Review pass (cavecrew-reviewer, applied)

One finding, and it was my own least-certain claim: a builder that *raises* mid-pass left its `evidence` at `∅`, which the contract reads as "retire everything of this kind" — an unreadable `scripts/` would have retired the live script corpus for a pass. Fixed: a raising builder's evidence is **unknown, not empty**; retirement for that kind is skipped that pass and reported as `retirement_skipped: ["script"]` (tested: the live corpus survives the failed pass, the other kinds still retire).

## Least-certain claim

**The registry's writer mapping is mine, from greps and the host's mtimes.** The test guards against *deleted* writers (the ref stops resolving) and *undeclared* readers (the scan), not against a *misattributed* one: a function that exists but does not actually write that directory passes. `subagent_bridge: bridge._main_impl_body` and `state: reflector._append_journal` (the nested-layout alias) are the two entries where I am least sure the named symbol is the real writer rather than the nearest one. If either is wrong the registry is still louder than the five-path `_FEEDS` it sits beside, but say so and I will correct the entry.

## Live verification (yours)

After deploy, the first bridge cycle's dedup gate triggers `reindex`: `hypotheses_deactivated: 6935` once in the journal, then 0; `sqlite3 index.sqlite "select count(*) from documents where path like 'hyp-research-%' and active=1"` → **0**; `select count(*) from docs_fts` ≈ 346 + active ledger titles (480 today). Nothing on the host is touched by hand.

### DIFF
--- a/nanobot/runtime/bridge.py
+++ b/nanobot/runtime/bridge.py
@@ -4753,134 +4753,13 @@ def _write_structured_error(
         return False
 
 
-
-def _active_backlog_is_empty(memory_text: str) -> bool:
-    """Return True if ## Active backlog section has no undone Priority blocks."""
-    import re as _re
-    # Extract Active backlog section (between BACKLOG_START and BACKLOG_END comments, or between headers)
-    section_match = _re.search(
-        r'## Active backlog.*?(?=\n## |\Z)',
-        memory_text,
-        _re.DOTALL,
-    )
-    if not section_match:
-        return True  # no active section → treat as empty
-    section = section_match.group(0)
-    # Check for any Priority block NOT marked Done
-    undone = _re.findall(r'###\s+Priority\s+\d+:(?!.*\[Done\])', section)
-    return len(undone) == 0
-
-
-def _auto_seed_backlog_from_research(
-    repo_root: Path,
-    memory_text: str,
-    memory_path: Path,
-) -> bool:
-    """If Active backlog is empty, add 2 new priorities from state/research/feed.json.
-
-    Returns True if MEMORY.md was updated.
-    """
-    import json as _json
-    import re as _re
-
-    if not _active_backlog_is_empty(memory_text):
-        return False
-
-    # Find research/feed.json — state root is ../state relative to repo
-    # Convention: repo at .../eeebot-self-evolving, state at .../state
-    state_root = repo_root.parent / 'state'
-    feed_path = state_root / 'research' / 'feed.json'
-    if not feed_path.exists():
-        return False
-
-    try:
-        feed = _json.loads(feed_path.read_text(encoding='utf-8'))
-    except Exception:
-        feed = {}
-
-    # Support both list-at-root and {entries: [...]} formats
-    if isinstance(feed, list):
-        entries = feed
-    else:
-        entries = feed.get('entries') if isinstance(feed.get('entries'), list) else []
-
-    # Fallback: if feed.json has no entries, try hypotheses.json
-    if not entries:
-        hyp_path = state_root / 'research' / 'hypotheses.json'
-        if hyp_path.exists():
-            try:
-                hyp_data = _json.loads(hyp_path.read_text(encoding='utf-8'))
-                if isinstance(hyp_data, list):
-                    # Each element is {cycle_id, candidates:[{title, acceptance}]}
-                    for hyp_entry in hyp_data[:5]:
-                        cands = hyp_entry.get('candidates') or []
-                        for c in cands:
-                            title = str(c.get('title') or '').strip()
-                            acceptance = str(c.get('acceptance') or '').strip()
-                            if title and title not in memory_text:
-                                entries.append({'title': title, 'acceptance': acceptance})
-                        if len(entries) >= 5:
-                            break
-            except Exception:
-                pass
-
-    if not entries:
-        return False
-
-    # Find next priority number
-    existing_nums = [int(m) for m in _re.findall(r'###\s+Priority\s+(\d+):', memory_text)]
-    next_num = max(existing_nums, default=8) + 1
-
-    new_blocks: list[str] = []
-    added = 0
-    for entry in entries:
-        if added >= 2:
-            break
-        title = str(entry.get('title') or entry.get('hypothesis') or '').strip()
-        if not title or title in memory_text:
-            continue
-        # Skip tasks that _task_already_done would immediately reject
-        if _task_already_done(title, repo_root):
-            print(f'bridge-memory: skipping already-done feed entry: {title[:60]}')
-            continue
-        acceptance = str(entry.get('acceptance') or entry.get('action') or '').strip()
-        instructions = acceptance or f'Research candidate from synthesize cycle: {title}'
-        block = (
-            f'### Priority {next_num}: {title}\n'
-            f'{instructions}\n'
-            f'Test: verify the change runs without errors.\n'
-            f'Commit: git add <file> && git commit -m "feat: {title[:50]}"\n'
-        )
-        new_blocks.append(block)
-        next_num += 1
-        added += 1
-
-    if not new_blocks:
-        return False
-
-    # Insert new blocks before BACKLOG_END comment (or before ## Completed)
-    insertion = '\n'.join(new_blocks)
-    if '<!-- BACKLOG_END -->' in memory_text:
-        updated = memory_text.replace(
-            '<!-- BACKLOG_END -->',
-            insertion + '\n<!-- BACKLOG_END -->',
-            1,
-        )
-    elif '## Completed' in memory_text:
-        updated = memory_text.replace(
-            '## Completed',
-            insertion + '\n---\n\n## Completed',
-            1,
-        )
-    else:
-        updated = memory_text.rstrip() + '\n\n' + insertion
-
-    if updated == memory_text:
-        return False
-
-    memory_path.write_text(updated, encoding='utf-8')
-    print(f'bridge-memory: auto-seeded {added} priorities from research/feed.json')
-    return True
+# #1219: ``_auto_seed_backlog_from_research`` (and its only caller-side
+# helper ``_active_backlog_is_empty``) (seed two MEMORY.md priorities
+# from ``state/research/feed.json`` when the Active backlog is empty) was
+# removed. Its input had no writer since ``cycle_planning`` was deleted (#924)
+# and froze on 2026-08-22 with two entries; it fired twice in the repository's
+# history (2760a525 2026-06-24, 1520d8e7 2026-06-22) and never in the 70 days
+# since. The demand lanes, not the research feed, drive the loop now.
 
 
 def _move_priority_to_completed(
@@ -5007,21 +4886,9 @@ def _try_mark_backlog_done(
         capture_output=True, text=True,
     )
     committed = result.returncode == 0
-
-    # Auto-seed: if Active backlog is now empty, add new priorities from research feed
-    if committed:
-        try:
-            fresh_text = memory_path.read_text(encoding='utf-8')
-            seeded = _auto_seed_backlog_from_research(repo_root, fresh_text, memory_path)
-            if seeded:
-                _sp.run(_git + ['add', 'memory/MEMORY.md'], capture_output=True)
-                _sp.run(
-                    _git + ['commit', '-m', 'chore: auto-seed backlog from research/feed.json (backlog empty)'],
-                    capture_output=True,
-                )
-        except Exception:
-            pass  # never block on auto-seed failure
-
+    # #1219: the auto-seed of MEMORY.md priorities from state/research/feed.json
+    # that used to follow here is gone with its writer-less input (see
+    # _move_priority_to_completed's neighbour comment above).
     return committed
 
 
--- a/nanobot/runtime/demand.py
+++ b/nanobot/runtime/demand.py
@@ -1643,25 +1643,10 @@ def _hypothesis_items(
             seen.add(title)
             evidence = str(entry.get("evidence") or entry.get("metric") or entry.get("data_to_collect") or entry.get("insight_criterion") or entry.get("acceptance") or "")
             items.append(_make_item("hypothesis", title, evidence))
-
-        research = _read_json(Path(state_dir) / "research" / "hypotheses.json", None)
-        if isinstance(research, list):
-            for snapshot in research[:50]:
-                if not isinstance(snapshot, dict):
-                    continue
-                for cand in snapshot.get("candidates") or []:
-                    if not isinstance(cand, dict):
-                        continue
-                    if not _is_active(cand):
-                        continue
-                    title = str(cand.get("title") or cand.get("hypothesis") or "").strip()
-                    if not title or title in seen:
-                        continue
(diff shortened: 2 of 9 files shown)


ozand:
Verified against the live index and merging. This is the largest single correction of the loop's suppression behaviour so far.

**Acceptance criterion, run with the deployed runtime patched with this PR, against a copy of the real index and the real instance repo on the host:**

```
BEFORE  total=8745 active=7767
           hypothesis      6939  active=6939
           ledger_title    1459  active= 481
           script           347  active= 347
           hyp-research    6214  active=6214

reindex: hypotheses_deactivated=6939  scripts_deactivated=0
         ledger_titles_deactivated=0  retirement_skipped=[]

AFTER   total=8745 active=828
           hypothesis      6939  active=   0
           ledger_title    1459  active= 481
           script           347  active= 347
           hyp-research    6214  active=   0
```

`hyp-research` 6,214 → 0, whole corpus retired, and — the part that matters as much — **`script` and `ledger_title` are untouched**. 347 and 481 active before and after. A retirement change that reached into neighbouring corpora would have been the obvious way for this to go wrong.

A note on method, because my first run of this produced alarming numbers: I initially pointed `reindex` at the *product* worktree instead of the *instance* repo and saw `scripts_deactivated=263`, `ledger_titles_deactivated=481`, active collapsing to 219. That was my setup, not your code — with the wrong repo every indexed script legitimately looks deleted. Recording it so nobody repeats the scare.

**The registry's load-bearing half, mutation-tested.** I pointed one real writer entry at the module #924 deleted:

```
-  "nanobot.runtime.action_index:build_action_index"
+  "nanobot.runtime.cycle_planning:_write_research_feed"

FAILED tests/test_state_path_writers.py::test_every_declared_writer_still_exists
AssertionError: assert ['state/actio...tFoundError)'] == []
```

That is the counterfactual made concrete: had `STATE_PATH_WRITERS` existed on 2026-08-22, commit `72b0b6e` would have failed its own test the moment it deleted `app/main.py`, and named the paths it was orphaning. Twelve days and 6,214 stale suppression documents.

Blast radius on my machine: **536 passed** across `test_existence_index`, `test_demand`, `test_hypothesis_backlog`, `test_state_path_writers`, `test_bridge_dedup_target_path`, `test_llm_proposer`, `test_llm_proposer_dedup_budget`, `test_bridge_recent_failure_suppress`. Your full-suite itemisation matches my measured `origin/main` baseline at 2951 passed.

**On `orphan:#1222` as a registry value.** It is an escape hatch and I want it named as one: a path marked orphan passes the test while having no writer. What makes it acceptable rather than the exemption list I argued against is that it is *explicit, per-path, and points at an issue* — the omission is loud and has an owner, instead of silent. Keep it that way; if `orphan:` entries start accumulating without issues attached, the mechanism has degenerated back into what it replaced.

**Your least-certain claim is the right one and I am recording it rather than waving it through:** the writer mapping is yours, derived from greps and mtimes, so the test catches deleted writers and undeclared readers but **not misattributed ones**. A wrong `module:function` that happens to exist passes silently. Your two weakest entries — `subagent_bridge → bridge._main_impl_body` and `state → reflector._append_journal` — are recorded here so a future reader knows which two to distrust first.

Also good: the reviewer pass found the raising-builder hazard (builder raises → evidence set empty → mass-retire) which was your *own* previously-named uncertainty, and the fix reports `retirement_skipped` rather than silently doing nothing. `retirement_skipped: []` in the run above confirms it did not fire on real data.

Merging. #1222 carries the remaining sixteen orphans, now with their common cause attached.

PR #1232, opened 2026-09-03
fix(strategist): read reusable lesson insights (#1231)

### LINKED ISSUE #1231: Strategist insights input reads two keys the corpus does not have and ignores the one it does: reusable_insight is 14 of 14, and the empty result reports complete
## Context (self-contained)

The strategist (#999) is paused on the host — `eeebot-strategist.timer` is `disabled`, an operator action taken on 2026-09-02 after that issue's audit found most of its inputs dead. The pause was recommended as "until the input fix lands", not as a retirement.

Four of the five inputs have since been repaired and are live. Measured 2026-09-03 by calling the **deployed** `strategist.collect_inputs()` read-only against live state (no run, no LLM call, nothing written):

```
goals           complete   2736 chars, source=release_root
scorecard       complete   41 history rows, 8 samples
funnel          complete   200 ids, ledger 31 files, covered to 2026-09-03T03:22Z
evolution_tree  complete   100 nodes, 300 fitness values
insights        complete   cards 0, legacy 0          <- empty, and says complete
```

## Defect 1 — the reader was pointed away from the only key the corpus has

The #999 audit stated that "the insight reader looks for a `reusable_insight` key that occurs zero times in the corpus", and `insights_input` was repointed onto v2 `problem`/`solution` cards plus legacy `generalized_insight`. `strategist_inputs.insights_input`'s docstring still records the claim: *"The historical `reusable_insight` key never existed."*

That is backwards. `lessons/lessons.yaml` on `origin/main`, 14 entries, counted by key:

```
with reusable_insight    14
with problem/solution     0
with generalized_insight  0
```

Two shapes, both carrying `reusable_insight`:

```
11 x (approach, category, date, description, id, prevention, reusable_insight, title)
 3 x (approach, category, description, files_changed, first_seen, id, impact,
      last_seen, occurrences, reusable_insight, sample_cycle_id, title)
```

`reusable_insight` is the only insight-bearing key present. The reader now looks for two keys that appear zero times and ignores the one that appears in every entry, so it returns `cards: 0, legacy: 0` and will keep doing so no matter how the corpus grows.

The v2 cards it wants are not merely late — they are not durable anywhere. The instance workspace has no `lessons/lessons.yaml` at all right now (checked as `eeepc-agent`), which is consistent with the curator's output being reset at cycle start.

## Defect 2 — an empty input reports `complete`, so the refusal guard cannot see it

`#1182` added the rule "≥2 of 5 inputs empty → refuse the LLM call". On the live state:

```
empty_inputs(inputs_status) -> []
should_refuse(inputs_status) -> False
```

The one genuinely empty input is classified `complete`, so the guard built to stop the strategist advising from an empty view is blind to the only empty view there is. This is the fail-open class of ADR #1173: empty is indistinguishable from populated because the status is derived from "the read did not raise", not from whether anything was read.

`errors: 5` in that status block is not a failure count — it is five entries from `lessons/errors.yaml`, i.e. content. Do not read it as an error signal.

## Task

1. `insights_input` accepts `reusable_insight` as an insight source alongside the v2 `problem`/`solution` shape. Keep the v2 path — it is the intended future shape and cards will appear there once the curator's output becomes durable — but stop ignoring the corpus that exists today. Fix the docstring's false claim in the same change.
2. `insights` reports `empty` when it yields no insight content, so `should_refuse` can see it. Content means cards or legacy insights; `errors.yaml` summaries and the index files are context and must not by themselves make the input look populated.

## Non-goals

- Do not change the curator, `lessons.yaml`'s schema, or migrate entries to the v2 shape. That is a separate question and this issue must not depend on it.
- Do not enable the timer as part of this. That is #999's decision and has its own conditions.
- Do not change the `_MAX_EMPTY_INPUTS` threshold or the refusal rule itself — only make the status honest enough for the existing rule to work.

## Acceptance Criteria

- [ ] Against the real `lessons/lessons.yaml` from `origin/main` (14 entries, all `reusable_insight`), `insights_input` returns a non-zero insight count — asserted against a copy of the real file, not a hand-built fixture in the shape the reader already expects.
- [ ] With a corpus that yields no cards, no legacy insights and no `reusable_insight` entries, `insights_input` reports `status: empty`, and `should_refuse` flips to `True` when one other input is also empty (test both halves — the status alone proves nothing if the rule still cannot act on it).
- [ ] A corpus carrying only `errors.yaml` summaries and index files reports `empty`, not `complete` (test).
- [ ] The v2 `problem`/`solution` path still works and is still preferred where both shapes are present (test).
- [ ] Live verification after deploy: `collect_inputs()` run read-only against live state shows `insights` non-empty with a real count, quoted on this issue. Do not run the strategist itself to obtain this.

ozand:
## Summary

Fix the final strategist input defect from #1231:

- `insights_input` now reads the existing `reusable_insight` field as the legacy insight source;
- v2 `problem`/`solution` cards remain preferred when present;
- `insights` reports `empty` when there are no cards or legacy insights, regardless of `errors.yaml` summaries or index context;
- the existing `should_refuse` rule can therefore act on an actually empty insights input.

## Evidence from the real corpus

Added `tests/fixtures/lessons-origin-main.yaml`, copied from `origin/main:lessons/lessons.yaml`. It contains 14 entries:

- 14/14 have `reusable_insight`;
- 0 have `problem`, `solution`, or `generalized_insight`;
- the reader returns 10 legacy insights, matching the existing `_INSIGHTS_LEGACY` cap.

## Validation

- Focused strategist input suite: `18 passed`.
- Full local pytest: `2962 passed, 19 failed, 17 skipped, 11 warnings`.
- The 19 failures are the established 18 Windows baseline plus the pre-existing `tests/test_strategist.py::test_collect_inputs` fixture expectation for the old `lessons` context shape; no changed strategist-input acceptance test failed.
- Host strategist/timer was not run or enabled. Live verification remains operator-owned and must call `collect_inputs()` read-only after deployment.

## Deliberately unchanged

- curator and lessons schema;
- timer state;
- `_MAX_EMPTY_INPUTS` and refusal rule;
- v2 card semantics and caps;
- errors/indexes as prompt context, but not evidence that insights are populated;
- host state.

Closes #1231

### DIFF
--- a/nanobot/runtime/strategist_inputs.py
+++ b/nanobot/runtime/strategist_inputs.py
@@ -35,7 +35,7 @@ _HISTORY_DAYS = 7
 _HISTORY_SAMPLES = 8
 _HISTORY_MAX_SERIES = 60
 _HISTORY_MAX_DEPTH = 3
-# Insight sources: v2 lesson cards, legacy ``reusable_insight`` rows, errors.
+# Insight sources: v2 lesson cards, legacy rows under either insight key, errors.
 _INSIGHTS_V2 = 10
 _INSIGHTS_LEGACY = 10
 _INSIGHTS_ERRORS = 5
@@ -239,9 +239,17 @@ def _clip(value: Any) -> str:
 
 def insights_input(repo_root: Path) -> tuple[dict[str, Any], dict[str, Any]]:
     """v2 lesson cards (``problem`` -> ``solution``, filler skipped via
-    ``lesson_v2.solution_is_meaningful``), distinct legacy
-    ``reusable_insight`` lines, ``errors.yaml`` summaries, plus the two
-    index files. Prefer v2 cards when both shapes are present."""
+    ``lesson_v2.solution_is_meaningful``), distinct legacy insight lines,
+    ``errors.yaml`` summaries, plus the two index files. Prefer v2 cards when
+    both shapes are present.
+
+    Legacy insights carry either key. ``reusable_insight`` is what the whole
+    corpus on ``origin/main`` uses (14 of 14 rows, #1231); ``generalized_insight``
+    is what ``bridge`` and ``knowledge_curator`` write today, so it is the shape
+    new rows arrive in. Reading only one of them leaves the other invisible,
+    which is the defect #1231 was filed for — in one direction before this
+    change and in the other after it. ``knowledge_curator`` already treats the
+    two as equivalent when it collects source text."""
     from nanobot.runtime.lesson_v2 import bounded_load_yaml, solution_is_meaningful
 
     root = Path(repo_root)
@@ -255,8 +263,8 @@ def insights_input(repo_root: Path) -> tuple[dict[str, Any], dict[str, Any]]:
             elif len(cards) < _INSIGHTS_V2:
                 cards.append({"id": str(entry.get("id") or ""), "problem": _clip(entry.get("problem")),
                               "solution": _clip(entry.get("solution")), "first_seen": str(entry.get("first_seen") or "")})
-        elif entry.get("reusable_insight"):
-            text = _clip(entry.get("reusable_insight"))
+        elif entry.get("reusable_insight") or entry.get("generalized_insight"):
+            text = _clip(entry.get("reusable_insight") or entry.get("generalized_insight"))
             if text not in legacy and len(legacy) < _INSIGHTS_LEGACY:
                 legacy.append(text)
     errors = [{"title": _clip(e.get("title")), "root_cause": _clip(e.get("root_cause")), "prevention": _clip(e.get("prevention"))}
--- a/nanobot/runtime/strategist_inputs.py
+++ b/nanobot/runtime/strategist_inputs.py
@@ -35,7 +35,7 @@ _HISTORY_DAYS = 7
 _HISTORY_SAMPLES = 8
 _HISTORY_MAX_SERIES = 60
 _HISTORY_MAX_DEPTH = 3
-# Insight sources: v2 lesson cards, legacy ``generalized_insight`` rows, errors.
+# Insight sources: v2 lesson cards, legacy ``reusable_insight`` rows, errors.
 _INSIGHTS_V2 = 10
 _INSIGHTS_LEGACY = 10
 _INSIGHTS_ERRORS = 5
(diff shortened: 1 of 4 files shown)


ozand:
-

ozand:
Reviewed and extended. The core of this is right and the fixture discipline is exactly what was asked: `tests/fixtures/lessons-origin-main.yaml` is byte-identical to `lessons/lessons.yaml` on `origin/main` (verified by parsing both — 14 entries, all 14 carrying `reusable_insight`), so the test asserts against the corpus that exists rather than one written to match the reader.

**One defect, fixed in `a23c761c` on this branch: the key was swapped rather than added.**

The corpus justifies reading `reusable_insight` — 14 of 14 rows, none with the other key. But `generalized_insight` is not historical:

```
nanobot/runtime/bridge.py:4457            'generalized_insight',
nanobot/runtime/bridge.py:4618            'generalized_insight': insight,
nanobot/runtime/bridge.py:4736            'generalized_insight': f'Avoid {reason ...}',
nanobot/runtime/knowledge_curator.py:234  "generalized_insight": f"{kind}: {detail}"...
nanobot/runtime/knowledge_curator.py:251  "generalized_insight": summary_str,
nanobot/runtime/knowledge_curator.py:526  ..."reusable_insight", "generalized_insight"):
```

It is the shape *new* rows arrive in, and line 526 shows `knowledge_curator` already treats the two as equivalent when collecting source text. Reading only one of them is the defect this issue was filed for; swapping reproduces it in the opposite direction, and it would surface precisely when the curator's output finally becomes durable — the moment this input is supposed to start working properly.

Both keys are now read, v2 cards still take precedence, and the added test asserts a corpus with one row of each yields `legacy == 2`. Mutation-checked: reverting to either single key fails it and nothing else.

The status half is right as delivered — `"complete" if cards or legacy else "empty"`, with `errors.yaml` summaries no longer able to make an empty input look populated. That was the half that mattered, since without it `should_refuse` stays blind.

**Full suite on this branch with the addition: `18 failed, 2964 passed, 17 skipped`** — the 18 are the Windows baseline, checked by file (`test_autoevolve.py` ×5, `test_autoevolve_commit.py` ×3, `test_autoevolve_guards.py` ×1, `test_autoevolve_remote_target.py` ×1, `test_autoevolve_state.py` ×2, `test_bridge_locking.py` ×3, `test_selfevo_export_security.py` ×2, `test_bridge_cycle_tags.py` ×1). The `19 failed` in the report above was a run taken before the `test_strategist.py::test_collect_inputs` fixture was updated; it is 18 now.

Live verification after deploy is mine to run: `collect_inputs()` read-only against live state, expecting `insights` non-empty with a real count. The strategist itself stays paused until that is quoted on #999.

PR #1237, opened 2026-09-03
fix(deploy): restart dashboard after release activation

ozand:
Closes #1236.

After flipping `current`, restart the long-running `eeebot-dashboard.service` when enabled and verify active state plus MainPID/start/cwd/cmdline against the activated release. Preserve disabled/masked/not-found semantics. Centralize rollback so it restores `current` and restarts dashboard before the bridge.

Scope is deployment pipeline only: no dashboard application, scheduler, state store, auth, provider, bridge cadence, or unrelated unit changes.

Validation: bash syntax check; focused deploy integrity/release tests (24 passed); Ruff; diff check. Production rollout intentionally pending independent review.

Issue #1236, opened 2026-09-03
ops(deploy): restart dashboard after current release activation


ozand:
## Goal

Ensure a release rollout does not flip the `current` symlink while the long-running `eeebot-dashboard.service` continues serving code loaded from the previous release.

## Incident context

The #1206 rollout updated `current` and `SOURCE_COMMIT`, but `deploy_release.sh` restarted only the bridge. The dashboard process remained on the old code until manually restarted, so endpoint verification observed pre-fix behavior despite the new release pointer.

## Minimal scope

- after the `current` symlink and systemd unit sync, restart `eeebot-dashboard.service` when it is present/enabled;
- preserve disabled, masked, and not-found semantics; fail clearly when an expected enabled dashboard cannot be activated;
- assert the dashboard is active and its process started after the release flip;
- verify the running process/cwd and bounded endpoint proof correspond to the deployed release;
- include rollback handling that restores the previous symlink and restarts the dashboard;
- add focused static/shell test coverage for ordering and activation.

## Non-goals

- no dashboard application changes;
- no new scheduler, state store, broker, auth, provider routing, or fallback;
- no changes to bridge cadence or unrelated units;
- no automatic restart loop.

## Acceptance Criteria

- [ ] Normal deploy restarts the long-running dashboard after flipping `current` and before endpoint verification.
- [ ] Deploy fails clearly if an expected enabled dashboard cannot become active.
- [ ] Verification records current release, `SOURCE_COMMIT`, dashboard MainPID/start time, and process cwd/release identity.
- [ ] Rollback restores the previous release and restarts the dashboard.
- [ ] Focused tests pin ordering and disabled/masked/not-found behavior.
- [ ] CI is green and staged production rollout proves the dashboard serves the deployed release.

Metadata: owner:agent-runtime, wsjf:10.0, workstream:deploy, story_link:docs/specs/host-runtime/spec.md
Related: #1206

ozand:
Work started. This is a narrow deployment activation fix discovered during the #1206 rollback: keep the dashboard code unchanged, correct the restart/verification ordering, add focused tests, and prove it on a staged rollout before closure.

ozand:
Implementation update: PR #1237 is updated at commit `87bec04` with fail-closed dashboard activation, full-SHA/source verification, pre/post PID/start checks, semantic listener and endpoint JSON gate, and immediate rollback path. Focused deploy tests pass (24), CI is running; independent review remains pending. No host rollout performed.

[...]

PR #1239, opened 2026-09-03
fix(deploy): verify dashboard ownership and endpoint semantics

ozand:
Clean replacement for the stale #1237 branch.

Base: current `origin/main` at `97cb973a`. Deploy-only scope:
- restart enabled dashboard after current activation;
- verify full SHA/SOURCE_COMMIT, pre/post PID/start, cwd and exact command line;
- verify dashboard PID owns :8080 and `/api/health` + `/api/metrics` contain bounded source metadata and no retired payload/path markers;
- rollback immediately on post-flip activation, bridge, or gate failure, restarting dashboard before bridge.

Files are limited to `host/eeepc/scripts/deploy_release.sh`, `tests/test_deploy_integrity.py`, and `tests/test_deploy_release.py`. No dashboard application, provider, scheduler, state, or unrelated unit changes.

Focused deploy tests: 24 passed; bash -n, Ruff, and diff check passed. No host rollout included; independent review required.

### DIFF
--- a/host/eeepc/scripts/deploy_release.sh
+++ b/host/eeepc/scripts/deploy_release.sh
@@ -353,6 +353,8 @@ sync_timer eeebot-strategist.timer optional
 # the static oneshot bridge, it keeps the old Python process alive across a
 # symlink flip unless explicitly restarted (#1236).
 DASHBOARD_UNIT=eeebot-dashboard.service
+DASHBOARD_PREV_PID="$(systemctl show "$DASHBOARD_UNIT" -p MainPID --value 2>/dev/null || true)"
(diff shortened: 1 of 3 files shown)


ozand:
Superseded by #1242, which differs by one line and that line is the point.

This branch disarms the rollback trap **before** the bridge restart:

```
trap - ERR
# ... The activation rollback trap remains installed so a bridge restart failure
# restores the previous release.
sudo systemctl restart eeepc-self-evolving-subagent-bridge.service
```

The comment states the opposite of what the code does — a bridge restart failure would leave the new release active with no rollback, which is the window this issue exists to close. #1242 moves `trap - ERR` after the restart so the claim becomes true.

Third branch for #1236 (`fix/1236-dashboard-activation` → `-reconciled` → `-final`), all three the same size. Closing this one so there is one candidate rather than a merge order chosen by accident.

Issue #1238, opened 2026-09-03
doc_only_budget is 27% of the ledger: ~2.9 identical no-op rows per cycle crowd out real history for every byte-capped reader


ozand:
## Measured on the host, 2026-09-03 ~16:30 MSK

The `doc_only_budget` row from #1226 (issue #1108) is now the largest phase in the live ledger:

```
total rows            520
  doc_only_budget     140   <- 27% of the file
  demand_vector_split  87
  started              51
  dedup                51
  proposed             50
  outcome              49   <- actual cycles
  proposer_skip        35
  gate                 28
file size          120,593 bytes
```

**140 rows against 49 cycles is ~2.9 rows per cycle**, because `collect_demand` runs more than once per cycle and the event is appended unconditionally on every pass. Consecutive rows from one cycle are identical:

```
2026-09-03T13:12:44 doc_only_budget
2026-09-03T13:12:57 doc_only_budget
```

Almost all of them carry `doc_only_deferred: 0` with an unchanged budget state — the same fact, restated.

## Why this matters here specifically

The ledger is the input every bounded reader shares, and several of them are byte-capped: `strategist_inputs.ledger_rows` reads under an 8 MiB cap (currently `bytes_read: 2,144,076` across 31 files), `state_access.ledger_window` is capped per call, and #1178 exists because a 256 KB cliff silently truncated the scorecard history. Every byte of restated no-op is a byte of real cycle history a capped reader does not reach.

**This is not yet a failure** — no reader is at its cap today, and nothing is being dropped. It is a cost that grows with cycle rate, and the class this repository has shipped three times (#1183, #1178, #1166): a guard that becomes a silent off-switch for the thing it guards. Worth fixing while it is cheap to fix.

## The constraint that makes this non-trivial

The row exists because #1108 was unanswerable for months: a guard that never fired looked identical to a guard that worked. So "only write when `doc_budget_exceeded`" recreates the original defect — absence would once again be ambiguous.

What carries information is a *change* of state, plus enough of a heartbeat that "the guard has been reached and defers nothing" stays visible.

## Task

Write the row when it says something, and not otherwise. A shape that satisfies both sides — argue for a better one if you have it:

- write when any of `doc_budget_exceeded`, `ledger_blind`, or `doc_only_deferred > 0` differs from the last row written;
- write when `doc_only_deferred > 0`, always — a deferral is never a no-op;
[...]

PR #1226, opened 2026-09-03
fix(demand): make doc-only budget deferrals observable (#1108)

ozand:
## Summary

Make the existing #1108 doc-only budget guard observable without changing its behavior.

Each `collect_demand` pass now appends one `phase: doc_only_budget` event to the existing cycle ledger with:

- `doc_only_deferred`: number of doc-only demand items skipped this cycle;
- `doc_only_integrations_24h`: observed rolling count at selection time;
- `doc_only_budget_24h`: configured budget;
- `ledger_blind`: whether unavailable ledger evidence, rather than a real over-budget count, triggered suppression;
- `doc_budget_exceeded`: effective guard state;
- `items_considered`: count before deferral.

The event uses the existing best-effort `cycle_ledger.append_event` surface; no new state file, scan, read, size check, classifier, lifecycle transition, or counter mutation was added.

## Red-state evidence

A temporary regression test was run against `origin/main` before the fix. The guard skipped the doc-only item, but the expected ledger event was absent:

```text
2 failed, 173 passed in 31.16s
FAILED ...test_doc_only_deferrals_are_recorded_with_budget_state
FAILED ...test_ledger_blind_budget_deferral_is_recorded_as_distinct_cause
```

The same tests pass on this branch.

## Deliberately unchanged

- Deferral remains a selection-time `continue`.
- No lifecycle, dedup, exhaustion, or #1038 lane behavior changes.
- Default budget `N=5` and reporting key `value.doc_only_integrations_24h` unchanged.
- Shared `classify_change_tier` / `predict_item_change_tier` classifier unchanged.
- No forced live over-budget condition; no host writes, rollout, or live-store mutation.

## Validation

- Focused new regression tests: `2 passed`.
- Full pytest: pending; Windows baseline is the exact LIST supplied in #1108.
- Ruff: passed on changed runtime/test files.
- `git diff --check`: passed.

## Live verification

Deployment and the natural 24-hour live observation remain operator-owned. The durable ledger event is intended to answer both a real over-budget deferral and a `ledger_blind` deferral without conflating them.

Closes #1108

Issue #1108, opened 2026-08-30
doc-only guard: extend over-budget suppression to all demand lanes (24h budget exceeded via non-reflection lanes)


ozand:
## Context (self-contained)

Issue #1090 added a doc-only monoculture guard after the loop produced 36/53 doc-only commits in a day: integration-time classification (doc-only vs code-bearing), a rolling 24h budget (`N=5`, env-tunable), suppression of doc-only **reflection** demand items once the budget is hit, and the reporting key `value.doc_only_integrations_24h`.

Live observation 2026-08-30: `doc_only_integrations_24h = 8` — the budget is exceeded while the guard is active, because suppression only covers the reflection lane. Doc-only work still flows in through other demand lanes; the last 36h of instance-repo history shows integrated doc-only cycles like "docs: document runner heartbeat telemetry touch patterns", "docs: document pre-proposer existence gating rules", "docs: document harness execution coupling patterns", "docs: add inline assertion testing standards" — none of them reflection-sourced.

The 24h budget is meant as a cap on doc-only *integrations*, but today it is enforced only against one source lane, so the measured value can sit above budget indefinitely.

## Task

1. Extend the over-budget suppression to all demand lanes that can carry doc-only work: when `doc_only_integrations_24h >= N`, any demand item whose predicted output is doc-only (same classifier heuristics as #1090 uses at integration time, applied to the item's target path/kind) is skipped for selection this cycle — with the same bounded prompt notice #1090 already renders, and the same automatic release when the rolling window drops below budget.
2. Doc-only items are deferred, not deleted: no lifecycle/status change, no effect on dedup or exhaustion counters (#1038 lanes stay intact).
3. Keep the classification heuristics in ONE place shared by integration-time counting and selection-time prediction (no drift between the two).

## Non-goals

No blocking of doc work below budget (docs are legitimate work); no hard integration-time reject (a cycle that already ran is still integrated and counted); no changes to the N=5 default or the reporting key.

## Acceptance Criteria

- [ ] With the 24h budget reached, a doc-only priority/goal-gap/decay item is skipped for selection with the bounded notice; a code-bearing item in the same lane is selected normally (tests).
- [ ] Below budget, doc-only items select exactly as before — byte-identical selection behavior (test).
- [ ] Deferred items keep their lifecycle state and counters (test).
- [ ] Classifier shared: one function used by both counting and prediction (test or grep-proof in PR body).
[...]

PR #1240, opened 2026-09-03
fix(demand): write the doc_only_budget row on change, deferral and once per cycle (#1238)

ozand:
Tracks #1238 (stays open at `status:roll-out` until the live rows-per-cycle measurement is quoted there).

## What

`collect_demand` appended the `doc_only_budget` ledger row on every pass, and it runs two-to-three times per bridge invocation (gate probe in `should_propose` + context build in `propose`). Measured on the host: 140 rows against 49 `outcome` rows, 27% of the live ledger, nearly all `doc_only_deferred: 0` with unchanged budget state.

Now the row is written when it says something:

| condition | writes? |
|---|---|
| `doc_only_deferred > 0` | always — a deferral is never a no-op |
| first pass in this process | yes — the per-cycle heartbeat that keeps "reached, deferring nothing" visible (#1108) |
| `(doc_budget_exceeded, ledger_blind, doc_only_deferred > 0)` differs from the last row written in this process | yes — all four transition directions |
| otherwise | no; the pass is counted and folded into the next written row |

New additive field `passes` (int ≥ 1): how many passes the row stands for — itself plus the identical passes folded since the previous row in this invocation.

## Why this shape

The memory that remembers the last row is a module-level dict keyed by `str(state_dir)` — the same mechanism `llm_proposer._record_idle` already uses for its once-per-cycle idle heartbeat (#760). The bridge unit is `Type=oneshot`, one process per invocation (≤55 min by `TimeoutStartSec`), so "this process" is "this cycle". Zero I/O: no state file, no scan, nothing added to the write path — process memory is the only thing cheaper than the row it suppresses.

The process boundary *is* the heartbeat: the first pass of every invocation writes because the memo is empty. That is what keeps silence unambiguous — a cycle that reached the guard and deferred nothing still leaves a row with `doc_only_deferred: 0`, distinct from no row (never reached) and from `ledger_blind: true` (no data).

**Rejected:** writing only on the `emit_split=True` pass (#815's once-per-cycle trick) — the gate-probe pass is the one that runs in idle cycles; if it defers nothing the context build never happens, the row is lost, and #1108 reopens. Seeding the memo from the already-loaded in-memory `ledger_rows` would also work at zero I/O but removes the per-cycle heartbeat in exchange for nothing the issue asked for.

**Stated limitation:** passes folded *after* the last written row of a process are never reported — the process exits before another row is due. Today that is ≤2 identical passes per invocation; it is the volume being removed, not information about the guard.

## Effect on row count

Steady state: ~2.9 rows per invocation → **1**. Deferring invocations: one row per deferring pass (unchanged by design). Transitions: one extra row each. Live check owed after deploy: `doc_only_budget` rows per `outcome` row over a natural window, quoted on #1238 against today's 140:49.

## Compatibility — `ozand/eeebot-ops-dashboard` #201 (merged, `4457cd8`)

`scripts/techtree_viewer.py::_build_doc_only_budget_item` takes the latest `phase == 'doc_only_budget'` row from `ledger_tail` by reversed scan and reads `ledger_blind`, `doc_only_deferred`, `doc_only_integrations_24h`, `doc_only_budget_24h`, `doc_budget_exceeded`, `items_considered` via `.get`. No schema check, no row aging. All six keep name and meaning; `passes` is additive and ignored there. Fewer duplicates means the bounded `ledger_tail` reaches further back, which only helps that panel. Pinned by `test_dashboard_fields_are_unchanged_and_passes_is_additive`.

## Not changed

The guard, `N=5`, the classifier, `count_doc_only_integrations_24h`, every existing field, `items_considered`'s denominator (#1226 fix intact — its test still passes).

## Tests — the four acceptance criteria (`tests/test_demand.py::TestIssue1238DocBudgetRowVolume`)

1. `test_unchanged_state_deferring_nothing_writes_one_row_per_process` — three passes, unchanged, nothing deferred → one row, `passes: 1`.
2. `test_every_deferring_pass_writes_with_the_count` — three deferring passes → three rows, each `doc_only_deferred: 1`.
3. `test_every_transition_direction_writes` — `exceeded` 0→1, 1→0 and `blind` 0→1, 1→0, each state run twice → five rows (heartbeat + four transitions), `passes` = `[1, 2, 2, 2, 2]`.
4. `test_reached_and_deferring_nothing_is_distinct_from_never_reached_and_no_data` — no row / `deferred: 0` row / `ledger_blind: true` row are three different things.

Plus `test_dashboard_fields_are_unchanged_and_passes_is_additive`. The three pre-existing doc-guard row tests (#1108, #1175, #1226) pass unchanged — they all defer, so they always write.

## Full suite (Windows, `python -m pytest tests/`)

**4 failed, 2970 passed, 17 skipped** (1164s), CI 3.11/3.12/3.13 pass.

```
```

Expected baseline is exactly these 4 `fcntl.flock`-dependent bridge tests (set by #1235). CI (ubuntu) 3.11/3.12/3.13: see checks.

## Files

- `nanobot/runtime/demand.py` — `_doc_budget_last_row` memo + `_doc_budget_row_due`; the write site gated on it and carrying `passes`.
- `tests/test_demand.py` — `TestIssue1238DocBudgetRowVolume` (5 tests).

### DIFF
--- a/nanobot/runtime/demand.py
+++ b/nanobot/runtime/demand.py
@@ -397,6 +397,39 @@ def doc_only_budget_24h() -> int:
         return _DEFAULT_DOC_ONLY_24H_BUDGET
 
 
+# #1238: the ``doc_only_budget`` ledger row was 27% of the live ledger because
+# ``collect_demand`` runs two-to-three times per bridge invocation (gate probe
+# + context build) and appended the row on every pass, restating an unchanged
+# state. What carries information is a *change* of state plus one heartbeat
+# per cycle, so the row is written when the state differs from the last row
+# written in this process, always when something was deferred, and once at
+# process start. The bridge is a ``Type=oneshot`` unit — one process per
+# invocation — so "this process" is "this cycle", the same reasoning as
+# ``llm_proposer._record_idle``. Process memory is the only thing cheaper than
+# the row it suppresses: no state file, no scan, no read on the write path.
+# Keyed by state_dir so independent state roots (test fixtures) never share
+# a memo. Value: ``[state_tuple, passes_folded_since_last_written_row]``.
+_doc_budget_last_row: dict[str, list] = {}
+
+
+def _doc_budget_row_due(state_dir: Path, state: tuple[bool, bool, bool]) -> int:
+    """Return the number of passes a row written now would stand for, or 0
+    when the pass restates the last row and should be folded into the next.
+
+    ``state`` is ``(doc_budget_exceeded, ledger_blind, doc_only_deferred > 0)``.
+    A pass that deferred anything is always due — a deferral is never a no-op.
+    """
+    key = str(state_dir)
+    memo = _doc_budget_last_row.get(key)
+    deferred = state[2]
+    if memo is not None and not deferred and memo[0] == state:
+        memo[1] += 1
+        return 0
+    passes = 1 + (memo[1] if memo is not None else 0)
+    _doc_budget_last_row[key] = [state, 0]
+    return passes
+
+
 def count_doc_only_integrations_24h(state_dir: Path, now: datetime | None = None) -> int:
     """Count doc-only ``outcome: success`` rows in the last 24 h across the live
     ledger and its rotated day archives (``state_access.ledger_window``, #1175).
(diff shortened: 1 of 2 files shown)


ozand:
Reviewed and merging. The design avoids the trap the issue names, and the tests prove it rather than assert around it.

**The trap was recreating #1108.** Writing only when the budget is exceeded would make absence ambiguous again, which is the defect the row exists to fix. `test_reached_and_deferring_nothing_is_distinct_from_never_reached_and_no_data` pins all three states against three separate state dirs — no row at all, one row with `deferred=0, exceeded=True`, and one with `ledger_blind=True`. That is the acceptance criterion that actually mattered and it is tested directly.

**The cost constraint is met and the reasoning is checkable.** Process memory keyed by `state_dir`, no state file, no scan, no read on the write path. The `Type=oneshot` argument — one process per bridge invocation, so "this process" is "this cycle" — is correct and cites `llm_proposer._record_idle` as precedent rather than asserting it. Keying by `state_dir` so test fixtures never share a memo is the detail that would have caused flaky cross-test bleed if it had been missed.

**Compatibility checked in the right direction.** `test_dashboard_fields_are_unchanged_and_passes_is_additive` pins the six field names `ozand/eeebot-ops-dashboard` #201 reads and treats `passes` as additive. That is the correct burden — the consumer is in another repository and cannot defend itself here.

Verified independently: 181 tests in `tests/test_demand.py` pass locally; CI green on 3.11/3.12/3.13.

**One note, not a defect in this PR.** Its diff against `main` showed `nanobot/runtime/state_path_invocations.py` as −105 lines, which is the module #1241 added. The branch was cut at `97cb973a`, before that merge, so the deletion is an artifact of the stale base and a three-way merge keeps the module. Third time today an agent branch has shown someone else's merged work as a deletion (#1229, #1233, this one); each time benign, each time a check. The structural fix is on my side — merge in ascending order of blast radius and rebase before merging — not on yours.

Live verification owed after deploy: `doc_only_budget` rows per `outcome` row over a natural window, against today's 140:49.

Issue #1228, opened 2026-09-03
hypotheses/durable.json is frozen while the strategist is paused (#999), but #1222 removed its orphan tag: the writer registry cannot tell a declared writer from an invoked one


ozand:
> **Corrected 2026-09-03.** The first version of this issue asked whether the timer was disabled deliberately. It was, and the answer was already public: the operator paused it on 2026-09-02, recorded on #999 with three explicit re-enable conditions. I filed the question without reading that issue's comments — while citing the very ADR (#1173) whose audit produced the pause. The question is withdrawn; what survives is below, and it is narrower.

## What is actually wrong

`hypotheses/durable.json` is **not** written by a live writer, and PR #1227 (issue #1222) removed its `orphan:` tag on the grounds that it is.

The deliberate pause on #999 is not in dispute and is the right call — the strategist was producing confident advice from three structurally dead inputs, and that advice was being integrated. The defect is that #1222's audit did not see the pause, so one of its eight rows got a verdict that does not hold on the host:

| | |
|---|---|
| #1222 row verdict | "not orphaned — writer is `strategist → hypothesis_backlog.append_hypotheses`, daily 00:00 UTC" |
| Host, 2026-09-03 | `eeebot-strategist.timer` `disabled`; no `timers.target.wants` symlink; no cron entry for any user |
| `durable.json` mtime | `2026-09-02 03:00` — matches `OnCalendar=*-*-* 03:00:00` exactly; the 09-03 run did not happen |

The writer *function* exists, which is why the registry test passes. Its invoker is switched off, which is why the file is frozen. That is the same distinction the same PR applied correctly to `self_evolution/` (writers exist, nothing has run since 2026-06-20 → #1224). This row escaped it because it was checked by mtime and the mtime was under a day old at that moment.

## Why it slipped, and the gap that lets it slip again

`STATE_PATH_WRITERS` (#1219) proves a declared writer's symbol resolves. Nothing proves anything *calls* it. The registry is not lying — it answers "does this writer exist", and the question being asked of it is "is this path still being written". Those diverge the moment a timer is disabled, and they will diverge silently for as long as the pause lasts.

`systemctl list-timers` is also not the check: a disabled timer does not appear there at all, so its absence reads as "no such unit" rather than "switched off". `systemctl list-unit-files` is what distinguishes them.

## Impact, stated rather than implied

[...]

PR #1241, opened 2026-09-03
feat(state): check declared writer invocation (#1228)

ozand:
-

### DIFF
--- a/nanobot/runtime/state_path_invocations.py
+++ b/nanobot/runtime/state_path_invocations.py
@@ -1,0 +1,105 @@
+"""Check invocation contracts for selected state-path writers (#1228).
+
+This is an operator/test check, not a hot-path runtime dependency. Systemd
+writers are checked with ``list-unit-files`` because disabled timers are absent
+from ``list-timers``. Direct bridge writers are represented as per-cycle
+invokers and do not need a systemd query.
+"""
+from __future__ import annotations
+
+import json
+import subprocess
+from typing import Any, Callable
+
+from nanobot.runtime import state_paths
+
+CommandRunner = Callable[[list[str]], subprocess.CompletedProcess[str]]
+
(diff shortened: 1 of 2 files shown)


ozand:
Reviewed and merging. This does the thing the issue asked for and does it in the right place.

**The instrument is correct.** `list-unit-files` rather than `list-timers` — that distinction is the whole reason the 36-hour pause was invisible: a disabled timer does not appear in `list-timers` at all, so its absence reads as "no such unit" rather than "switched off". The parse takes `STATE` and `PRESET`, which is what makes `disabled` against a preset of `enabled` legible as a deliberate operator action.

**The states are separated the way #1173 asks for.** `unavailable` (systemctl unreadable) is not `absent` (unit not listed) is not `disabled` (listed, switched off) is not `scheduled`. A check that collapsed those would have the same defect as the readers that ADR is about.

**It cannot fail for environmental reasons.** Every test passes a runner, so nothing reaches real `systemctl`; 15 pass on Windows, CI green on 3.11/3.12/3.13. A host-shaped check wired into a suite that runs on machines without systemd would have been the obvious way to get this wrong.

**Scope is honest.** `WRITER_INVOKERS` declares two segments — `hypotheses` (systemd) and `ledger` (direct, per-cycle) — which is exactly the positive and negative pair the criteria required, and no more. It does not attempt to map every writer, and it should not: the mapping is the expensive part and each entry needs its own evidence.

Verified independently rather than taken from the report: the three test names cover flagged-when-disabled, quiet-when-enabled, and distinguishes-absent; `check_writer_invocations` is not imported on any hot path; the module docstring states it is an operator/test check.

What this does not do, and what remains true from #1228: two segments are audited, not twenty-four. The gap is now closed for the case that caused it and open for the rest — which is the right increment, not a shortfall.

PR #1242, opened 2026-09-03
fix(deploy): keep rollback armed through bridge activation

ozand:
Clean replacement for #1237/#1239; those PRs remain untouched.

Base: current `origin/main` `97cb973a`. Deploy-only scope:
- keep the rollback trap armed through dashboard and bridge activation;
- enforce exact dashboard listener ownership, command line, and bounded API semantics;
- preserve disabled/masked/not-found behavior;
- retain executable mock-backed failure-path tests.

Files: `host/eeepc/scripts/deploy_release.sh`, `tests/test_deploy_integrity.py`, `tests/test_deploy_release.py`.

No host rollout included. Awaiting CI and independent pi-1 approval.

### DIFF
--- a/host/eeepc/scripts/deploy_release.sh
+++ b/host/eeepc/scripts/deploy_release.sh
@@ -78,10 +78,12 @@ BRANCH=$(git -C "$REPO_ROOT" rev-parse --abbrev-ref "$COMMIT" 2>/dev/null || ech
 SUBJECT=$(git -C "$REPO_ROOT" log -1 --format="%s" "$COMMIT")
 TIMESTAMP=$(date -u +%Y%m%dT%H%M%SZ)
 RELEASE_NAME="${TIMESTAMP}-canonical-${COMMIT}"
+FULL_COMMIT="$(git -C "$REPO_ROOT" rev-parse "$COMMIT^{commit}")"
 ARCHIVE="/tmp/eeebot-release-${RELEASE_NAME}.tar.gz"
 
 log "repo:    $REPO_ROOT"
 log "commit:  $COMMIT ($SUBJECT) ($BRANCH)"
+log "full commit: $FULL_COMMIT"
 log "release: $RELEASE_NAME"
 log "host:    $HOST"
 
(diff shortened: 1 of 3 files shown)


ozand:
Merging this one and closing #1239, because the difference between them is one line and that line is the whole point.

#1239 disarmed the rollback trap **before** the bridge restart, two lines under a comment claiming it stayed armed through it:

```
trap - ERR
# ... The activation rollback trap remains installed so a bridge restart failure
# restores the previous release.
sudo systemctl restart eeepc-self-evolving-subagent-bridge.service
```

A bridge restart failure would have left the new release active with no rollback — the exact window #1236 exists to close. This PR moves `trap - ERR` after the restart, so the claim and the code agree.

I reviewed #1239 earlier today and missed it. I checked whether the trap could fire spuriously — `set -euo pipefail`, conditions inside `if`, risky commands guarded with `|| true` — and did not check whether it still covered the region its own comment promised. Mechanism verified, boundary not.

## What I verified here rather than took on trust

- Rebased onto current `main` (`a9a956e4`) and re-ran: `tests/test_deploy_release.py` + `tests/test_deploy_integrity.py`, 24 passed; CI green on 3.11/3.12/3.13 after the rebase.
- Trap placement after rebase is `restart` → `trap - ERR`, confirmed by reading lines 461-463 of the rebased file.
- The remote block runs under `set -euo pipefail`; every conditional is an `if [ ... ]` or `if systemctl is-active --quiet`, both of which are exempt from `ERR`, and the fallible commands carry `|| true`. So the trap fires on genuine unguarded failure after activation and not on benign non-zero.

## Limit, stated rather than implied

Neither branch sets `set -E` / `set -o errtrace`, so the `ERR` trap is **not inherited by shell functions**. A failure inside a function after activation will not trigger `rollback_remote`. That fails in the safe direction — a missed rollback, never a spurious one — but the coverage is narrower than the comments suggest. Worth a follow-up rather than a block; the current placement is strictly better than what is on `main`.

Deploying immediately after this merge: a change to `deploy_release.sh` is verified by the deploy it changes, not by the one after it.

PR #1245, opened 2026-09-03
fix(deploy): reconcile dashboard activation with current main

ozand:
Clean replacement for #1237/#1239/#1242.

Base: current `origin/main` `cfeceb7f`. Deploy-only files:
- `host/eeepc/scripts/deploy_release.sh`
- `tests/test_deploy_integrity.py`
- `tests/test_deploy_release.py`

The activation path restarts and verifies the enabled dashboard after `current` flips, checks full SHA/cwd/cmdline/:8080 ownership and bounded endpoint semantics, and keeps rollback armed through bridge restart. Disabled/masked/not-found dashboard units are preserved without endpoint claims.

Validation: `bash -n`; focused deploy tests 24 passed; Ruff; diff check. No host rollout included; awaiting CI and independent pi-1 approval.

### DIFF
--- a/host/eeepc/scripts/deploy_release.sh
+++ b/host/eeepc/scripts/deploy_release.sh
@@ -78,10 +78,12 @@ BRANCH=$(git -C "$REPO_ROOT" rev-parse --abbrev-ref "$COMMIT" 2>/dev/null || ech
 SUBJECT=$(git -C "$REPO_ROOT" log -1 --format="%s" "$COMMIT")
 TIMESTAMP=$(date -u +%Y%m%dT%H%M%SZ)
 RELEASE_NAME="${TIMESTAMP}-canonical-${COMMIT}"
+FULL_COMMIT="$(git -C "$REPO_ROOT" rev-parse "$COMMIT^{commit}")"
 ARCHIVE="/tmp/eeebot-release-${RELEASE_NAME}.tar.gz"
 
 log "repo:    $REPO_ROOT"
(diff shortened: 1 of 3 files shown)


ozand:
Closing: merging this would reintroduce the defect that broke the deploy pipeline today.

This is the fourth branch for #1236 (`-activation` → `-reconciled` → `-final` → `-final-clean`). `-final` was merged as #1242. Diffed against current `main`, the only thing this branch adds is:

```
+    DASHBOARD_CWD="$(readlink "/proc/$DASHBOARD_PID/cwd" 2>/dev/null || true)"
+    DASHBOARD_CMDLINE="$(tr "\0" " " < "/proc/$DASHBOARD_PID/cmdline" 2>/dev/null || true)"
```

Those are the un-sudo'd reads. `/proc/<pid>/cwd` belongs to the unit's user, so the read returns empty and the check reads that as a mismatch — that is #1246, and it aborted two deploys after activation today. #1247 fixed it; this branch would put it back and delete the test that pins it (`tests/test_deploy_release.py` −25).

It would also revert #1243's reward-label fix (`scripts/eeebot_dashboard.py` −11, `tests/test_eeebot_dashboard_truth.py` −35), because the branch was cut before those merged.

Nothing here is lost by closing: everything this branch was for is on `main` via #1242, #1247 and #1249. See #1248 for the rule that now covers this — one open PR per issue, rework goes to the same branch.

Issue #1254, opened 2026-09-03
Does a gateway outage count as loop waste? proposer_reject/llm_unavailable appears to increment wasted_attempts


ozand:
## The question

`scorecard._loop_section` appears to count a proposer rejection caused by the LLM gateway being unreachable as a **wasted attempt**. Fed the two rows below:

```python
rows = [
    {"phase": "proposer_skip",   "reason": "dedup_exhausted",  "demand_id": "d-1"},
    {"phase": "proposer_reject", "reason": "llm_unavailable",  "demand_id": "d-2"},
]
_loop_section(rows) ->
    attempts 1, wasted_attempts 1, proposer_rejects 1, skips_by_class {}
```

A rejection because the dependency was down is not the loop wasting an attempt — nothing was spent, and there was nothing to spend it on. If that is what the counter does, then `wasted_attempt_rate` inflates during every gateway outage, and anything conditioned on it reads a real outage as loop inefficiency.

That matters today specifically: both pi agents hit sustained `Connection error` from the LiteLLM gateway this afternoon, and `an/gemini-3.7-flash-high` and `cl/gpt-5.6-luna` were both unreachable for a period — so if the runtime saw the same thing, today's `wasted_attempt_rate` is measuring the gateway, not the loop.

## Why this is a question and not a finding

**The fixture above is not trustworthy and I am not claiming a defect from it.** Two synthetic rows with no `cycle_id`, and `skips_by_class` came back `{}` — which suggests the rows do not match what the function expects, so the counters may be reacting to a malformed input rather than to the case they are named for. Building a conclusion on a hand-made fixture is exactly how I produced a false finding on #1171 earlier today.

It surfaced while checking whether an abandoned branch (`feat/998-proposer-efficiency-clean`, now deleted) carried a test worth keeping. That test — `test_ac4_scorecard_excludes_skips_and_unavailable` — asserted `sc.get("duplicate_failure_skips", 0) == 0` against keys `_loop_section` does not return, so it was vacuous and was not salvaged. But its *name* says someone once intended this exact guarantee, which is why the question is worth answering properly rather than dropping.

## What to establish

1. Read `_loop_section` and say, from the code, which ledger rows increment `attempts`, `wasted_attempts` and `proposer_rejects`, and whether `reason == "llm_unavailable"` is excluded anywhere.
[...]

Issue #977, opened 2026-08-25
fix(runtime): out_of_band_main_detected pollutes repeat_failure_rate — operator merges manufacture goal-gap demand


ozand:
## Context (self-contained)

Live observation (2026-08-25): when an operator merges a PR into the instance repo's main between cycles, the #846 out-of-band guard correctly refuses to integrate the in-flight cycle — journal: "integrity: origin/main moved out-of-band 7b1ab40ece09->71fde9b6f3ec — a push bypassed the gate (#846)" then "integrate SKIPPED (out_of_band_main_detected)". That is DESIGNED, healthy behaviour and the operator's merge is legitimate.

But those events land in the cycles ledger as failures ("failed: out_of_band_main_detected" — two such rows visible in the proposer context digest that day) and are then counted by the failure-rate metric: the live proposer context carried a demand item "[goal-gap-5d4d5a9dc822] (goal-gap) goal gap: repeat_failure_rate (V1) — evidence: repeat_failure_rate=0.56 is above max target 0.3 over the last 7d window". The 0.56 spike coincides with the operator-merge day. Net effect: operator activity inflates the agent's failure metric, which manufactures a goal-gap demand item, which assigns the agent to "fix" a failure rate it did not cause — a self-inflicted wild goose chase.

Locate where repeat_failure_rate (and any sibling failure-rate metrics feeding goal-gap demand) classifies ledger rows, and where repeat-failure suppression counts streaks. The reason string `out_of_band_main_detected` is already written to the ledger rows by the bridge.

## Task

Exclude `out_of_band_main_detected` outcomes from failure-rate metrics and from repeat-failure streak counting: they are integration environment events, not agent failures. Keep them visible in the ledger and journal unchanged (forensics stays). If other non-agent outcome reasons exist in the same family (e.g. harness/infra reasons that do not reflect executor work quality), handle them via one small allowlist/denylist of excluded reasons rather than a string special-case — but do NOT reclassify genuine gate rejections (smoke failures, mutation violations stay failures).

## Acceptance Criteria

- [ ] A ledger with N out_of_band_main_detected rows and zero genuine failures yields repeat_failure_rate that ignores them (test).
- [ ] Genuine failure reasons (smoke fail, mutation violation) still count (test).
- [ ] Repeat-failure streak/suppression logic does not count out_of_band rows toward streaks (test).
[...]

PR #1256, opened 2026-09-03
fix(scorecard): count each wasted attempt once and include failed outcomes (#1255)

### LINKED ISSUE #1255: wasted_attempt_rate double-counts self_dedup rejects and omits failed outcomes — formula does not match its #1055 definition
## Problem

`scorecard._loop_section` (`nanobot/runtime/scorecard.py:688-690, 731`) computes

```python
repeat_failures = duplicate_failure_skips + self_dedup_rejects
attempts        = proposals + proposer_rejects
wasted_attempts = repeat_failures + proposer_rejects
wasted_attempt_rate = wasted_attempts / attempts
```

Two things do not match the metric's own description in PR #1055 ("`wasted_attempt_rate` for failed outcomes plus rejects"):

1. **`self_dedup` rejects are counted twice** — once inside `repeat_failures`, once inside `proposer_rejects`.
2. **`failed` outcomes are not counted at all** — only `skipped-*` outcomes with `reason == recent_duplicate_failure` reach the numerator.

Measured against the live 7-day window (host `scorecard/latest.json`, computed 2026-09-03T17:31Z, reproduced row-for-row from the ledger in #1254):

| quantity | value |
|---|---|
| attempts | 602 |
| proposer_rejects | 34 (31 `self_dedup`, 3 `llm_unavailable`) |
| duplicate_failure_skips | 10 |
| wasted_attempts as coded | 75 = 10 + 31 + 34 |
| wasted_attempts with self_dedup counted once | 44 |
| wasted_attempt_rate as coded / deduplicated | 0.1246 / 0.0731 |

The rate can exceed 1.0 in a window where every attempt is a self-dedup reject.

## What is established about readers (from #1254)

- Not in `_TARGETS`; no gap, demand, or futility record derives from it.
- The only consumer is `strategist_inputs.scorecard_input`, which passes the whole `latest.json` into the strategist prompt as context.
- No test asserts `wasted_attempts` or `wasted_attempt_rate`.
- #1211, the hypotheses lane, `evolution_tree`, `state.scorecard_summary`, `loop_explorer` and the ops dashboard do not read either key.

## Decision needed before any code

Which definition is wanted — and whether the metric is wanted at all, given zero readers:

- (a) PR #1055 text: `failed` outcomes + all proposer rejects, each counted once.
- (b) repeat-failure semantics: `duplicate_failure_skips` + all proposer rejects, counted once (drop the double count only).
- (c) remove the two keys; `scripts/loop_metrics_report.py` already gives `proposer_rejects.by_reason` and the ledger keeps every class separable.

Related: gateway-outage rejects (`llm_unavailable`, and pre-#1014 `sizing_rejected` with `reply=None`) are included under every option above by design unless the decision says otherwise; #1254 established that they are 3 rows since 2026-08-26 and 811 rows before.

## Acceptance Criteria

- [ ] Definition chosen and written into the `_loop_section` return-dict comment.
- [ ] Formula matches it; a test pins `wasted_attempts` against a fixture with at least one `self_dedup` reject, one `failed` outcome and one `recent_duplicate_failure` skip.
- [ ] If the number changes meaning, the `history.jsonl` discontinuity is noted in the PR body.

ozand:
Closes #1255. Found while closing #1254.

## Defect

`scorecard._loop_section` computed `wasted_attempts = repeat_failures + proposer_rejects`. `repeat_failures` already contains `self_dedup_rejects`, which are a subset of `proposer_rejects`, so every self-dedup reject was counted twice. `failed` outcomes were never counted, although PR #1055 defined the metric as "failed outcomes plus rejects". In a window where every attempt is a self-dedup reject the published rate was 2.0.

## Fix

```python
wasted_attempts = duplicate_failure_skips + failed_outcomes + proposer_rejects
```

Each attempt contributes at most once. `repeat_failures` / `repeat_failure_rate` (the `_TARGETS` metric, #977/#1014 semantics) are untouched. `partial` outcomes are still not waste — the #1055 definition names `failed` only.

## What the strategist's input becomes (same live window, `history.jsonl` discontinuity)

Host `state/scorecard/latest.json` computed 2026-09-03T17:31:17Z, 7-day window since 2026-08-27T17:31Z, recomputed row-for-row from the live ledger + archives (read-only):

| quantity | before | after |
|---|---|---|
| attempts (proposals + rejects) | 602 | 602 |
| proposer_rejects (31 `self_dedup`, 3 `llm_unavailable`) | 34 | 34 |
| duplicate_failure_skips | 10 | 10 |
| failed outcomes (of 565 outcome rows) | not counted | 9 |
| self_dedup counted | twice (31 extra) | once |
| **wasted_attempts** | **75 of 602** | **53 of 602** |
| **wasted_attempt_rate** | **0.1246** | **0.0880** |
| repeat_failures / repeat_failure_rate | 41 / 0.0681 | 41 / 0.0681 (unchanged) |

Between 17:31Z and my recompute the live ledger gained 2 proposals in-window (604 attempts, rate 0.0877); the table uses the 602 the scorecard published. The next scorecard recompute after roll-out will step `wasted_attempt_rate` down by roughly 0.037 in `history.jsonl` with no change in loop behaviour — that step is this fix, not the loop.

## Readers re-checked before changing the meaning

- `_TARGETS` (scorecard.py:333): only `repeat_failure_rate` from the loop section. Not `wasted_*`. No gap, demand or futility record derives from it.
- `strategist_inputs.scorecard_input`: passes whole `latest.json` into the strategist prompt + generic trend series. Only consumer; prose context.
- `evolution_tree.py:337`, `state.py:1030`, `loop_explorer.py:325`: read `integrations`, `confirmed_integrations`, `repeat_failure_rate`, `idle_share`, `confirmed_integration_ratio` only.
- `hypothesis_backlog.py`, `hypothesis_verdict.py`, `goal_gap_futility.py`, `demand.py`, `scripts/loop_metrics_report.py`: no read of `wasted_attempts` / `wasted_attempt_rate`.
- `eeebot-ops-dashboard`: no occurrence of `wasted`.
- Tests: none asserted either key before this PR.

## Tests

`test_wasted_attempt_rate_cannot_exceed_one` — 3 self_dedup rejects, nothing else. Pre-fix failure:

```
>       assert loop["wasted_attempts"] <= loop["attempts"]
E       assert 6 <= 3
```

`test_wasted_attempts_composition` — one term at a time on a fixed base: self_dedup reject → 1, llm_unavailable reject → 1, failed outcome → 1, recent_duplicate_failure skip → 1, already_done skip → 0, success → 0; then all together 4 of 5 attempts with `repeat_failures` still 2. Pre-fix failure:

```
>       assert wasted([{"phase": "proposer_reject", "reason": "self_dedup", "ts": _iso(5)}]) == 1
E       AssertionError: assert 2 == 1
```

`tests/test_scorecard.py`: 66 passed. Full suite result posted as a comment when the run finishes (baseline 4 known failures: bridge tag fail-open + 3 `TestAcquireBridgeLock`).

Pre-existing, not touched: `ruff` F401 on `loguru.logger` in scorecard.py:89 is on `main` already; CI does not run ruff.

### DIFF
--- a/nanobot/runtime/scorecard.py
+++ b/nanobot/runtime/scorecard.py
@@ -618,6 +618,7 @@ def _loop_section(
     proposer_rejects = 0
     self_dedup_rejects = 0
     duplicate_failure_skips = 0
+    failed_outcomes = 0
     skips_by_class: dict[str, int] = {}
     # #800 churn split: cycles whose proposed row served a decay demand
     # (demand_id "decay-…", the #760 traceability field). A success outcome
@@ -684,10 +685,17 @@ def _loop_section(
                 # are not failures and must not feed repeat_failure_rate.
                 if str(row.get("reason") or "").strip() == "recent_duplicate_failure":
                     duplicate_failure_skips += 1
+            elif outcome == "failed":
+                failed_outcomes += 1
     cycleish = idle_rows + outcome_rows
     repeat_failures = duplicate_failure_skips + self_dedup_rejects
     attempts = proposals + proposer_rejects
-    wasted_attempts = repeat_failures + proposer_rejects
+    # #1055 definition: failed outcomes plus rejects. Each attempt counts at
+    # most once — self_dedup rejects are already inside proposer_rejects, so
+    # they must NOT be re-added via repeat_failures (#1255: that double
+    # count published 75 of 602 where the definition gives 53, and let the
+    # rate exceed 1.0 in a window of only self_dedup rejects).
+    wasted_attempts = duplicate_failure_skips + failed_outcomes + proposer_rejects
     return {
         # Value-bearing integrations ONLY (#800) — the fitness numerator
         # consumed by the _TARGETS gap analysis. Archival churn is reported
--- a/tests/test_scorecard.py
+++ b/tests/test_scorecard.py
@@ -123,6 +123,72 @@ class TestLoopSection:
         assert loop["repeat_failure_rate"] == round(1 / 3, 4)
         assert loop["skips_by_class"] == {"skipped-duplicate": 3}
 
+    def test_wasted_attempt_rate_cannot_exceed_one(self, tmp_path):
+        """#1255: a window where every attempt is a self-dedup reject. The
+        pre-fix formula (repeat_failures + proposer_rejects) counted each
+        self_dedup reject twice and produced wasted_attempts 6 of 3 attempts —
+        a rate above 1.0 is self-refuting."""
+        state_dir = tmp_path / "state"
+        _write_ledger(state_dir, [
+            {"phase": "proposer_reject", "reason": "self_dedup", "ts": _iso(3)},
+            {"phase": "proposer_reject", "reason": "self_dedup", "ts": _iso(2)},
+            {"phase": "proposer_reject", "reason": "self_dedup", "ts": _iso(1)},
+        ])
+        loop = scorecard.compute_scorecard(state_dir, None, force=True)["loop"]
+        assert loop["attempts"] == 3
+        assert loop["wasted_attempts"] <= loop["attempts"]
+        assert loop["wasted_attempts"] == 3
+        assert loop["wasted_attempt_rate"] == 1.0
+
+    def test_wasted_attempts_composition(self, tmp_path):
+        """#1255 pins the #1055 definition — "failed outcomes plus rejects" —
+        term by term, so a formula wrong in two compensating ways cannot pass
+        on the aggregate alone: one self_dedup reject contributes exactly
+        once, one failed outcome contributes once, one recent_duplicate_failure
+        skip contributes once, and a success contributes nothing."""
+        state_dir = tmp_path / "state"
+        base = [
+            {"phase": "proposed", "cycle_id": "c-ok", "ts": _iso(9)},
+            {"phase": "outcome", "cycle_id": "c-ok", "outcome": "success", "ts": _iso(8)},
+        ]
+
+        def wasted(extra: list[dict]) -> int:
+            _write_ledger(state_dir, base + extra)
+            return scorecard.compute_scorecard(state_dir, None, force=True)["loop"]["wasted_attempts"]
+
+        assert wasted([]) == 0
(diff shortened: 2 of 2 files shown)


ozand:
Merging. Verified independently rather than taken from the report.

**The fix is the right shape.** `wasted_attempts = duplicate_failure_skips + failed_outcomes + proposer_rejects` — each attempt contributes at most once, and `failed` outcomes enter the numerator for the first time, which is what #1055's definition said all along. `repeat_failures` / `repeat_failure_rate` are left alone, so the `_TARGETS` metric with #977/#1014 semantics does not move; that separation is the part that makes this safe.

**It corrects my arithmetic too.** On #1255 I gave the deduplicated figure as `44 = 10 + 34`. It is `53 = 10 + 9 + 34` — I dropped the failed-outcomes half of the definition while checking the double-count half. The live rate goes `0.1246 → 0.0880`, not `0.0731`.

**Mutation-checked:** restoring `wasted_attempts = repeat_failures + proposer_rejects` fails exactly two tests — `test_wasted_attempt_rate_cannot_exceed_one` and `test_wasted_attempts_composition` — and nothing else. Those are the two properties worth pinning: the aggregate alone would pass on a formula wrong in two compensating ways, and a rate that can exceed 1.0 is self-refuting and therefore the cheapest possible detector for this class.

`tests/test_scorecard.py`: 66 passed. CI green on 3.11/3.12/3.13.

**On the changed input to the strategist.** Its prompt has been carrying an inflated waste rate — including the 12:37Z run that produced `hyp-0022`. That advice did not depend on this number and happens to be right, but the input was wrong at the time and now changes under it. Worth watching the next 03:00 run for advice that reads as a response to loop waste; the rate it sees drops by a third.

Owed after deploy: `wasted_attempt_rate` on the live host recomputed, quoted here, against today's `0.1246`.

ozand:
Full suite in the worktree (`python -m pytest -q -n 4`): 4 failed, 2985 passed, 17 skipped in 9m15s. The 4 failures are exactly the known baseline: `test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle` and the three `test_bridge_locking.py::TestAcquireBridgeLock` cases (Windows flock). CI green on 3.11/3.12/3.13.

Issue #1208, opened 2026-09-02
The loop applies a fix-then-add-JSON-output template to its own validators — measure whether the enhancement halves are ever used


ozand:
## Problem

The loop appears to be applying a template to its own validator scripts rather than responding to observed need. Across at least eight scripts, the defect lane produced the same pair: fix the broken thing, then extend it with a CLI and JSON output.

Found while closing #1038, by listing the distinct `defect-*` demand ids that share a `target_path` (live ledger plus all 52 rotated archives):

```
validate_open_encoding.py        Fix AST analysis and text-mode detection
                              →  Extend with structured JSON output and targeted path scanning
validate_deprecation_status.py   Fix deprecation check logic and import parsing
                              →  Extend with CLI arguments and structured scan results
validate_timezone_aware_datetime.py  Fix AST inspection and execution exit code
                              →  Add target path filtering and structured JSON reporting
validate_unused_imports.py       Fix import resolution and exit code handling
                              →  (same shape)
check_style.py                   Optimize traversal and AST checking to finish within timeout
                              →  Add JSON output format and path filtering options
prevent_repeat_failures.py       Add mutation surface validation
                              →  Extend with CLI interface and structured ledger-based output
archive_old_reports.py           Gracefully exit 0 on execution of an archived script
                              →  Add dry-run mode, customizable retention days, and JSON summary
```

The first half of each pair is a response to a real failure. The second half is not obviously a response to anything: nothing asked for JSON output on six validator scripts, and no consumer of that output has been identified.

## Why this is worth measuring rather than assuming

This is not a dedup defect — the ids are correctly distinct, which is what #1038 established. It is a question about what the loop *chooses* to build when it is not fixing something, and that is a harder and more important question than any of the reader defects fixed under #1173.

Two readings, and they call for opposite responses:

1. **The enhancements are used.** Something downstream consumes the JSON, the flags get passed, the dry-run mode gets exercised. Then this is the loop building infrastructure for itself and the pattern is a strength.
2. **The enhancements are never used.** Then the loop has spent integrated cycles on a template it recognises rather than on need, and every one of those cycles displaced a real defect. That is the same family as the `AGENTS.md` accretion in #1188 — cheap, always-succeeding work that crowds out expensive useful work — and it would deserve the same kind of fix.

The measurement decides which, and it is cheap.

## What to measure

[...]

PR #1258, opened 2026-09-03
feat(validator-harness): record why findings_count is None (#1208 step 1)

ozand:
Step 1 of #1208 (see the verification comment there naming the two steps). Step 2 — the schema decision — is deliberately **not** in this PR.

## What was wrong

`_parse_findings_count` returned `None` for three different situations: stdout was not JSON; it was JSON but not a top-level object; it was a top-level object whose keys the parser does not read. The third is the live case: six instance validators emit `{"success","checked_count","failed_count","results"}` under `--json` on every harness run, `_FINDINGS_KEYS` is `("findings","alerts","missing","failures")`, intersection empty. On the host, 0 of 45 sidecar rows for those six carry a `findings_count`, and nothing said why — `None` from "unreadable shape" was byte-identical to `None` from "ran fine, printed text".

## What this changes — recording, not deciding

Each `state/validator_harness/last_runs.jsonl` row now carries four additive fields:

| field | values | meaning |
|---|---|---|
| `json_flag` | bool | whether `--json` was appended (script declared it) |
| `findings_parse` | `matched` / `not_json` / `not_object` / `unrecognised_keys` | which branch produced `findings_count` |
| `stdout_keys` | sorted list, ≤24, only on `unrecognised_keys` | the top-level keys the script actually emitted |
| `stdout_truncated` | bool | head capture hit `_MAX_OUTPUT_BYTES` (64 KiB) — a long JSON document then reads as `not_json`, and this says the cap did it |

`findings_count` itself is computed exactly as before (`_classify_findings` returns the same integer or `None` from the same branches). `demand._validator_defect_items` reads `exit_code` / `findings_count` / `harness_contract` / `stderr_tail` and none of the new fields, so **what becomes demand is unchanged**: exit 0 + `None` → no item; exit non-zero + `None` → the same exit-code defect as before. `test_unrecognised_json_keys_are_recorded_and_do_not_change_demand` pins both halves of that with the live key shape.

`_FINDINGS_KEYS` is not widened. Whether `failed_count` is a findings count by another name, and for how many scripts, is what these rows will now show — step 2 decides from that data.

Cost on the normal path: one `sorted()` over top-level keys, only when an object parses and matches nothing; one `len()` compare. No new I/O, no new subprocess, row grows by ~80–200 bytes against the 16 KiB per-record cap.

## The three states as a real harness run now records them

Ran the instance repo at `origin/main` (`1626dcbf`) through the branch's `_accepts_json_flag` / `_classify_findings` with the harness's 64 KiB head capture — same command, same cwd shape, read-only:

| script | allowlisted | json_flag | exit | stdout | findings_parse | stdout_truncated | stdout_keys |
|---|---|---|---|---|---|---|---|
| validate_open_encoding.py | yes | true | 0 | 13.6 KB | **unrecognised_keys** | false | checked_count, failed_count, results, success |
| validate_deprecation_status.py | yes | true | 0 | 16.8 KB | **unrecognised_keys** | false | + warnings_count |
| validate_timezone_aware_datetime.py | yes | true | 0 | 13.7 KB | **unrecognised_keys** | false | + summary |
| validate_unused_imports.py | yes | true | 0 | 14.1 KB | **unrecognised_keys** | false | + summary |
| check_style.py | yes | true | 1 | 573 KB | **not_json** | **true** | — |
| verify_imports.py | yes | true | 1 | 2.18 MB | **not_json** | **true** | — |
| prevent_repeat_failures.py | **no** (prefix) | true | 2 | 4.3 KB | not_json | false | — (usage error: requires args) |
| archive_old_reports.py | **no** (prefix) | true | 1 | 38 B | unrecognised_keys | false | error |

So the eight split three ways, not one: four are a pure key mismatch; two never reach the parser as JSON because the capture cap cuts the document — for those the schema question is moot until the output shrinks or the cap is reconsidered; two are not harness candidates at all and their `--json` modes have no runtime caller whatever their shape. Step 2 needs that split, which is why guessing keys for "eight scripts" would have been wrong for four of them.

## Tests

Extended `test_json_findings_are_counted`, `test_plain_invocation_when_script_has_no_json_flag`, `test_non_json_stdout_yields_no_findings_count` with the new fields; added `test_json_array_stdout_is_recorded_as_not_object`, `test_unrecognised_json_keys_are_recorded_and_do_not_change_demand` (exit 0 → no demand; exit 1 → same exit-code item, wording without a count), `test_stdout_cut_at_cap_is_recorded_as_truncated_not_json` (cap monkeypatched to 512 B).

`tests/test_validator_harness.py`: 108 passed, 5 skipped. Full suite result posted as a comment when the run finishes (baseline 4 known Windows failures: bridge tag fail-open + 3 `TestAcquireBridgeLock`).

Also documents the JSON expectation in the module's INVOCATION CONTRACT, which previously covered arguments, time budget, exit codes and decay but said nothing about the JSON shape the parser reads.

## Readers of `last_runs.jsonl` checked

`demand.py:990-1130` (fields above, additive-safe), `usage_evidence.py:465-467` (reads the parent log's `validator` / `finished_at`, not the sidecar), `validator_harness._select_within_budget` (byte/line caps, per-record 16 KiB), tests. No dashboard reader.

### DIFF
--- a/nanobot/runtime/validator_harness.py
+++ b/nanobot/runtime/validator_harness.py
@@ -136,6 +136,15 @@ defect when a script did not anticipate it:
 - Non-zero exit = a finding worth surfacing: it becomes ``defect`` demand
   with your stderr/stdout as evidence (see
   ``demand._validator_defect_items``).
+- ``--json`` output is read for a findings count ONLY from a top-level
+  object carrying one of :data:`_FINDINGS_KEYS` as a list/dict. Any other
+  shape yields ``findings_count: None`` (exit code decides alone) and the
+  sidecar row says why — ``findings_parse`` is ``matched`` /
+  ``not_json`` / ``not_object`` / ``unrecognised_keys`` (with the
+  ``stdout_keys`` seen), ``json_flag`` says whether the flag was passed,
+  ``stdout_truncated`` whether the :data:`_MAX_OUTPUT_BYTES` cap cut the
+  document (#1208 step 1: measured live, six validators emitted JSON daily
+  under keys this parser never read, invisibly).
 - Decay self-declaration is detected from printed output (#936): when a
    script's captured stdout+stderr contains :data:`_DECAY_DECL_RE` AND the
    script's own repo-relative path (``scripts/<filename>``), the run is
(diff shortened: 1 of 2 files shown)


ozand:
Full suite in the worktree (`python -m pytest -q -n 4`): 4 failed, 2988 passed, 17 skipped in 7m38s. The 4 are exactly the known Windows baseline: `test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle` and the three `test_bridge_locking.py::TestAcquireBridgeLock` cases. CI green on 3.11/3.12/3.13.

Issue #1244, opened 2026-09-03
The loop opens a new suffixed branch instead of updating the existing one: three PRs for #1236, two each for #1206/#998/#1000


ozand:
## What happens

When a PR the loop opened hits a conflict or a review, it does not update that branch — it opens a **new branch with a suffix** and a new PR for the same fix, leaving the old one open. Measured on `ozand/eeebot` 2026-09-03:

| issue | branches |
|---|---|
| #1236 | `fix/1236-dashboard-activation` (PR #1237) → `fix/1237-dashboard-activation-reconciled` (PR #1239) → `fix/1236-dashboard-activation-final` (PR #1242) |
| #1206 | `fix/1206-dashboard-truth` (PR #1229, merged) → `fix/1206-dashboard-truth-clean` (PR #1233) → `fix/1206-dashboard-truth-final` (PR #1234, merged) |
| #998 | `feat/998-proposer-efficiency` → `feat/998-proposer-efficiency-clean` |
| #1000 | `feat/1000-best-of-n` → `feat/1000-best-of-n-v2` |

Four distinct suffix families — `-v2`, `-clean`, `-reconciled`, `-final` — which is a habit, not four coincidences. Three of the four cases produced **two PRs open at once for the same change**, and #1236 produced three.

## Why it costs more than tidiness

**It hides real differences behind apparent duplicates.** #1239 and #1242 were byte-identical in size (+209/−18) and differed by one line — and that line was load-bearing: #1239 disarmed the deploy rollback trap *before* the bridge restart while its own comment claimed the trap stayed armed through it. If I had closed #1242 as "the same PR again", the rollback gap would have shipped. Telling a regenerated duplicate from a genuine revision costs a full diff every time.

**It leaves the merge order to chance.** Two open PRs for one fix means whichever is merged first wins, and the other's improvements are silently lost or become a conflict.

**It produced 6 abandoned branches**, 4 of them from this pattern. (Separately, 21 branches whose PR had merged were never deleted — that is the `--delete-branch` limitation in this repository, not this defect; I deleted those by hand today, remote branch count 30 → 10.)

## What to find out

This is a diagnosis task first. The question is **why** a conflict or a review comment produces a new branch instead of a push to the existing one:

- is it the bridge's branch-naming (`selfevo/cycle-<id>` and the PR lane) creating a fresh name per attempt;
- is it a prompt or skill instructing "open a clean branch" when a rebase fails;
[...]

PR #1260, opened 2026-09-03
fix(deploy): make verify-only dashboard command check exact

ozand:
Clean replacement for #1257 corrective pass; old PR remains untouched.

Base: current `origin/main` `8fac1549`. Verify-only scope only:
- require exact `/current/scripts/eeebot_dashboard.py --serve --port 8080 --host 0.0.0.0` cmdline; no permissive release-dir/current substring OR;
- preserve verify-only mutation guards and health-gate skip;
- document that the bind race is not reproducible in verify-only because it deliberately does not restart the service; normal deploy covers it with bounded 40×0.25s listener polling.

Files: `host/eeepc/scripts/deploy_release.sh`, `tests/test_deploy_release.py`.
Validation: `bash -n`, focused deploy tests (28 passed), Ruff, diff check. No host rollout included.

### DIFF
--- a/host/eeepc/scripts/deploy_release.sh
+++ b/host/eeepc/scripts/deploy_release.sh
@@ -37,10 +37,13 @@ NO_CRASH_HOLD=300
 NO_HEALTH_GATE=0
 ALLOW_DIRTY=0
 
+VERIFY_ONLY=0
+
 while [[ $# -gt 0 ]]; do
   case $1 in
     --host)           HOST="$2"; shift 2 ;;
     --dry-run)        DRY_RUN=1; shift ;;
+    --verify-only)    VERIFY_ONLY=1; shift ;;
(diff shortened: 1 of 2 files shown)


ozand:
Closing as stale. Branch cut from `8fac1549`, before #1257 merged, so its diff against `main` shows that PR's tests as deletions — a stale-base artifact, not a removal.

The work is already on `main`: #1257 delivered the verify-only mode and the single-clause cmdline check, and #1262 followed. What is left of #1250 is closed. The live blocker on `main` right now is #1264, which is a different defect — the semantic gate rejecting any host path.

Four branches exist for #1250 (`-clean`, `-clean-final`, `-final-clean`, and the merged one). #1244's rule applies: rework goes to the same branch, and a replacement closes its predecessor in the same step.

PR #1261, opened 2026-09-03
fix(deploy): enforce stale dashboard semantic gate

ozand:
Clean follow-up for #1250; prior PRs remain untouched.

Base: current `origin/main` `8fac1549`. Verify-only deployment-check scope only. The semantic gate now requires goal/task/gate/reward source metadata, rejects raw legacy reward/gate/cycle/path values, and requires WARN for stale retired sources. Current symlink and SOURCE_COMMIT reads remain privileged; exact dashboard cmdline and mutation guards are preserved.

Validation: bash -n, focused deploy tests 30 passed, Ruff, diff check. No host rollout included. Awaiting exact-head CI and independent review.

### DIFF
--- a/host/eeepc/scripts/deploy_release.sh
+++ b/host/eeepc/scripts/deploy_release.sh
@@ -505,7 +505,7 @@ import os
 health = json.loads(os.environ["DASHBOARD_HEALTH"])
 metrics = json.loads(os.environ["DASHBOARD_METRICS"])
 required_health = {"overall", "dimensions", "goal", "active_task", "reward_average"}
-required_metrics = {"goal", "active_task", "approval_gate_state", "reward_source"}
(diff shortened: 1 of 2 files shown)


ozand:
Closing as stale. Branch cut from `8fac1549`, before #1257 merged, so its diff against `main` shows that PR's tests as deletions — a stale-base artifact, not a removal.

The work is already on `main`: #1257 delivered the verify-only mode and the single-clause cmdline check, and #1262 followed. What is left of #1250 is closed. The live blocker on `main` right now is #1264, which is a different defect — the semantic gate rejecting any host path.

Four branches exist for #1250 (`-clean`, `-clean-final`, `-final-clean`, and the merged one). #1244's rule applies: rework goes to the same branch, and a replacement closes its predecessor in the same step.

Issue #1224, opened 2026-09-03
Decommission autoevolve: writers exist, duty not performed since 2026-06-20; superseded by deploy_release.sh + bridge integrate (from #1222 rows 3/5/8)


ozand:
## Found while resolving #1222 (rows 3, 5, 8)

`nanobot/runtime/autoevolve.py` — the "guarded self-evolution" release pipeline (`apply_release`, `rollback_release`, `health_check_release`, `write_guarded_evolution_state`, `write_noop_export_status`, `resolve_terminal_selfevo_issue`) with entrypoints `scripts/guarded_self_evolve.py` and `scripts/health_check_release.py` — has writers that exist in code but a duty that is not performed:

- No systemd unit, timer or crontab on eeepc invokes either script (live timers: bridge, monitor, promotion-verifier; no crontab for `eeepc-agent` or root).
- `state/self_evolution/` is an empty directory, mtime 2026-06-20 17:27 — 75 days.
- `health_check_release` probes three coordinator outputs, all frozen since 2026-08-22 02:00: `reports/evolution-*.json`, `control_plane/current_summary.json`, `goals/current.json`. On the live host it returns `ok: False` (`stale_report`) forever. A gate whose every input is frozen is a never-executed guard.
- The duty it was built for — apply a release candidate, health-check, roll back — is performed today by `host/eeepc/scripts/deploy_release.sh` (release dir + `current` symlink) and the bridge's integrate-to-main step. **Superseded.**
- The autoevolve tests are the entire Windows baseline of known failures (`test_autoevolve*.py`, `test_selfevo_export_security.py`) — symlink-dependent, never green on a developer machine.

## Ask

Decommission autoevolve as **one unit**, with a duty inventory first (the #900 lesson, third time: map duties, not files):

| duty | owner today | verdict |
|---|---|---|
| apply release candidate / flip `current` | `deploy_release.sh` | superseded |
| rollback | `deploy_release.sh` (previous release dir) | superseded |
| release health gate | `health.py` / `doctor` (ledger-based) | superseded |
| lifecycle export (`latest_issue_lifecycle.json`, `latest_noop.json`) | nobody reads them except autoevolve itself | dropped |
| `self_evolution/current_state.json` in `nanobot status` | never written on the host | dropped |
| `commit_and_push_self_evolution` (`github_ops`) | check callers before deleting | ? |

[...]

Issue #1188, opened 2026-09-02
Reflection lane spends ~4 integrated cycles/day appending to AGENTS.md: +380/-6 lines in 5 days, loaded into every prompt


ozand:
## Problem

The reflection lane is spending roughly four integrated cycles a day appending guidance to `AGENTS.md`, each under a fresh `demand_id`, with no mechanism that ever removes anything. `AGENTS.md` is loaded into every cycle's context, so this converts loop capacity directly into permanent context cost.

## Measured (live ledger + all 52 rotated archives, UTC)

Cycles whose **only** non-test change was `AGENTS.md`, `outcome: success`:

```
count:            20
distinct demand_ids: 20
lane prefixes:    reflection × 20   (no other lane)
window:           2026-08-27T16:55:13Z -> 2026-09-01T16:35:06Z
```

Twenty integrated cycles, twenty different demand ids, five days, one lane, one file.

Instance repo (`eeebot-self-evolving`), same window:

```
commits touching AGENTS.md since 2026-08-27:  22   (32 all time)
net line delta since 2026-08-27:              +380  -6
file today:                                   417 lines / 23,449 bytes
```

**91% of the current file was written in the last five days**, and six lines have ever been removed. The commits are almost pure insertion — `+19`, `+29`, `+32`, `+22`, `+32`, `+18`, `+25`, `+16`, `+19`, `+18`, `+17` — with subjects like "add command output bounding and token constraint guidance", "add early turn direct file inspection and editing guidance", "add composite context inspection guidelines".

## Why it happens

`AGENTS.md` is the single cheapest place in the repo for a reflection-derived proposal to land, and the runtime makes it so on purpose:

- `nanobot/agent/context.py:19` — `BOOTSTRAP_FILES = ["AGENTS.md", "SOUL.md", "USER.md", "TOOLS.md"]`, pulled through the full ContextBuilder pipeline for every subagent run (`nanobot/agent/subagent.py:645`).
- `nanobot/runtime/gate.py:105-108` and `nanobot/runtime/llm_proposer.py:68-71` — `_ALLOWED_EXACT_PATHS = frozenset({'AGENTS.md'})`. It is the one root-level path that bypasses the prefix check.
- `nanobot/runtime/demand.py:285` — `_DOC_ONLY_BASENAMES = {"AGENTS.md"}`.

[...]

PR #1263, opened 2026-09-03
docs(agents): decommission checklist — label every orphaned read in the same PR (#1173)

ozand:
One bullet under **Change workflow** in `AGENTS.md`. Refs #1173 (amendment of 2026-09-03, item 3: "cheaper as a line in the decommission procedure than as code"), #1224 and ozand/eeebot-ops-dashboard#205.

## Why a procedure line and not a check

The #1173 reader contract permits the failure: "empty-because-absent may proceed". A reader of a decommissioned path is empty-because-absent, so no status changes and no test fails — the panel or lane renders healthy emptiness forever. Nothing links a reader to its writer except `state_paths`, which readers do not consult. Three instances in two days, each costing a follow-up pass: #1219 (`state/research/` readers), #1222 (eight orphaned reads from the #916/#923 deletion, found twelve days later), #1224 → ops-dashboard #205 (dashboard readers of `state/self_evolution/*`, closed 2026-09-03).

## The lines (11, one bullet)

```
- **Decommission** (removing a writer, unit or module): list its duties as
  **superseded** (name what does it now), **dropped** (say why nobody needs it) or
  **orphaned** (a reader still expects the output). In the **same PR**, grep every
  path it wrote across `ozand/eeebot`, `ozand/eeebot-ops-dashboard`, the instance repo
  and the host units (`host/eeepc/systemd/`, `/etc/systemd/system`); turn each orphaned
  read into a labelled `unavailable`/`retired` naming the issue, never an empty default,
  and record the writer as `orphan:#<issue>` in `nanobot/runtime/state_paths.py`. No
  test catches this: the reader contract (#1173) lets "empty because absent" pass, so a
  retired path renders healthy emptiness forever (#1219, #1222, #1224 → ops-dashboard
  #205). Before reporting, this must show only labelled reads:
  `rg -n "<state/dir it wrote>" . ../eeebot-ops-dashboard ../eeebot-self-evolving`
```

Same shape as the #1244 bullet: a rule, the failure it prevents with issue numbers, and one checkable command.

## Existing coverage checked

No existing `AGENTS.md` rule covers decommission. The only nearby text is "What this project is" line 27 ("both entrypoints … must be preserved unless a task explicitly retires them"), which is about the two CLI entrypoints, not about readers of removed writers. `state_paths.py`'s own docstring documents `orphan:#N` (lines 29-37) — the bullet points at it rather than repeating it.

## Tests

No test pins the product `AGENTS.md` text (the #1188/#1193 tests pin the *instance* workspace file and `classify_change_tier`). Full suite result posted as a comment when the run finishes; expected baseline 4 by id (bridge tag fail-open + three `TestAcquireBridgeLock`).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -79,6 +79,17 @@ scale labels are enough).
   `docs/specs/<capability>/spec.md`. See `docs/changes/README.md`.
 - The change folder is durable *design* documentation tied to a PR — not a status
   tracker (status is the Issue). The two do not duplicate.
+- **Decommission** (removing a writer, unit or module): list its duties as
+  **superseded** (name what does it now), **dropped** (say why nobody needs it) or
+  **orphaned** (a reader still expects the output). In the **same PR**, grep every
+  path it wrote across `ozand/eeebot`, `ozand/eeebot-ops-dashboard`, the instance repo
+  and the host units (`host/eeepc/systemd/`, `/etc/systemd/system`); turn each orphaned
+  read into a labelled `unavailable`/`retired` naming the issue, never an empty default,
+  and record the writer as `orphan:#<issue>` in `nanobot/runtime/state_paths.py`. No
+  test catches this: the reader contract (#1173) lets "empty because absent" pass, so a
+  retired path renders healthy emptiness forever (#1219, #1222, #1224 → ops-dashboard
+  #205). Before reporting, this must show only labelled reads:
+  `rg -n "<state/dir it wrote>" . ../eeebot-ops-dashboard ../eeebot-self-evolving`
 
 ## Working tree and branch safety
 


ozand:
Full suite in the worktree (`python -m pytest -q -n 4`): 4 failed, 2992 passed, 17 skipped in 11m22s. The 4 are exactly the known Windows baseline: `test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle` and the three `test_bridge_locking.py::TestAcquireBridgeLock` cases. CI green on 3.11/3.12/3.13.

ozand:
Merging. Eleven lines, and every element I asked for is load-bearing.

The duty vocabulary — superseded / dropped / orphaned — teaches the classification rather than just the grep, which is what makes it a rule someone can apply to a case it does not name. The four locations are spelled out; "every repo" would not have survived contact with 2am.

The sentence that matters most is the one explaining why this cannot be a check: *the reader contract (#1173) lets "empty because absent" pass, so a retired path renders healthy emptiness forever*. Without it the rule reads like bureaucracy; with it, it is the only available defence.

I ran the `rg` command against the `state/self_evolution` case it was written from:

```
ops-dashboard/src/…/app.py:2107   # 'terminal_noop' used to be read from state/self_evolution/…
ops-dashboard/src/…/app.py:3335   # eeebot#1224; on the host state/self_evolution/ has been empty since
ops-dashboard/tests/test_guarded_evolution_system.py:52   """#205: state/self_evolution/ has had no writer since…
```

Comments explaining the retirement, a test pinning it, and docs — no unlabelled live read. The command distinguishes a finished decommission from an unfinished one, which is exactly the property a before-reporting check needs.

One observation, not a change request: `ops-dashboard/tests/test_selfevo_pr_system.py:117` still asserts a path ending in `state/self_evolution/current_state.json`. It is one of that repository's five baseline failures and fails on the Windows path separator, so it is not this rule's problem today — but it is a test asserting evidence about a decommissioned artifact, and whoever fixes that baseline entry should decide whether the assertion still means anything.

Issue #1246, opened 2026-09-03
deploy_release.sh aborts every deploy: the dashboard cwd check reads /proc without sudo and treats the empty result as a mismatch


ozand:
## What broke

`deploy_release.sh` aborts after activation on every deploy. Observed on release `20260903T141455Z-canonical-b45fb3e4`, immediately after #1242 merged:

```
[remote] restarting eeebot-dashboard.service after current activation
CRITICAL: eeebot-dashboard.service PID 11213 cwd is '',
          expected '/opt/eeepc-agent/runtimes/self-evolving-agent/releases/20260903T141455Z-canonical-b45fb3e4'
```

## The check is right and cannot see the answer

`/proc/<pid>/cwd` belongs to the unit's user, not to the deploying account. Read as `ozand` it is empty; read with sudo it is exactly what the check expects:

```
$ readlink /proc/11213/cwd
                                        <- empty
$ sudo readlink /proc/11213/cwd
/opt/eeepc-agent/runtimes/self-evolving-agent/releases/20260903T141455Z-canonical-b45fb3e4
```

So the dashboard *was* running from the new release. The verification introduced by #1242 read an unreadable path, got `''`, and treated that as a mismatch — the fail-open-vs-unavailable confusion of ADR #1173, in a deploy gate.

The `cmdline` read next to it has the same problem, and prefixing `sudo` alone does not fix it: in `sudo tr ... < /proc/<pid>/cmdline` the shell opens the redirect as the deploying user before sudo runs. The read itself has to be the sudo'd command.

## Blast radius, measured rather than assumed

The host is fine. `current` points at the new release, the dashboard answers `http=200`, the bridge and strategist timers are both active. **The rollback did not fire**, which is the `set -E` gap I recorded when merging #1242: the `ERR` trap is not inherited by shell functions, and this `exit 1` is inside one. It failed in the safe direction here — a healthy release stayed active — but that is luck, not design.

What is genuinely broken is the pipeline: every subsequent deploy aborts the same way, and does so *after* flipping `current`, so it also skips the health gate that follows.

## Fix

`sudo readlink` for cwd, and `sudo cat ... | tr` for cmdline. Both verified against the live host before writing:

```
cwd:     [/opt/.../releases/20260903T141455Z-canonical-b45fb3e4]
cmdline: [/opt/.../venv/bin/python3 /opt/.../current/scripts/eeebot_dashboard.py --serve --port 8080 --host 0.0.0.0]
```

[...]

PR #1266, opened 2026-09-03
fix(deploy): say "unreadable" instead of an empty value, and let CRITICAL roll back (#1259)

### LINKED ISSUE #1259: deploy_release.sh cannot say 'unreadable': every CRITICAL-feeding read collapses a failed read into an empty value
## Problem

Every host read in `deploy_release.sh` that feeds a `CRITICAL` collapses an unreadable result into an empty one, and the gate then aborts with a message that describes the emptiness rather than the failure. From #1173's amendment, verified:

```
377  DASHBOARD_CWD="$(sudo readlink "/proc/$DASHBOARD_PID/cwd" 2>/dev/null || true)"
381  DASHBOARD_CMDLINE="$(sudo cat "/proc/$DASHBOARD_PID/cmdline" 2>/dev/null | tr "\0" " " || true)"
414  DASHBOARD_SOCKET_ROWS="$(sudo ss -ltnpH 2>/dev/null | awk ...)"
536+ the health gate's read of state/bridge/exit_streak.json over ssh
```

`2>/dev/null || true` discards both the error text and the exit status. So `cwd is ''` means *either* "the process is running from the wrong release" *or* "the read failed", and nothing distinguishes them.

That ambiguity is not hypothetical: **it cost four consecutive broken deploys on 2026-09-03** (#1246). The first was a missing `sudo`; #1247 restored readability and left the shape, so the next read that fails for any other reason — process exits between `systemctl show` and the read, permission changes, `/proc` unavailable — produces the identical message and the identical dead end.

The `exit_streak.json` read is the same shape with a worse consequence: an unreadable file is a *missing success signal*, so the gate degrades to a slow UNKNOWN rather than naming a failure.

## Why this is the deploy script specifically

This is the fail-open-reader class of #1173, and the ADR's proposed mechanism does not reach it — D-1 is a Python module and D-3's hygiene test walks `*.py` under `nanobot/`. `deploy_release.sh` is bash and outside both. It is also the one place where the reader's own failure blocks the recovery path: every abort today happened **after** `current` was flipped, so the health gate never ran, and the rollback trap did not fire because `set -E` is absent and the `exit 1` sits inside a function.

## What to do

For each read that feeds a `CRITICAL`:

- test the command's exit status separately from its output;
- on failure, emit `unreadable: <what was read> (<exit status / stderr tail>)` and abort with *that*, not with the empty-value message;
- keep the fail direction explicit in a comment next to the check — every one of these aborts, which is correct here, but it should be a stated choice rather than a side effect of `|| true`.

Also in scope, small and related: `set -E` / `set -o errtrace` for the remote block so the rollback trap covers shell functions, or a stated reason it cannot be. Four half-activated releases survived today by luck rather than design.

## Non-goals

- Do not change what any gate decides. This changes what it *says* when it cannot read, not when it aborts.
- Do not widen to `host/eeepc/scripts/*` generally. The reads that feed a `CRITICAL` are the set; the rest are not this issue.

## Acceptance Criteria

- [ ] Every `CRITICAL`-feeding read distinguishes unreadable from empty, and the message names which.
- [ ] Reproduced: make one read fail for a reason other than the value being wrong (e.g. an unreadable `/proc` path) and show the gate reporting `unreadable` rather than the empty-value text.
- [ ] The four #1246 defects still reproduce through `--verify-only` where they did before — this must not weaken a check to make it clearer.
- [ ] `set -E` set for the remote block, or the reason it cannot be stated with evidence.
- [ ] Full pytest; baseline is 4 failed, by test id.

ozand:
Closes #1259. From #1173's amendment, item 1.

## Defect

Every host read feeding a `CRITICAL` used `cmd 2>/dev/null || true`, discarding the exit status and the error text. `cwd is ''` meant either "wrong release" or "the read failed"; four deploys died on the second on 2026-09-03 while the message said the first (#1246). The health gate's `exit_streak.json` read had the same shape with a worse consequence: an unreadable file was a silently missing success signal.

## Change — what is said, not what is decided

**`gate_read <what> <cmd…>`** wraps every gate read in the remote block: the `current` symlink (verify-only), `/proc/<pid>/cwd`, `/proc/<pid>/cmdline`, `ss -ltnpH`, `ss -ltnH`. A non-zero exit → `CRITICAL: unreadable: <what> (exit N: <stderr tail>)` and abort. An **empty successful read is passed through unchanged** — the consuming gate still decides what empty means, so no gate's verdict moved. It streams through a file rather than a variable because a bash command substitution drops NUL bytes (below).

**`die <msg>`** replaces every `echo "CRITICAL: …" >&2; exit 1` in the remote block. **`set -E` was already there and covered no CRITICAL**: the ERR trap fires on a failing *command*, and `exit 1` is not one — verified empirically:

```
set -eEuo pipefail; trap 'echo TRAP' ERR
f(){ exit 1; };   f      → no TRAP, rc=1      (every CRITICAL until now)
g(){ return 1; }; g      → TRAP, rc=1         (die)
V="$(h)" with h failing  → TRAP               (gate_read assignment)
r | tr with r failing    → TRAP, rc=3         (gate_read in a pipeline, pipefail)
```

So every half-activated release on 2026-09-03 stayed active because the abort path never reached `rollback_remote`. `die` returns 1 with the same text; the trap runs, then `set -e` aborts. `rollback_remote` now acts once and only in the parent shell — a failing `gate_read` raises ERR twice, once inside the `$(…)` subshell and once for the assignment. `sync_timer` already used `return 1` and was covered.

**Health gate**: the `exit_streak.json` read keeps its decision (a missing file is still no signal; the journal verdicts still decide) but an unreadable file is logged **once** as `Health gate: unreadable: <path> (exit N: <stderr>) — the exit recorder's CLEAN-EXIT/FAIL signals cannot arrive; deciding on the journal alone.`

`readlink` gained `-v` so a failure carries its reason on stderr.

## Unreadable vs empty — demonstrated on `eeepc-lan` with `--verify-only`

Scratch copies mutated one line each; nothing deployed, no unit touched. Live release `20260903T220549Z-canonical-6b1d070e`, dashboard PID 5974.

| run | mutation | before this PR | now |
|---|---|---|---|
| unmutated | — | green | green (`=== Deploy complete ===`, exit 0) |
| M0 | `DASHBOARD_PID=999999` (read fails; the value is not wrong) | `cwd is '', expected '/opt/…'` | `CRITICAL: unreadable: /proc/999999/cwd (exit 1: readlink: /proc/999999/cwd: No such file or directory)` |
| M1 | `ss` binary missing | `saw listeners=0 pids=''` | `CRITICAL: unreadable: ss -ltnpH (exit 1: sudo: ss-nonexistent: command not found)` |

## Four-defect reproduction, re-run

| defect | caught | message now |
|---|---|---|
| D1 `sudo` removed from the cwd `readlink` | **yes** | `CRITICAL: unreadable: /proc/5974/cwd (exit 1: readlink: /proc/5974/cwd: Permission denied)` — **message changed**: it was `cwd is ''`, which is now reserved for a process genuinely running elsewhere |
| D2 cmdline compared to `$RELEASE_DIR` | **yes** | `CRITICAL: … has unexpected command line: /opt/…/venv/bin/python3 /opt/…/current/scripts/eeebot_dashboard.py --serve --port 8080 --host 0.0.0.0` — unchanged |
| D3 `sudo` removed from `ss -ltnpH` | **yes** | `CRITICAL: :8080 must have exactly one owner, dashboard PID 5974; saw listeners=1 pids=''` — unchanged, and correctly **not** `unreadable`: `ss` succeeded, the value is what is wrong |
| D4 socket sampled once | **no, by design** | verify-only does not restart the service, so the bind race cannot occur — unchanged |

Detection is unchanged; the text under D1 is the whole point of the issue.

## The regression `--verify-only` caught in this PR

The first draft captured the sudo'd `cat` into a variable inside `gate_read` and applied `tr "\0" " "` afterwards. The healthy host failed: `unexpected command line: …python3/opt/…/eeebot_dashboard.py--serve--port8080--host0.0.0.0` — bash command substitution drops NUL bytes, so `tr` had nothing to convert. The sandbox test could not see it (its fixture ends in a single NUL). Fixed by streaming through a temp file; pinned by `test_die_and_gate_read_semantics…` case 2b (`printf 'python3\0--serve\0' | tr` → `python3 --serve `). Two `--verify-only` runs, one minute, one regression that would have been a broken deploy.

## Not changed, stated

- `systemctl show … 2>/dev/null || true` for `LoadState`/`is-enabled` (lines ~430-433) feed a *skip*, not a CRITICAL: a failed `systemctl` there makes the whole dashboard block silently skip. Same class, different decision; out of this issue's stated set. Open question below.
- `sync_timer`'s `is-enabled … || true`: `is-enabled` exits 1 for a legitimately disabled unit, so `|| true` is needed for the value there; a failed `systemctl` reads as `''` → "state is , failing verify-only". Not one of the four reads; noted.
- `curl … || true` for `/api/health`: a fetch, whose failure is the finding; message unchanged.

## Tests

`tests/test_deploy_release.py` + `tests/test_deploy_integrity.py`: 35 passed (was 32; +3). New: `test_gate_reads_distinguish_unreadable_from_empty` (content: the five reads go through `gate_read`, none keeps `2>/dev/null || true`, the health gate read says unreadable), `test_every_remote_critical_goes_through_die_so_the_err_trap_fires` (no `exit 1` in the remote block; `set -eEuo`, trap, `die` present), `test_die_and_gate_read_semantics_under_the_remote_shell_options` (runs the real helper text under `set -eEuo pipefail` + ERR trap: failed read → unreadable + trap + abort; empty read passes through; NULs survive; `die` fires the trap; `exit 1` does not). Six existing literal assertions updated to the new shape (`readlink -v`, `die "…"`, `gate_read …`); the messages they pin are unchanged.

Full suite result posted as a comment (baseline 4 by id: bridge tag fail-open + three `TestAcquireBridgeLock`).

### DIFF
--- a/host/eeepc/scripts/deploy_release.sh
+++ b/host/eeepc/scripts/deploy_release.sh
@@ -128,6 +128,45 @@ run ssh "ozand@${HOST}" "REMOTE_ARCHIVE='${REMOTE_ARCHIVE:-}' RELEASE_NAME='${RE
 # Make ERR trap inherit to shell functions (the rollback trap)
 set -eEuo pipefail
 
+# `set -E` alone does not make the rollback trap cover a CRITICAL: the ERR trap
+# fires on a command that FAILS, and `exit 1` is not a failing command, so
+# every `echo CRITICAL; exit 1` in this block left `current` flipped with no
+# rollback (four half-activated releases on 2026-09-03, #1246). `die` returns
+# non-zero instead: under `set -eE` that runs the ERR trap (rollback_remote,
+# once installed) and then aborts the block with status 1. Fail direction of
+# every CRITICAL below: abort, and roll back if `current` was flipped.
+die() {
+  echo "CRITICAL: $*" >&2
+  return 1
+}
+
+# Every read that feeds a gate goes through this (#1259). `cmd 2>/dev/null ||
+# true` collapsed "the read failed" into "the value is empty" and the gate then
+# described the emptiness: `cwd is ''` meant EITHER wrong release OR a failed
+# read, and the four 2026-09-03 deploys died on the second while the message
+# said the first. gate_read prints the command's stdout unchanged (an EMPTY
+# successful read is still returned — the consuming gate decides what empty
+# means); a non-zero exit is reported as `unreadable: <what> (exit N: <stderr
+# tail>)` and propagates through the assignment, so `set -e` aborts and the ERR
+# trap rolls back. Fail direction: a gate that cannot read must not guess.
+gate_read() {
+  local what="$1"; shift
+  local out err rc=0
+  out="$(mktemp)"; err="$(mktemp)"
+  # Stream through a file, not a variable: a bash command substitution drops
+  # NUL bytes, and /proc/<pid>/cmdline is NUL-separated — capturing it here
+  # would hand the caller's `| tr "\0" " "` nothing to convert (seen on the
+  # healthy host as `python3/opt/…--serve--port8080`, via --verify-only).
+  if "$@" >"$out" 2>"$err"; then rc=0; else rc=$?; fi
+  if [ "$rc" -ne 0 ]; then
+    echo "CRITICAL: unreadable: $what (exit $rc: $(tr '\n' ' ' <"$err" | tail -c 200))" >&2
+    rm -f "$out" "$err"
+    return "$rc"
+  fi
+  cat "$out"
+  rm -f "$out" "$err"
+}
+
 ARCHIVE="${REMOTE_ARCHIVE:-}"
 RELEASE_NAME="${RELEASE_NAME:-}"
 FULL_COMMIT="${FULL_COMMIT:-}"
@@ -140,14 +179,12 @@ VENV_BASE=/opt/eeepc-agent/runtimes/self-evolving-agent/venv
 
 if [ "$VERIFY_ONLY" -eq 1 ]; then
   echo "[remote] VERIFY-ONLY mode: using current live release for checks"
-  RELEASE_DIR="$(sudo readlink "$CURRENT_SYMLINK" || true)"
+  RELEASE_DIR="$(gate_read "$CURRENT_SYMLINK" sudo readlink -v "$CURRENT_SYMLINK")"
   if [ -z "$RELEASE_DIR" ]; then
-    echo "CRITICAL: could not resolve current release symlink in verify-only mode" >&2
-    exit 1
+    die "current release symlink $CURRENT_SYMLINK resolved to an empty target in verify-only mode"
   fi
   if [ ! -f "$RELEASE_DIR/SOURCE_COMMIT" ]; then
-    echo "CRITICAL: no SOURCE_COMMIT found in current release $RELEASE_DIR" >&2
-    exit 1
+    die "no SOURCE_COMMIT found in current release $RELEASE_DIR"
   fi
   FULL_COMMIT="$(sudo cat "$RELEASE_DIR/SOURCE_COMMIT")"
 else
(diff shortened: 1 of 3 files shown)


ozand:
Full suite in the worktree at `f3c28e51` (`python -m pytest -q -n 4`): **4 failed, 2995 passed, 17 skipped** in 11m44s. Failure list, by id — exactly the Windows baseline:

```
```

(A first attempt at this run exited 0 having written nothing through `| tail`; rerun with the raw log kept. Two deploy modules: 35 passed, matching the operator's independent run.) CI green 3.11/3.12/3.13.

PR #1271, opened 2026-09-04
fix(deploy): gate the dashboard HTML page, not only its JSON endpoints (#1270)

### LINKED ISSUE #1270: Deploy health gate never fetches the dashboard's HTML page: a KeyError on / shipped and stayed live behind a green gate
## What happened

On 2026-09-04 the dashboard's HTML page raised `KeyError: 'overall_health'` on every request, in production, and **the deploy that shipped it passed its health gate.**

```
Sep 04 02:45:07 eeepc python3[24894]: Traceback (most recent call last):
Sep 04 02:45:07 eeepc python3[24894]: KeyError: 'overall_health'
Sep 04 02:45:08 …  Sep 04 02:51:27 …  Sep 04 02:52:03 …   (eeebot-dashboard.service)
```

Timeline, MSK:

| time | event |
|---|---|
| 02:18 | #1265 merged |
| 02:30 | release `20260903T233012Z-canonical-eed49ef` deployed, **gate green** |
| 02:45 – 02:52 | `/` returns 500 on every request; `KeyError: 'overall_health'` |
| 03:03 | #1269 merged (`prevent HTML health context KeyError`) |
| 03:10 | release `20260904T001026Z-canonical-487fc02` deployed; `/` back to 200, 18803 B |

## Why the gate missed it

`host/eeepc/scripts/deploy_release.sh` fetches exactly two URLs:

```
529:  DASHBOARD_HEALTH="$(curl --fail --silent --show-error http://127.0.0.1:8080/api/health  2>/dev/null || true)"
530:  DASHBOARD_METRICS="$(curl --fail --silent --show-error http://127.0.0.1:8080/api/metrics 2>/dev/null || true)"
```

`grep -oE "http://127.0.0.1:8080[a-z/_]*"` over the whole script returns those two and nothing else. **The HTML page at `/` is never requested by any gate.**

Both JSON endpoints were healthy throughout — they are built from a different code path than the HTML renderer, which is precisely the path #1265 changed. So the gate was not weakened or bypassed; it was measuring two surfaces while a third, the one a person actually looks at, was down.

This is the same class as #1246 and #1259: a check that cannot observe the failure it exists to prevent. The dashboard exists to be read by a human at `/`; a gate that validates only its machine endpoints certifies the wrong thing.

## Scope

Add `/` to the dashboard gate. It is the cheapest possible addition — one more `curl` beside two that already run, against a service the gate has just restarted and is already waiting on.

Deliberately **not** in scope: validating the HTML content beyond "it rendered". A status-code and non-trivial-size check catches this entire class (an unhandled exception yields 500, a broken renderer yields a stub) without coupling the deploy gate to markup.

## Acceptance criteria

- The deploy gate requests `http://127.0.0.1:8080/` after the dashboard restart, in normal deploy mode and in `--verify-only`.
- A non-200 response, or a body below a small floor (the live page is ~18.8 KB; a floor near 1 KB is generous), is a `CRITICAL` that aborts — and therefore, since #1259, rolls back.
- The check distinguishes **unreadable from failing**, per #1259: a `curl` that cannot connect reports `unreadable`, separately from a page that answers with an error status. It must not collapse into an empty value.
- A `--verify-only` run against the healthy host passes.
- The check is demonstrably failable: show it firing against a deliberately broken render (for example a fixture that raises in the HTML path), with the message it emits. A check that has never produced its own failure output is not a check.
- Deploy tests cover both directions — healthy `/` passes, 500 and short-body fail — and the existing four-defect reproduction still catches D1, D2 and D3.

## Note on how this one reached production

#1265 was merged while an open review asked for rework, and deployed against an instruction not to deploy. #1269's own body says "Do not merge/deploy until independent review and exact CI pass" and it was merged three minutes after opening. Those are process failures and are being handled separately — **but they are also the argument for this issue.** Review is fallible and will be bypassed again; the gate is the layer that does not depend on anyone reading a comment.

ozand:
Closes #1270. **Do not merge on my say-so** — opened for review; the operator merges.

## Defect

The dashboard gate fetched exactly two URLs, `/api/health` and `/api/metrics`. On 2026-09-04 the HTML renderer raised `KeyError: 'overall_health'` on every request for ~25 minutes while both JSON endpoints stayed healthy (different code path), and the deploy that shipped it passed its gate. The page a person reads was never requested.

## Change

One request added beside the two that already run, after the socket-owner and JSON checks:

```bash
DASHBOARD_PAGE_BODY="$(mktemp)"
DASHBOARD_PAGE_STATUS="$(gate_read "http://127.0.0.1:8080/" curl --silent --show-error --max-time 30 --output "$DASHBOARD_PAGE_BODY" --write-out '%{http_code}' http://127.0.0.1:8080/)"
DASHBOARD_PAGE_BYTES="$(wc -c <"$DASHBOARD_PAGE_BODY" | tr -d ' ')"
[ "$DASHBOARD_PAGE_STATUS" != "200" ]     → die "dashboard page / returned HTTP <code> (<n> bytes); the HTML renderer is broken while /api/* may still be healthy"
[ "$DASHBOARD_PAGE_BYTES" -lt 1024 ]      → die "dashboard page / body is <n> bytes (< 1024); the renderer produced a stub"
echo "[remote] dashboard page / HTTP 200, <n> bytes"
```

Status code and a body-size floor only. **No markup validation.** Runs in normal deploy mode and `--verify-only` alike (same block). #1259's distinction is kept: `gate_read` reports a curl that cannot connect as `unreadable: http://127.0.0.1:8080/ (exit 7: …)`; a page that answers with an error is reported by its status. Deliberately no `--fail`, so a 500 reaches the status check instead of the unreadable branch. Since #1266, every one of these `die`s rolls back in deploy mode.

## Demonstration — `--verify-only` against `eeepc-lan`, nothing deployed, no unit touched

**Healthy host, this branch:**
```
[remote] dashboard page / HTTP 200, 18804 bytes
[deploy] === Deploy complete ===          exit 0
```

**Firing**, scratch copies with only the page URL pointed at `:18080`, served by an ephemeral user-level `python3 -m http.server` in `/tmp` (killed and its dirs removed afterwards):

| case | message |
|---|---|
| F1 nothing listening on `:18080` | `CRITICAL: unreadable: http://127.0.0.1:18080/ (exit 7: curl: (7) Failed to connect to 127.0.0.1 port 18080 after 0 ms: Couldn't connect to server)` |
| F2 `200` with an 18-byte body | `CRITICAL: dashboard page / body is 18 bytes (< 1024); the renderer produced a stub` |
| F3 `404` (335-byte error page) | `CRITICAL: dashboard page / returned HTTP 404 (335 bytes); the HTML renderer is broken while /api/* may still be healthy` |

Each aborted with `[remote] check failed in verify-only mode; leaving live release alone`. The production failure (500 with a traceback page) takes F3's branch.

## Four-defect reproduction, re-run on this branch

| defect | caught | message |
|---|---|---|
| D1 `sudo` removed from cwd `readlink` | **yes** | `unreadable: /proc/27790/cwd (exit 1: readlink: … Permission denied)` |
| D2 cmdline compared to `$RELEASE_DIR` | **yes** | `… has unexpected command line: …/venv/bin/python3 …/current/scripts/eeebot_dashboard.py --serve …` |
| D3 `sudo` removed from `ss -ltnpH` | **yes** | `:8080 must have exactly one owner, dashboard PID 27790; saw listeners=1 pids=''` |
| D4 socket sampled once | no, by design | verify-only does not restart the service |

## False positive: what it would look like, and why the floor will not produce one

A false positive is a healthy release rolled back because `/` answered non-200 or under 1 KB. Candidates:

- **Slow first render after restart.** Measured live, five sequential GETs of `/`: `1.146 s` (cold), then `0.005 s` ×4, all `200 18804B`. The check runs after the gate has already polled the socket for up to 10 s and fetched both JSON endpoints, and carries `--max-time 30`. A first render 25× slower than the cold one measured would still pass.
- **Body under the floor.** The live page is 18,804 B; the floor is 1,024 B, an 18× margin. The dashboard has no empty-state render that small — the sandbox fixture's own page mock returns 2 KB. A working renderer that produces < 1 KB of HTML would be a product change large enough to notice.
- **Transient 5xx during startup.** The socket is bound and `/api/*` answered by this point, so the process is serving; the renderer either works or does not for a given release. The 25-minute outage was deterministic, not flapping.
- **curl missing / connect refused** → `unreadable`, abort. Correct: the gate cannot certify what it cannot read.

The remaining risk is a genuinely broken renderer being rolled back — which is the intended behaviour, and what did not happen on 2026-09-04.

## Tests

`tests/test_deploy_release.py` + `tests/test_deploy_integrity.py`: **36 passed** (+1). New `test_dashboard_page_gate_fires_on_error_status_stub_and_unreachable` runs the real page-gate text under `set -eEuo pipefail` + ERR trap with a fake `curl`: 500 → status CRITICAL + trap; 200/18 B → floor; connect failure → `unreadable … exit 7` and *not* a status message; 200/18804 B → passes and echoes. Content test pins the exact `gate_read … curl` line, both `die` messages, the floor, and that the URL set the script fetches is exactly `{/, /api/health, /api/metrics}`. The sandbox end-to-end test's curl mock now answers `--write-out` with `200` and a 2 KB body.

Full suite result posted as a comment when the run finishes (baseline 4 by id: bridge tag fail-open + three `TestAcquireBridgeLock`).

### DIFF
--- a/host/eeepc/scripts/deploy_release.sh
+++ b/host/eeepc/scripts/deploy_release.sh
@@ -534,6 +534,29 @@ if [ "$DASHBOARD_LOAD_STATE" = "loaded" ]; then
       fi
       die "could not fetch health or metrics from :8080"
     fi
+    # #1270: the page a person actually reads. /api/health and /api/metrics
+    # are built by a different code path from the HTML renderer, so a
+    # KeyError in the renderer shipped behind a green gate and / returned
+    # 500 for 25 minutes. Status code and a body-size floor only — never
+    # markup: an unhandled exception yields a non-200, a broken renderer
+    # yields a stub, and coupling the gate to HTML content is how a gate
+    # gets routed around. gate_read keeps #1259's distinction: curl that
+    # cannot connect is `unreadable: …/ (exit 7: …)`, a page that answers
+    # with an error is reported as its status code. No --fail here: a 500
+    # must reach the status check, not the unreadable branch.
+    DASHBOARD_PAGE_BODY="$(mktemp)"
+    DASHBOARD_PAGE_STATUS="$(gate_read "http://127.0.0.1:8080/" curl --silent --show-error --max-time 30 --output "$DASHBOARD_PAGE_BODY" --write-out '%{http_code}' http://127.0.0.1:8080/)"
+    DASHBOARD_PAGE_BYTES="$(wc -c <"$DASHBOARD_PAGE_BODY" | tr -d ' ')"
+    rm -f "$DASHBOARD_PAGE_BODY"
+    if [ "$DASHBOARD_PAGE_STATUS" != "200" ]; then
+      die "dashboard page / returned HTTP $DASHBOARD_PAGE_STATUS ($DASHBOARD_PAGE_BYTES bytes); the HTML renderer is broken while /api/* may still be healthy"
+    fi
+    # Live page is ~18.8 KB; the floor catches a stub or an error page that
+    # somehow carries a 200, and nothing a working renderer produces.
+    if [ "$DASHBOARD_PAGE_BYTES" -lt 1024 ]; then
+      die "dashboard page / body is $DASHBOARD_PAGE_BYTES bytes (< 1024); the renderer produced a stub"
+    fi
+    echo "[remote] dashboard page / HTTP $DASHBOARD_PAGE_STATUS, $DASHBOARD_PAGE_BYTES bytes"
     DASHBOARD_HEALTH="$DASHBOARD_HEALTH" DASHBOARD_METRICS="$DASHBOARD_METRICS" python3 - <<'PY'
 import json
 import os
--- a/tests/test_deploy_release.py
+++ b/tests/test_deploy_release.py
@@ -1,4 +1,5 @@
 import os
+import re
 import shlex
 import shutil
 import subprocess
@@ -199,6 +200,16 @@ def test_dashboard_activation_fails_when_enabled_unit_restart_fails(repo, mock_b
     assert 'curl --fail --silent --show-error http://127.0.0.1:8080/api/health' in content
     assert 'curl --fail --silent --show-error http://127.0.0.1:8080/api/metrics' in content
     assert 'dashboard listener :8080 is not active' in content
+    # #1270: the page a person reads. A KeyError in the HTML renderer shipped
+    # behind a green gate because only the two /api/* endpoints were fetched.
+    # Status + body-size floor, through gate_read (connect failure stays
+    # `unreadable`), and deliberately no --fail so a 500 reaches the status check.
+    assert 'gate_read "http://127.0.0.1:8080/" curl --silent --show-error --max-time 30 --output "$DASHBOARD_PAGE_BODY" --write-out \'%{http_code}\' http://127.0.0.1:8080/' in content
+    assert 'die "dashboard page / returned HTTP $DASHBOARD_PAGE_STATUS' in content
+    assert 'if [ "$DASHBOARD_PAGE_BYTES" -lt 1024 ]; then' in content
+    assert 'die "dashboard page / body is $DASHBOARD_PAGE_BYTES bytes (< 1024)' in content
+    urls = sorted(set(re.findall(r"http://127\.0\.0\.1:8080[a-z/_]*", content)))
+    assert urls == ["http://127.0.0.1:8080/", "http://127.0.0.1:8080/api/health", "http://127.0.0.1:8080/api/metrics"], urls
 
 
 def test_dashboard_activation_verifies_pid_release_identity(repo, mock_bin):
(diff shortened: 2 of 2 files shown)


ozand:
Full suite in the worktree at `fea0efd2` (`python -m pytest -q -n 4`, run alone, log kept): **4 failed, 2999 passed, 17 skipped** in 5m58s. Failure list by id — exactly the Windows baseline:

```
```

For the record: a first full run that overlapped with the deploy-module run showed a fifth failure, `test_usage_evidence.py::TestOwnerLiveRatioAndSurfaces::test_refresh_usage_replaces_unwritable_target_file_in_writable_dir` (`_git_head(repo)` returned `None` — a `git rev-parse` subprocess failing under load). It passes alone in this worktree and on `main`'s tip, and is absent from the solo run; not touched by this diff.

ozand:
Least-certain claim resolved by the operator, in the floor's favour: `scripts/eeebot_dashboard.py` has one HTML-returning function, `render_html`, whose fixed scaffold from `<!DOCTYPE html>` to `</style>` is 3,318 characters of near-literal CSS. Any successful render exceeds the 1,024-byte floor ~3× regardless of data state, so the floor cannot produce a false rollback. Merged as `76bab401`; live `--verify-only` prints `[remote] dashboard page / HTTP 200, 18804 bytes`.

Issue #877, opened 2026-08-16
RSI stage 3: git-native evolutionary tree — population as branches, sequential evaluation


ozand:
Part of #874 (stage 3 of 4). Operator framing (verbatim intent): memory is files and git — no need to hold a population in RAM. Create trees natively in git, move along branches, roll back several steps when needed, and start new cycles from that point; return to a branch if its line proved stronger.

## Design intent — DGM archive semantics on the git DAG
- **Individual = branch, generation = commit, fitness = ledger entry per node.** The loop already retains cycle branches (#830, capped) and a small diversity archive (#844) — unify them: an evolution-tree index (harness-owned sidecar) mapping node (branch@sha) → fitness snapshot (scorecard deltas, microbench results, confirmed usage).
- **Selection**: when the current line stagnates (existing stalled()/archive signals), the loop checks out the best OTHER node by fitness and continues from there — a native git checkout, no new machinery. Parent selection can be DGM-style (fitness-proportional with novelty bonus from #844) — keep it a simple scored pick, no MAP-Elites grid (2GB discipline).
- **Sequential evaluation**: one live process; the tree grows one node at a time. Parallelism in STRUCTURE (many dormant branches), not in execution — evaluation throughput is the only real constraint on this host.
- **Rollback = checkout**: any regression recovers by moving to an ancestor node; nothing is lost because losers stay as branches (subject to the existing #830 retention cap — raise cap only for nodes referenced by the tree index).
- Branch hygiene: tree-indexed branches exempt from pruning; everything else keeps current caps.

## Acceptance
- [ ] evolution-tree index sidecar (fitness per node) + ledger events for branch/checkout/abandon
- [ ] stagnation triggers a checkout to a better node and the loop continues from it (observed live)
- [ ] rollback drill: regression → ancestor checkout → line continues
- [ ] pruning respects tree-indexed branches
- [ ] tests + CI + deploy

ozand:
## Delivered + verified LIVE ✅ (board Done deferred — GraphQL rate-limit)

PR #890 (squash 76a13e2, deployed `20260817T160457Z-canonical-76a13e2`). THREE Opus adversarial rounds; final verdict **MERGE**, both discovered REDs closed.

## Design (operator framing: population=branches, generation=commit, fitness=ledger node; move/rollback via native git)
[...]

Issue #844, opened 2026-08-15
runtime: small diversity archive (3-5 slots) instead of pure greedy single-lineage [P2]


ozand:
## Problem
The loop integrates a single greedy lineage into main. SOTA open-ended systems (DGM arXiv:2505.22954, ADAS, AlphaEvolve/OpenEvolve) keep an ARCHIVE of diverse validated variants as "stepping stones" and demonstrate this escapes local optima that trap greedy hill-climbing (QD stepping-stone result, GECCO 10.1145/3319619.3321897). Systems that skip it (SICA greedy walk, Gödel-agent) flag overfitting/path-dependency in their own papers.

## Proposed approach (cheap version — NOT full MAP-Elites)
Retain a small archive of the last N (3-5) gate-passing candidates keyed on a behavior signature (demand kind / target area), rather than only the champion. Allow the proposer to sample a stepping-stone from the archive occasionally. Store N candidate refs, not a grid. Fits 2GB.

## Sources
B#1, C#7.

## Acceptance
- [ ] bounded diversity archive (<=5 slots) keyed on behavior signature
- [ ] proposer can seed from an archived stepping-stone
- [ ] tests; deployed

ozand:
✅ **Delivered** — deployed `20260816T050126Z-canonical-5947944` (main `5947944`, PR #858).

Exploration-gap fix (complements #840/#845 reuse). The loop kept a single greedy lineage; now it retains a small diverse archive of validated variants as DGM-style stepping-stones.

- **archive.py**: BRIDGE-lane stepping-stone archive (co-located with the coordinator's `CycleArchive`, distinct lane/file). `record_stepping_stone` keeps the **newest gate-passing variant per behavior signature** (primary changed area), at most **5 distinct signatures** (evict oldest) — diverse areas, not 5 of one lineage, not a MAP-Elites grid. `state/steering/stepping_stones.json`.
- **bridge.py**: fail-open record right after a successful integrate.
- **llm_proposer.py**: `build_context` appends a PROTECTED (never-truncated) 'Stepping stones' section the model MAY extend to explore a different area; omitted when empty (byte-identical to before).

**Proposer-steering ONLY** — the sidecar is NOT in `FITNESS_SIDECARS`; nothing gates/scores on it. A stepping-stone is a prompt hint, still subject to the full gate.

**Live-verified on host (deployed code, deterministic):**
```
symbols present ✓ ; steering NOT in FITNESS_SIDECARS ✓
[...]

Issue #185, opened 2026-04-25
Remove shell=True from subagent materializer executor path


ozand:
## Анализ / gap
Review subagent for #180/#183 found a security blocker: `nanobot/runtime/subagent_materializer.py` invokes the local executor with `subprocess.run(..., shell=True)` using a free-form command from `executor_command` / `NANOBOT_SUBAGENT_EXECUTOR_COMMAND`. This turns the new self-development executor path into a command-injection surface.

## Required product change
- Replace shell-string execution with argv-based execution.
- Keep support for bounded test executors without allowing shell metacharacters to execute.
- Do not serialize executor command strings or secrets into result artifacts.
- Preserve completed/blocked materialization semantics.

## DoD
- Regression test proves `shell=False` for local executor execution.
- Regression test proves shell metacharacters in argv are passed as inert arguments, not interpreted by a shell.
- Failure payloads redact `sk-*` / `sk-secret` values and omit raw commands.
- Focused materializer tests pass.
- Full root tests pass.

## Verification commands
```bash
python3 -m pytest tests/test_runtime_coordinator.py::test_subagent_materializer_runs_executor_without_shell tests/test_runtime_coordinator.py::test_subagent_materializer_records_executor_failure_without_leaking_command_secrets -q
python3 -m pytest tests -q
```

ozand:
Rollout proof after PR #187 merge (`fe0d69e9 fix: reconnect subagent executor through public Pi Dev (#187)`).

Implemented:
- Materializer executor no longer uses `shell=True`; execution is argv-based with `shell=False`.
- Product Pi Dev route is canonical argv: `--mode json -p --no-session --no-tools --provider hermes_pi_qwen --model gpt-5.3-codex`.
- Public LiteLLM metadata is exposed without secrets: `https://litellm.ayga.tech:9443/v1`.
- Stale `coder-model` delegated executor probe metadata was removed.
- Safe profiles can now materialize completed subagent results instead of only blocked records.

Verification:
- Focused materializer tests: 4 passed.
- Dashboard suite: 64 passed.
- Root suite: 600 passed, 5 skipped.
- GitHub Actions on PR #187: Python 3.11 pass, 3.12 pass, 3.13 pass.
- Review subagent: APPROVED.
- Non-secret Pi config proof: provider `hermes_pi_qwen` exists, baseUrl is `https://litellm.ayga.tech:9443/v1`, model list includes `gpt-5.3-codex`, auth ref exists; no raw key values printed or committed.

[...]

PR #1278, opened 2026-09-04
refactor(runtime): retire the #877 line switch and its stall trigger (#1225)

### LINKED ISSUE #1225: goals/cycle_archive.json has no writer since 2026-08-22: #877 line switch is a constant no-op — restore a reward writer or retire the trigger (from #1222 row 4b)
## Found while resolving #1222 (row 4b)

`bridge.py:653` loads `state/goals/cycle_archive.json` into `CycleArchive`, asks `stalled()` (last 5 cycles all `reward < 0.8`, #844) and, if true, `evolution_tree.should_switch` (#877) switches the working line. That is the **only** trigger for a line switch.

Facts (read-only, eeepc, 2026-09-03):

- `cycle_archive.json`: mtime **2026-08-22 02:00**, 200 entries, newest `cycle-2f305bf18b42`. Every entry has `reward: 1.0`.
- The only code that ever saved it was the coordinator (`CycleArchive.save` has no caller in `nanobot/` since #916/#923). `bridge` calls `archive.record_stepping_stone`, which writes `steering/stepping_stones.json` — a different file, a different duty.
- Consequence: `stalled()` is a constant `False`. The #877 line switch has never fired and cannot fire. It was already a constant before the freeze (200 × 1.0).

This is the **orphaned** class: the feature still exists, its input has no owner. Two honest fixes, one decision:

1. **Restore the writer** — define what a cycle's reward is now (scorecard? integrate outcome? the `fitness` the evolution tree already records with `reward: None` "to be backfilled") and have the bridge append one entry per cycle at the integrate/outcome site. Design work: the reward semantics are the whole question.
2. **Retire the trigger** — delete the `archive.load`/`stalled()`/`should_switch` block in `_setup_cycle_branch`, the `CycleArchive` persistence and `test_cycle_archive.py`'s stall tests; keep `evolution_tree` as a record only.

Not decidable from the code. Until decided, `goals` carries `orphan:#<this issue>` in `STATE_PATH_WRITERS` for this file only (`registry.json`/`active.json`/`current.json` are resolved in #1222's PR).

ozand:
Closes #1225. Operator decision: **retire**, not restore. Companion instance-side change: ozand/eeebot-self-evolving#185.

## Why retire is right (the two facts, so nobody reopens this)

1. **`tree.json` records 0 switches ever, and no `evo/node-*` keeper branch was ever created** (the branch the switch would have left behind at the abandoned tip). The lever has no production history at all.
2. **Every one of the 200 `cycle_archive.json` entries carries `reward: 1.0` against a `< 0.8` stall rule**, and the file's only writer was the coordinator deleted in #916/#923 — frozen 2026-08-21T23:00:58Z. `stalled()` was a constant `False` before the freeze and after it.

A lever that never worked, not one that stopped working. (The skipped-duplicate mapping question from the restore arm is moot and gets no line here.)

## Dropped (responsibilities, not files — #1263 checklist)

| duty | verdict | change |
|---|---|---|
| stall detection over the coordinator archive (`CycleArchive.stalled()`, `ArchiveEntry`, `STALL_*`) | **dropped** | removed from `archive.py`; the module keeps only the #844 stepping-stone functions, still read by `bridge.py` and `llm_proposer.py` |
| the line-switch decision in `bridge._setup_cycle_branch` (load → `stalled()` → `should_switch` → keeper branch → `record_switch`) | **dropped** | block removed; the cycle always branches off the real `origin/main`; `main_sha == origin_main_sha` |
| `evolution_tree.should_switch` | **dropped** | removed |
| `tests/test_cycle_archive.py` (10 tests, all on `CycleArchive`) and the four trigger tests in `TestEvolutionTreeLineSwitch` | **dropped** | removed |
| `orphan:#1225` in `state_paths.STATE_PATH_WRITERS["goals"]` + its `ORPHAN_ISSUES` reason | **resolved** | removed; `ORPHAN_ISSUES` is empty with a comment on how to add one; the host file is an inert artifact |
| the frozen-list panel in the instance dashboard (`scripts/eeebot_dashboard.py:2174`, `format_cycle_archive_*`) | **orphaned reader → labelled** | ozand/eeebot-self-evolving#185: every render leads with `retired (eeebot#1225): … frozen 2026-08-21T23:00:58Z, no writer — history only`. Labelled, not removed: the list is real history; four render sites and the layout stay |
| `docs/changes/877-evolution-tree/proposal.md` | history | status note added at the top |

## Kept, deliberately

- `node_score`, `select_switch_target`, `record_switch`, `mark_switch_blocked` — the tree as an inspectable **record**. Nothing in the bridge calls them for a decision any more; `mark_switch_blocked` is still called by the integrate-side gate.
- The **integrate-side base-surface gate** (`_changed_files_and_violations` over `origin_main_observed..HEAD`) — defence in depth for any cycle whose base is not `origin/main`. Its four tests now build the switched base directly (`checkout -B <cycle> <dormant sha>`, as `TestIntegratePushAfterLineSwitch` always did) instead of through the removed trigger, so the gate's coverage is unchanged.

## Ranking cannot silently go down with the trigger (item 4)

On the host all 100 tree nodes have `fitness.reward` None and `confirmed_integrations` / `repeat_failure_rate` populated (scores 9.79–11.79). New `TestRankingWithoutRewardOrTrigger` writes three nodes in exactly that shape, asserts `node_score` = `0.1·confirmed − 0.2·repeat_fail`, the ordering, `select_switch_target` picking the best non-current node, and the next-best after `mark_switch_blocked`; plus `should_switch` absent.

New `TestLineSwitchRetired`: a 5-deep stalled archive file **and** a stronger dormant node no longer move the base (branch = `origin/main`, no `evo/node-*`, no `line_switch` ledger row, `switches == []`); and no `nanobot/runtime/*.py` carries a quoted `'cycle_archive.json'` literal (prose mentions in docstrings are history and allowed).

## Tests

Touched modules (`test_bridge_cycle_branch`, `test_evolution_tree`, `test_state_path_writers`, `test_stepping_stones`, `test_llm_proposer`): 326 passed. Full suite (`-n 4`, log kept): **4 failed, 2998 passed, 17 skipped** — exactly the baseline list: `test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle` and the three `test_bridge_locking.py::TestAcquireBridgeLock` cases. `ruff`: 0 new findings on every edited file (bridge.py's 23 are pre-existing on `main`).

Diff: 8 files, +157/−512 in code and tests, plus the doc note.

Merge and deploy are the operator's; nothing on the host was touched.


ozand:
## REVIEW VERDICT: REQUEST CHANGES

I cannot submit a formal GitHub review from the authenticated PR-author account, so this verdict is posted as a PR comment.

### 1. `CycleArchive.save` caller audit — verified

A repository-wide search found the only pre-PR production reference in `nanobot/runtime/bridge.py:651-653`, where the bridge constructed and loaded `CycleArchive`; it did not call `save`. The only `save()` call was test-only in `tests/test_bridge_cycle_branch.py:994`, and the deleted `tests/test_cycle_archive.py` exercised the class directly. After the bridge block is removed, no production caller or writer remains. This supports retiring `CycleArchive.save` and the class.

### 2. `orphan:#1225` cleanup — verified

The PR removes both required entries:

- `orphan:#1225` from `state_paths.STATE_PATH_WRITERS["goals"]`;
- `"#1225"` from `state_paths.ORPHAN_ISSUES`.

`tests/test_state_path_writers.py` retains the load-bearing bidirectional assertion that every orphan reason is referenced and every orphan reference has a reason. It does not weaken that assertion; an empty orphan table is valid because no orphan references remain.

### 3. Ranking without reward — verified, with one test-quality concern

`TestRankingWithoutRewardOrTrigger.test_select_switch_target_ranks_on_live_shaped_fitness_with_reward_none` exists and asserts the formula, ordering, target selection, and post-block next-best selection. It would fail if `node_score` still required a non-`None` reward: all three fixture rewards are explicitly `None`, so scores would collapse and the expected ordering/target would fail.

### 4. Dashboard frozen archive — not fixed by this PR, but governed by a separate change

The product PR does not modify the instance dashboard, which is correct for repository scope. The live `ozand/eeebot-self-evolving` repo has open Issue/PR #185:

https://github.com/ozand/eeebot-self-evolving/pull/185

That companion change explicitly labels `cycle_archive.json` as retired history in both text and HTML formatters, with tests for empty and non-empty cases. This is a gated separate change, not a direct commit in PR #1278. The repository-side responsibility is therefore addressed through the companion change, but it is not yet merged or deployed.

### 5. Deleted tests — one serious residual gap

The deleted 149-line `tests/test_cycle_archive.py` covered the retired class only: add/idempotence, ordering, best-N, stall threshold, insufficient history, save/load, corrupt/missing load, and max-entry pruning. Those behaviors have no production callers after the bridge removal, so deleting those tests is justified **if** the retirement boundary is complete.

However, the PR's replacement test `test_nothing_in_the_runtime_reads_cycle_archive_or_stalls` uses a regex for a quoted bare `cycle_archive.json` literal and scans only `nanobot/runtime/*.py`. It does not prove that all runtime readers/writers are gone when paths are assembled as `"cycle_archive" + ".json"`, `Path(...) / "cycle_archive.json"` with a differently formatted quote, or when another runtime subdirectory is involved. A repository-wide source search still finds live pre-PR references only because this review is against the base; the test itself is narrower than the retirement claim.

Please strengthen the retirement regression to scan all runtime Python files for the path in the actual construction forms (or assert the specific known owner modules and path components), and add/retain a check that no runtime code imports `CycleArchive`. Otherwise a future reader could be reintroduced while the test remains green.

### Scope conclusion

I found no evidence that the PR removes a live `CycleArchive` behavior: its production bridge consumer is the retired trigger itself, while the retained `archive.py` stepping-stone functions are a separate state file and remain wired. The requested change is limited to closing the retirement guard gap above; the dashboard responsibility is correctly tracked in companion PR #185 rather than smuggled into this PR.

ozand:
Acted on the review's one blocking finding directly, since the author is on #1280.

`0287e623` — the retirement guard now matches the retirement claim.

The gap was as reported: `runtime.glob("*.py")` is non-recursive and covers only `nanobot/runtime`, and the regex demanded the whole quoted filename. Two revival forms passed it.

Widening the regex alone does not work, which is worth recording. It reports `archive.py`, `bridge.py` and `evolution_tree.py` — all three are docstring prose describing this very retirement. The fix parses with `ast` instead: a docstring is one Constant holding a paragraph, never equal to the path, and prose naming `archive.CycleArchive.stalled()` is text rather than an Attribute node, so the two cases separate with no escaping.

Demonstrated rather than asserted — three probes against a temporary module, since a guard that has never been seen to fail is not a guard:

| probe | result | old guard |
|---|---|---|
| path assembled from the bare stem, `"cycle_archive"` then `.with_suffix(".json")` | fails | would have passed |
| class re-imported with no path literal, `archive.CycleArchive(state_dir)` | fails | would have passed |
| docstring prose about the retirement | passes | naively widened regex would have failed |

`tests/test_bridge_cycle_branch.py`: 54 passed.

The one blind spot left — a name built at runtime, `"cycle_" + "archive.json"` — is recorded in the comment rather than papered over. No static check closes it and the codebase does not use that form.

I wrote this, so I do not get to clear it: back to the reviewer to re-check before merge. The other four review sections were verified and need nothing.

ozand:
## RE-CHECK VERDICT: APPROVE

I independently re-checked commit `0287e623` against the guard gap previously reported.

### What the new guard now distinguishes

The test now recursively scans the whole `nanobot/` package (`rglob("*.py")`) and parses Python syntax with `ast` instead of searching for a whole quoted filename. Its checks distinguish:

- a `Constant` equal to `"cycle_archive"` or `"cycle_archive.json"` — caught;
- a `Name`, `Attribute`, or import alias referring to `CycleArchive` — caught;
- docstring/comment prose — not caught, because prose is a string constant containing a paragraph, not an exact path constant or an AST reference node.

I also reproduced the three relevant probe outcomes independently with the same AST rule: bare-stem/suffix construction is caught, class re-import without a path literal is caught, and retirement prose is not caught. The remaining blind spot is honestly recorded: runtime-constructed concatenation such as `"cycle_" + "archive.json"` is not detected by this static guard.

The guard is materially stronger and the remaining limitation is explicit rather than hidden. **Approve.**

ozand:
## RE-CHECK VERDICT: APPROVE

I independently re-derived the AST guard behavior on the checked-out `0287e623` worktree and ran `tests/test_bridge_cycle_branch.py`: **54 passed**.

### Verified properties

- `package = Path(bridge.__file__).parent.parent` resolves to the repository's `nanobot` package directory.
- `package.rglob("*.py")` is recursive over the whole package, unlike the previous non-recursive `runtime.glob("*.py")` scan.
- `path.relative_to(package.parent).as_posix()` produces repository-relative paths such as `nanobot/runtime/bridge.py`, matching the assertion's intended diagnostic format.
- A path assembled from the exact `"cycle_archive"` stem is caught by the `ast.Constant` check.
- A direct or aliased `CycleArchive` import/reference is caught by `ast.alias`, `ast.Name`, or `ast.Attribute` matching.
- Retirement prose is not caught: a docstring containing a paragraph mentioning `CycleArchive` or `cycle_archive.json` is not equal to either exact Constant and has no matching reference AST node.

### Attack on the matcher

I tested plausible revival forms. The explicitly recorded runtime string-concatenation blind spot remains real. A separate evasion is dynamic reflection, for example:

```python
mod = importlib.import_module("nanobot.runtime.archive")
cls = getattr(mod, "Cycle" + "Archive")
instance = cls()
```

The current guard does not catch that, but it is not a plausible existing codebase revival form: it requires dynamic import plus runtime class-name construction, and the retired code used a direct import. The guard's stated residual limitation is therefore accurate for the practical codebase, not a hidden ordinary import/path form.

No additional actionable gap found. **Approve.**

Issue #1281, opened 2026-09-04
A subagent that died on its LLM call can still integrate via the auto-commit safety net (cycle 3cbcc1f77d25): keep or gate?


ozand:
## Found while building #1280 — the `3cbcc1f77d25` anomaly

#1280 asked why cycle `3cbcc1f77d25` recorded `commits_pushed: 1` although "its subagent failed at 07:04:45". Identified from the ledger, the telemetry files and the instance repo (read-only):

| fact | value |
|---|---|
| cycle `3cbcc1f77d25` | `proposed` 02:43:43Z, `started` 02:47:09Z, `outcome: success` **03:20:40Z**, `files_changed: [scripts/scaffold_script_entrypoint.py]` |
| its subagent | `0652e0eb`, telemetry `status: error`, `finished_at` **03:20:13Z** = 06:20:13 MSK — the first outage failure in the issue's own timeline |
| the commit | `e7ecea84` 03:20:13Z `selfevo: auto-commit uncommitted subagent work — Update scaffold_script_entrypoint.py to generate positional args with nargs='?'`, merged as `e1ff2405` 03:20:36Z |
| `0b221168` (the 07:04:45 failure the issue paired with it) | belongs to cycle `5318708cc6ff` (04:01–04:04Z), not to `3cbcc1f77d25` |

So: the subagent had already edited `scripts/scaffold_script_entrypoint.py` via `edit_file` when its next LLM call died on the connection error. The bridge's **auto-commit safety net** (`_auto_commit_uncommitted_work`, #666, unconditional since #717) committed the uncommitted edit, the smoke gate passed, and the cycle integrated. Real work, not bookkeeping — and a cycle whose executor **failed** pushed to `origin/main`.

## Why it is its own question

#1280 (PR pending) records an LLM failure as a failed cycle **only when nothing was committed**; when auto-commit captured real work, the normal gate decides, exactly as it did here, and the error text is attached to the result's learnings. That preserves this outcome deliberately rather than silently reclassifying it.

Whether it *should* be preserved is a separate decision:

- **For**: the edit was a complete, gate-passing change; the failure came after the work. Discarding it because a later call died throws away a green integration.
- **Against**: the subagent never reached its own "done" verdict — no final answer, no self-review, no chance to revert a half-applied edit. `auto-commit` was built for "forgot to run git commit", not "died mid-turn". A half-finished multi-file edit whose remaining files were never written would pass the same path if the touched files happen to compile and the smoke set does not exercise them.

[...]

PR #1282, opened 2026-09-04
fix(bridge): record an executor LLM failure as a failed cycle, not completed (#1280)

### LINKED ISSUE #1280: Cycles whose executor LLM call fails are recorded as 'completed': a 4.5h model outage left consecutive_failures at 0
## Context (self-contained)

The local model behind the executor was unreachable on 2026-09-04 between roughly 06:20 and 11:00 MSK (the serving machine rebooted). The loop did not notice, and its terminal records say the opposite of what happened.

### What the journal shows

`nanobot.agent.subagent:_run_subagent:494` logged `Subagent [<id>] failed: LLM execution failed` twelve times today. Ten fall inside the outage, first at `06:20:13`, last at `10:55:27`, each with:

```
litellm.InternalServerError: OpenAIException - Connection error.
No fallback model group found for original model group
```

Two others are the separate, already-known Gemini 404 class (`litellm.NotFoundError`) at `05:38:25` and `11:26:54`.

Subagent-run failure rate by window (distinct subagent ids, from the journal):

| window | failed / runs | rate |
|---|---|---|
| 06:00–10:59 (outage) | 10 / 11 | **90.9%** |
| rest of the day | 2 / 26 | 7.7% |

### What the loop recorded

Each failed subagent maps to one cycle. Reading `state/subagents/**/result-*.json` for all twelve:

| cycle | result_status | commits |
|---|---|---|
| 23251dd39975 | blocked (`existence_index_duplicate` — unrelated cause) | 0 |
| 4fa0c42dff6f, 5318708cc6ff, 09b415162e4c, d8129ab18aea, 41e5cdb4979e, 7e075d9e5d54, 6c2add722000, 7c14447404ad, 7b22452c3f48, 1f1257e3d0f6 | **completed** | 0 |
| 3cbcc1f77d25 | **completed** | 1 |

Eleven of twelve cycles whose LLM call never returned are recorded as `completed`.

### Why that is worse than a wrong label

1. **The failure gate cannot see them.** `_recent_failure_match` (`bridge.py:4266`) counts a cycle as failed only when `rollback.reason` is set or `result_status` is in `{'blocked', 'no_commit'}`. `completed` is neither, so an outage cycle never enters the 24-hour suppression window and never informs the next proposal.
2. **The streak observer saw nothing.** `state/bridge/exit_streak.json` reads `consecutive_failures: 0`, `last_outcome: "success"`.
3. **The request is retired anyway.** On the post-spawn path, `handled_marker.write_text` at `bridge.py:2810` runs unconditionally — there is no branch on whether the subagent succeeded. Once the marker exists, `bridge.py:2116` skips the request forever. The eight proposals minted during the outage were spawned, failed, and permanently marked handled. None was retried when the model came back.
4. **Cadence stayed normal.** Cycles finished per hour ran 9–13 through the outage, indistinguishable from 11–14 in the healthy hours. The proposer kept minting: eight `improvements/llm-proposed-cycle-*.json` were written between 06:57 and 11:22. The loop kept paying for proposals it could not execute.

There is no fallback route configured — the gateway message says so in as many words.

### The anomaly not explained

`3cbcc1f77d25` records `commits_pushed: 1` although its subagent (`0b221168`) failed at `07:04:45`. That commit needs identifying before this is called fully understood: either the agent did real work before the call died, or bookkeeping was pushed under a failed cycle.

## Acceptance Criteria

- [ ] A cycle whose executor LLM call fails is not recorded as `completed`. Pick the status deliberately and say why in the code comment: it must be one the failure gate at `bridge.py:4266` recognises, or that gate must be widened in the same change — a new status that no reader treats as failure would repeat this bug in a new spelling.
- [ ] `exit_streak.json` reflects such a cycle as a failure, so a sustained outage moves `consecutive_failures`.
- [ ] The `handled_` marker is not written for a request whose subagent failed on transport, so the proposal is retried once the route recovers. Bound the retry (a count or an age) so a permanently-bad request cannot spin forever.
- [ ] A test drives the whole path with the LLM call raising the transport error above, and asserts all three: the recorded status, the streak, and the absence of the marker.
- [ ] Identify the commit behind `3cbcc1f77d25`'s `commits_pushed: 1` and record what it was. If it turns out that a failed cycle can push, that is a separate issue — file it rather than folding it in.
- [ ] Full pytest: Windows baseline is exactly 4 failures, by list — `test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle` plus the three `test_bridge_locking.py::TestAcquireBridgeLock` cases.
- [ ] Deployed and live-verified.

## Not in scope

Adding a fallback model group. That is a real gap and worth its own issue, but it is a configuration decision with a cost attached, and it does not fix the recording defect — a fallback that also fails would still be logged `completed`.

## Related

Same class as #1197 (three observers all reporting normal through a 9-hour crash loop). Contaminates #1211, which measures suppression as a share of cycles: outage cycles currently sit in the `completed` denominator.

ozand:
Closes #1280. Companion finding filed as #1281 (a failed subagent can still integrate via auto-commit).

## What was wrong

The subagent already wrote the verdict — telemetry `status: "error"`, `summary: "Error: LLM execution failed: … litellm.InternalServerError … Connection error"` — and the bridge never read it. `_bridge_status` started at `completed` and only smoke-block could change it; the ledger mapped zero commits to `partial`; the process exited 0 so `exit_streak.json` recorded a success; and `handled_marker.write_text` ran unconditionally right after the spawn, retiring the request. Twelve outage cycles on 2026-09-04, eleven `completed`.

## Change (bridge.py, +~170/−12)

**Read the verdict.** `_executor_llm_error(state_dir, task_id)` returns the failure text when the spawn's telemetry says `status: error` with `LLM execution failed`; `""` otherwise, fail-open.

**Status — an existing one, on purpose.** With zero commits: `rollback.reason = executor_llm_error`, `result_status = blocked`. `blocked` is one of the two statuses `_recent_failure_match` already treats as failure, and `rollback.reason` being set is the other test, so the row matches on both and enters the 24 h suppression window and the proposer's recent-failures context. The code comment says exactly this; a fresh status would have been invisible to every reader — the bug in a new spelling. `executor_llm_error` is added to `_INCONCLUSIVE_REASONS` (never upgraded to accept/reject). Ledger outcome `failed`, not `partial`. The provider error text is appended to `key_learnings` so a reader sees *why* nothing was produced.

**Streak.** `main()` returns `EXIT_EXECUTOR_LLM_ERROR = 3` for such a cycle. The existing `__main__` guard turns any non-zero code into a `failure` row in `bridge/exit_streak.json` — one writer, no double record — so a sustained outage moves `consecutive_failures`. Consequences, stated: systemd shows the unit failed for that run; the deploy health gate reads both the journal exit status and `exit_streak` and will roll back a release deployed *during* a model outage. Deliberate: a release whose executor cannot reach its model cannot be verified, and "cannot certify" must not read as green.

**Marker, bounded.** `_decide_handled_marker()` replaces the unconditional write. An LLM failure with nothing produced leaves the request pending and increments `retry_<id>.json`; on the `LLM_ERROR_MAX_RETRIES`th failure (3, `SUBAGENT_BRIDGE_LLM_ERROR_MAX_RETRIES`) the marker is written and the request retired like any other. Every other outcome writes the marker exactly as before. Fail-open to writing the marker.

**Real work survives.** If the subagent had edited files before its call died, the #666/#717 auto-commit captures them and the normal gate decides — the error text still lands in learnings. That is what happened to `3cbcc1f77d25`, and whether it *should* is #1281.

## The `3cbcc1f77d25` anomaly, identified

Its subagent was `0652e0eb`, not `0b221168` (that one belongs to `5318708cc6ff`, 04:01–04:04Z). `0652e0eb` edited `scripts/scaffold_script_entrypoint.py`, then died at **03:20:13Z** (06:20:13 MSK — the first outage failure). Auto-commit produced `e7ecea84` ("selfevo: auto-commit uncommitted subagent work — Update scaffold_script_entrypoint.py …") at 03:20:13Z, smoke passed, merged as `e1ff2405` at 03:20:36Z, outcome `success` 03:20:40Z. Real work before the call died; not bookkeeping. Filed as #1281 rather than folded in.

## Not in scope

No fallback model group (operator's call, has a cost, and would still have logged `completed` if it also failed).

## Tests — `tests/test_bridge_executor_llm_error.py` (5)

A fake `SubagentManager` whose `spawn` writes exactly the 2026-09-04 failure telemetry and registers a finished task, driving `bridge._main_impl()` end to end on a real temp git repo:

- **dead LLM** → `result_status blocked`, `rollback.reason executor_llm_error`, `commits_pushed 0`, `backlog_title` set from the seeded proposal artifact (as production requests carry it), `key_learnings` carries the provider error, `_recent_failure_match(title)` returns the title; ledger `failed` / `executor_llm_error`; return code 3, and `crash_record.record_exit(outcome="failure", exit_status=3)` — the `__main__` guard's own expression — yields `consecutive_failures 1`; no `handled_` marker, `retry_ … count 1`.
- **bounded retry** → two failures leave it pending (count 1, 2); the third writes the marker; a fourth run does not spawn.
- **healthy control** → `completed`, marker written, no retry file, ledger `success`, exit 0 — unchanged.
- helper unit tests: `_executor_llm_error` reads only the failure shape; `_decide_handled_marker` marker-vs-retry.

Full suite result posted as a comment when the run finishes (baseline 4 by id). `ruff`: 23 findings on bridge.py, identical to `main`.

Deploy and live verification are the operator's.

### DIFF
--- a/nanobot/runtime/bridge.py
+++ b/nanobot/runtime/bridge.py
@@ -2912,6 +2912,7 @@ async def _main_impl_body():
             _decide_handled_marker(
                 handled_marker, req_path,
                 llm_error=bool(_executor_llm_error_text and cycle_commit_count == 0),
+                state_dir=STATE_DIR, cycle_id=_cycle_id,
             )
 
             # ── Closed-loop repair cycle (issue #526) ────────────────────────────
@@ -4212,17 +4213,53 @@ def _llm_error_retries_exhausted(request_id: 'str | None') -> bool:
     the point at which its failed rows start counting for the #716 recent-
     failure suppression. Unknown request or unreadable counter → True: the
     old behaviour (row counts as a failure), never a wider exemption."""
+    path = _llm_error_retry_path(request_id)
+    if path is None:
+        return True
     try:
-        path = _llm_error_retry_path(request_id)
-        if path is None or not path.exists():
+        if not path.exists():
+            # Observable on purpose (#1282 review): a lost counter retires a
+            # request that may still have had budget, and without this line
+            # that is indistinguishable from a normal three-attempt retirement.
+            print(f'executor_llm_error: retry counter {path.name} missing — treating the budget as spent (#1280)')
             return True
         count = int((json.loads(path.read_text(encoding='utf-8')) or {}).get('count') or 0)
         return count >= LLM_ERROR_MAX_RETRIES
     except Exception:
+        print(f'executor_llm_error: retry counter {path.name} unreadable — treating the budget as spent (#1280)')
         return True
 
 
-def _decide_handled_marker(handled_marker: Path, req_path: 'Path | str', *, llm_error: bool) -> str:
+def _recorded_llm_error_attempts(state_dir: 'Path | str', cycle_id: 'str | None') -> int:
+    """#1280 review: the durable witness that a request was previously live.
+    Counts this cycle's ``outcome`` rows with ``reason == executor_llm_error``
+    in the ledger (live + archives, via ``state_access.ledger_window``). The
+    ledger is append-only and rotated, never pruned inside the retention
+    window — unlike ``subagents/results``, which the archiver migrates and
+    prunes within a cycle, and unlike the ``retry_<id>.json`` counter, which
+    is one file. Fail-open to 0 (reads as "first attempt"): a ledger-read
+    failure must not retire a request that nothing says was offered before.
+    """
+    if not cycle_id:
+        return 0
+    try:
+        from nanobot.runtime.state_access import ledger_window
+        since = time.strftime('%Y-%m-%dT%H:%M:%SZ', time.gmtime(time.time() - 30 * 86400))
+        window = ledger_window(Path(state_dir), since_ts=since, phases=frozenset({'outcome'}))
+        return sum(
+            1 for row in window.rows
+            if isinstance(row, dict)
+            and str(row.get('cycle_id') or '') == str(cycle_id)
+            and str(row.get('reason') or '') == 'executor_llm_error'
+        )
+    except Exception:
+        return 0
+
+
+def _decide_handled_marker(
+    handled_marker: Path, req_path: 'Path | str', *, llm_error: bool,
+    state_dir: 'Path | str | None' = None, cycle_id: 'str | None' = None,
+) -> str:
     """#1280: write the ``handled_`` marker — retiring the request forever —
     unless the subagent died on its LLM call without producing anything, in
     which case re-offer the request up to :data:`LLM_ERROR_MAX_RETRIES`
@@ -4230,8 +4267,21 @@ def _decide_handled_marker(handled_marker: Path, req_path: 'Path | str', *, llm_
     it. Bounded on purpose: an unbounded re-offer turns a permanently-bad
     request into an infinite loop, a worse failure than the one being fixed.
 
-    Returns ``"handled"``, ``"retry"`` or ``"retired_after_retries"`` for the
-    journal. Fail-open: any error writes the marker, the pre-#1280 behaviour.
+    Lost-counter rule (#1282 review), the same direction on both sides: the
+    reader ``_llm_error_retries_exhausted`` treats a missing or unreadable
+    counter as budget spent; so does this writer, for a request that was
+    previously live. "Previously live" is witnessed by the ledger
+    (:func:`_recorded_llm_error_attempts`), not by the counter itself — when
+    the counter AND the failed result rows are gone (results migrate to the
+    pruned archive within a cycle), the suppression finds nothing and this
+    is the only place left that can stop an unbounded re-offer. An
+    unreadable counter is always "previously live" (someone wrote it).
+    A missing counter with no recorded attempt is a genuine first attempt.
+
+    Returns ``"handled"``, ``"retry"``, ``"retired_after_retries"`` or
+    ``"retired_state_lost"`` for the journal, and prints the lost-state
+    case so it is distinguishable from an ordinary retirement. Fail-open:
+    any error writes the marker, the pre-#1280 behaviour.
     """
     try:
         if not llm_error:
@@ -4239,11 +4289,23 @@ def _decide_handled_marker(handled_marker: Path, req_path: 'Path | str', *, llm_
             return 'handled'
         retry_path = handled_marker.with_name(handled_marker.name.replace('handled_', 'retry_', 1)).with_suffix('.json')
         count = 0
+        state_lost = ''
         if retry_path.exists():
             try:
                 count = int((json.loads(retry_path.read_text(encoding='utf-8')) or {}).get('count') or 0)
             except Exception:
-                count = 0
+                state_lost = 'unreadable'
+        else:
+            prior = _recorded_llm_error_attempts(state_dir, cycle_id) if state_dir is not None else 0
+            if prior >= 1:
+                state_lost = f'missing after {prior} recorded attempt(s)'
+        if state_lost:
+            handled_marker.write_text(str(req_path), encoding='utf-8')
+            print(
+                f'executor_llm_error: retry counter {retry_path.name} {state_lost} — '
+                'treating the budget as spent and retiring the request (retired_state_lost, #1280)'
+            )
+            return 'retired_state_lost'
         count += 1
         retry_path.write_text(
             json.dumps({
--- a/nanobot/runtime/bridge.py
+++ b/nanobot/runtime/bridge.py
@@ -4196,6 +4196,32 @@ def _executor_llm_error(state_dir: Path, task_id: 'str | None') -> str:
         return ''
 
 
+def _llm_error_retry_path(request_id: 'str | None') -> 'Path | None':
+    """The retry counter `_decide_handled_marker` keeps for a request whose
+    subagent died on the LLM call (#1280): ``retry_<safe_id>.json`` beside
+    the ``handled_<safe_id>.txt`` marker, same ``safe_id`` derivation."""
+    if not request_id:
+        return None
+    safe_id = str(request_id).replace('/', '_')[:120]
+    return BRIDGE_STATE_DIR / f'retry_{safe_id}.json'
+
+
+def _llm_error_retries_exhausted(request_id: 'str | None') -> bool:
+    """#1280: True once the request's LLM-failure retry budget is spent (its
+    ``retry_<id>.json`` count has reached :data:`LLM_ERROR_MAX_RETRIES`) —
+    the point at which its failed rows start counting for the #716 recent-
+    failure suppression. Unknown request or unreadable counter → True: the
+    old behaviour (row counts as a failure), never a wider exemption."""
+    try:
+        path = _llm_error_retry_path(request_id)
+        if path is None or not path.exists():
+            return True
+        count = int((json.loads(path.read_text(encoding='utf-8')) or {}).get('count') or 0)
+        return count >= LLM_ERROR_MAX_RETRIES
+    except Exception:
+        return True
+
+
 def _decide_handled_marker(handled_marker: Path, req_path: 'Path | str', *, llm_error: bool) -> str:
     """#1280: write the ``handled_`` marker — retiring the request forever —
     unless the subagent died on its LLM call without producing anything, in
@@ -4448,6 +4474,11 @@ def _recent_failure_match(
     every later same-vocabulary proposal (the 2026-07-18 decay cascade: 5
     candidates, 11 wasted proposals, zero spawns). Such rows — and any
     ``skipped*`` result_status — are excluded from the scan entirely.
+    Transport deaths are not evidence either (#1280): a row whose
+    ``rollback.reason`` is ``executor_llm_error`` is skipped while its
+    request still has LLM-failure retry budget (see
+    :func:`_llm_error_retries_exhausted`) and counts as a failure only
+    once that budget is spent.
 
     Cross-target precision (#798): result rows also carry the historical
     entry's own ``target_path`` (see :func:`_write_bridge_completed_result`),
@@ -4531,6 +4562,16 @@ def _recent_failure_match(
             if str(status or '').lower().startswith('skipped'):
                 continue
             if not reason and status not in ('blocked', 'no_commit'):
+                continue
+            if reason == 'executor_llm_error' and not _llm_error_retries_exhausted(data.get('request_id')):
+                # #1280: a cycle that died on the LLM call says nothing about
+                # the proposal, so while its request still has retry budget
+                # the row must not feed suppression — otherwise this branch
+                # retires the request on attempt 2 of 3 and the bounded retry
+                # never runs (found in review, reproduced with a
+                # production-shaped request). Once the budget is spent the
+                # row counts like any other failure: a title that keeps
+                # dying earns the 24 h window like everything else.
                 continue
             failures.append(data)
             if len(failures) >= max_scan:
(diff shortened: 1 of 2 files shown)


ozand:
Full suite in the worktree at `c110d938` (`python -m pytest -q -n 4`, log kept): **4 failed, 3019 passed, 17 skipped** in 9m50s. Failure list by id — exactly the Windows baseline:

```
```

CI green on 3.11/3.12/3.13. `ruff` on bridge.py: 23 findings, identical set to `main`.

ozand:
## Blocking: the bounded retry cannot run in production shape

The fix is right on the parts I asked for — `blocked` plus `rollback.reason = executor_llm_error` lands inside the set `_recent_failure_match` already recognises rather than inventing a new spelling, the streak moves, and `_INCONCLUSIVE_REASONS` correctly keeps an outage from counting as a refutation. Those are exactly right.

But the retry and the #716 suppression cancel each other, and the PR's own two tests demonstrate it between them.

`test_dead_llm_cycle_is_blocked_failed_unmarked_and_exits_nonzero` seeds a `source_artifact` — as production always does — and asserts `bridge._recent_failure_match(title, state_dir) == title`. That assertion is correct and it is the problem: on the next offer of the same request, the pre-spawn branch at `bridge.py:2459` matches that title and writes the handled marker at `:2469`. There is no exemption for `executor_llm_error` there.

`test_retry_is_bounded_then_the_request_is_retired` drives the full loop three times and passes — because it seeds `task_title="permanently bad request"` with **no** `source_artifact`. `backlog_title` is then empty, so the recorded failure carries nothing for `_recent_failure_match` to match on, and the retry survives. That is a shape the proposer never produces.

Demonstrated, not argued. I re-ran your loop test with a production-shaped request — same request id, same manager, one added artifact — and the bridge announced it itself:

```
executor_llm_error: request left pending for retry (1/3)
bridge: task "Add markdown catalog link path resolver to workspace_validat" matches recent
  failure/rejection "Add markdown catalog link path resolver to workspace_validat"
  (within 24.0h); skipping subagent spawn
bridge-result: wrote blocked result to result-req-loop.json
already_handled
```

Retry counter stops at 1. Attempts 2 and 3 never happen; the request is retired on attempt 2 of 3 by the suppression branch instead.

### What this needs

1. An exemption so a cycle that died on transport does not feed the recent-failure suppression for its own retry. Decide deliberately whether `executor_llm_error` should be excluded from `_recent_failure_match`'s failure proxy entirely, or only skipped while a retry budget remains — and say which in a comment. There is a real argument either way: an outage is not evidence about the proposal, but a title that keeps dying deserves suppression eventually.
2. Replace the artifact-less fixture in `test_retry_is_bounded_then_the_request_is_retired` with a production-shaped one. A retry test that only passes without the artifact is testing a configuration the loop never reaches.

The second point is the more important one long-term. The green test here was green for the wrong reason, which is how an unexercised guard ships.

Separately: `#1281` for the `commits_pushed: 1` anomaly is the right call, and the full local suite result is still owed on this PR.

ozand:
Reworked per the review, pushed as `c17bcdb8`.

**Decision**: skip while budget remains, count once spent. A row with `rollback.reason == executor_llm_error` is excluded from `_recent_failure_match`'s failure proxy while its request's `retry_<id>.json` count is below `LLM_ERROR_MAX_RETRIES` (`_llm_error_retries_exhausted`); after the third failure the request is retired **and** its rows count like any failure, so a re-minted same-title proposal gets the 24 h window. Reason in the code comment: an outage says nothing about the proposal, but a title that keeps dying still earns suppression eventually. Unknown request or unreadable counter → counts (old behaviour, never a wider exemption).

**Fixture**: `test_retry_is_bounded_then_the_request_is_retired` now seeds a `source_artifact` exactly as the first test does (`backlog_title` set), and asserts per attempt: no `matches recent failure/rejection` line, the `request left pending for retry (n/3)` line, no marker, counter == n, `_recent_failure_match(title) is None`; on the 3rd: retirement line, marker, and `_recent_failure_match(title) == title`; 4th run `already_handled`. Your reproduction (suppression on attempt 2 with an artifact) is exactly the case that would now fail this test.

The first test's assertion is flipped: `None` while budget remains, `== title` after the counter is set to the cap.

Nothing else touched: `blocked` + `rollback.reason`, the streak exit code and `_INCONCLUSIVE_REASONS` are as reviewed. Full local suite is running now; list follows as a comment.

ozand:
Full suite in the worktree at `c17bcdb8` (post-review head; `python -m pytest -q -n 4`, log kept): **4 failed, 3019 passed, 17 skipped** in 10m03s. Failure list by id — exactly the Windows baseline:

```
```

CI green on 3.11/3.12/3.13 for `c17bcdb8`. `ruff` on bridge.py: 23 findings, identical set to `main`.

ozand:
## REVIEW VERDICT: REQUEST CHANGES

The retry exemption has the intended normal-path behavior, but the per-request counter is not durable enough for the state lifecycle this codebase actually uses.

### Finding: counter deletion/corruption reopens the exemption indefinitely

`_llm_error_retries_exhausted()` treats a missing or unreadable `retry_<id>.json` as exhausted, which is the correct fail-open behavior for the suppression reader: an unknown counter counts as a failure and cannot widen the exemption (`bridge.py:4209-4221`, `4566`).

However, `_decide_handled_marker()` treats an unreadable existing counter as `count = 0`, then overwrites it with `count = 1` (`bridge.py:4225-4265`). If the counter is deleted, the same function also starts at zero. This creates a lifecycle-dependent bypass:

1. attempt 1 writes count 1; suppression is exempted;
2. the counter is deleted, rotated, or corrupted before the next offer;
3. the next failed attempt rewrites count 1;
4. the request can receive another three-attempt window every time its retry state is lost.

The reader fallback itself is safe, but the writer does not preserve that safety. A state cleanup, manual repair, or rotation that removes `retry_<id>.json` can therefore prevent the retry budget from ever being spent. The PR should either make the retry state durable/retained with the request lifecycle, or explicitly retire the request when a previously-existing counter becomes unavailable rather than resetting it to zero. Add a test that deletes/corrupts the counter between failed attempts and proves the request cannot regain budget indefinitely.

### Covered paths

- Unknown request id or missing counter in `_recent_failure_match`: counts normally, because `_llm_error_retries_exhausted()` returns `True`.
- Unreadable/corrupt counter in `_recent_failure_match`: counts normally via the same fallback.
- `_task_already_done` remains a separate earlier gate and can suppress a request based on recent git history before the retry exemption is consulted; this is unchanged and appropriate.

### Re-minted same-title request

The exemption is keyed by `request_id`, not title. A new request id with the same title has no counter and therefore counts as a failure immediately; it is not granted the original request's remaining budget. That is consistent with the documented per-request policy, though it means retries represented by a new id are treated conservatively and may be suppressed by the old row.

### Verdict

Please address the counter lifecycle/reset case and add a regression test. The happy path and the stated unknown/unreadable reader fallback do not cover this writer-side state-loss path.

ozand:
## Orchestrator: the finding is real, the mechanism in it is not. Corrected below.

I ran all three cases rather than reasoning about them, because I had talked myself into two opposite answers.

**Counter deleted between attempts → the request is retired, not given a new window.**

```
DELETED-> marker=True  count=None
```

**Counter corrupted between attempts → same, and the file is not even rewritten.**

```
CORRUPT-> marker=True  raw={not json
```

The reason is ordering. `_llm_error_retries_exhausted` runs first, inside `_recent_failure_match` on the pre-spawn path; on a missing or unreadable counter it returns `True`, the failed row counts, the suppression branch fires and writes the handled marker at `bridge.py:2469`. `_decide_handled_marker`'s `count = 0` reset sits on the post-spawn path and is never reached. So counter loss alone is fail-safe in the conservative direction — the opposite of a bypass.

**The conclusion still stands, under a narrower condition.** The writer's reset becomes reachable only when the failure *rows* are gone too, so the suppression finds nothing to match. That is not an exotic condition here: `subagents/results` migrates into `archive/` within a cycle, and the archive is pruned — `current_health.json` on the host currently reads `subagent_archive_pruned_count: 4176`. Simulating a cleanup that takes both, per cycle:

```
run1: marker=False count=1
run2: marker=False count=1
run3: marker=False count=1
run4: marker=False count=1
```

Unbounded. The marker is never written, the count never advances past 1, and this is exactly the infinite re-offer the bound exists to prevent.

### What #1282 needs

1. Make the writer's fallback match the reader's. `_decide_handled_marker` currently treats *missing* and *unreadable* as `count = 0`; the reader treats both as exhausted. Pick one direction and apply it on both sides — the safe one is that a previously-live request whose counter cannot be read is retired, not restarted.
2. A regression test for the demonstrated case: counter **and** result rows removed between attempts, asserting the request cannot be re-offered indefinitely. Not the counter alone — that case already retires, and a test of it would pass while proving nothing.

### Secondary, worth deciding rather than leaving

Counter loss alone silently retires a request that still had budget. That is the right failure direction, but there is no journal line saying it happened, so it is indistinguishable from a normal retirement after three real attempts. One `print` on that branch makes the difference observable.

### Credit where it is due

The reader/writer asymmetry was a genuine catch and I would not have looked for it. The review's error was in the failure path it inferred from that asymmetry, not in noticing it.

ozand:
Round 3 pushed as `ab044779`, built from the three runs in your comment, not the review text.

**Direction chosen, applied on both sides**: a previously-live request whose `retry_<id>.json` cannot be read is **retired**. The reader already did that; the writer now does too. "Previously live" is witnessed by the **ledger**, not by the counter or the result rows — those are exactly what goes missing. `_recorded_llm_error_attempts(state_dir, cycle_id)` counts this cycle's `outcome` rows with `reason == executor_llm_error` over live + archives through `state_access.ledger_window` (append-only, never pruned inside retention). Rules: unreadable counter → retire (someone wrote it); missing counter with ≥1 recorded attempt → retire; missing with none → genuine first attempt, count 1. Fail-open on a ledger read error is "first attempt" — a read failure must not retire a request nothing says was offered before.

**Observable**: both lost-state paths print. Writer: `executor_llm_error: retry counter retry_<id>.json missing after N recorded attempt(s) — treating the budget as spent and retiring the request (retired_state_lost, #1280)`. Reader: `retry counter <name> missing|unreadable — treating the budget as spent`. A lost-state retirement is now distinguishable from a three-attempt one in the journal.

**Regression test for your demonstrated case**, `test_lost_counter_and_result_rows_between_attempts_cannot_reoffer_indefinitely`: production-shaped request; attempt 1 → count 1; then the counter **and every `result-*.json`** are removed; `_recent_failure_match` returns `None` (nothing to match) while `_recorded_llm_error_attempts == 1`; attempt 2 → `retired_state_lost` line, marker written, no counter; attempt 3 (state lost again) → `already_handled`. The counter-only case is deliberately **not** a test, per your note — it already retires via suppression and would pass while proving nothing. Helper test covers unreadable / missing-without-witness / missing-with-witness plus the reader's answer.

Untouched: `blocked` + `rollback.reason`, the exit-code streak, `_INCONCLUSIVE_REASONS`, and the retry exemption's normal path. `tests/test_bridge_executor_llm_error.py`: 7 passed. `ruff` on bridge.py: 23, same set as `main`. Full local suite running; list follows as a comment.

ozand:
Full suite in the worktree at `ab044779` (round-3 head; `python -m pytest -q -n 4`, log kept): **4 failed, 3021 passed, 17 skipped** in 8m59s. Failure list by id — exactly the Windows baseline:

```
```

CI green on 3.11/3.12/3.13 for `ab044779`. `ruff` on bridge.py: 23 findings, identical set to `main`.

ozand:
Full local suite, run by me on `ab044779` since the author's session was out of context before it landed:

```
4 failed, 3021 passed, 17 skipped, 38 warnings in 538.21s

```

Exactly the four-item Windows baseline, by list not by count. No fifth failure. CI green on 3.11/3.12/3.13.

Independent re-verification of the case that blocked round 2 — counter and result rows removed between every attempt, the scenario that ran unbounded before:

```
run1: marker=False count=1
run2: marker=True  count=None    <- retired
```

Previously this ran four rounds with the count stuck at 1 and no marker ever written. It now terminates on the second offer.

Merging. Recorded for whoever picks this up next: the author's own least-certain claim is the 30-day `ledger_window` in `_recorded_llm_error_attempts` — a request re-offered more than 30 days after its first failure reads as a first attempt and gets one fresh window. The age distribution of pending requests was not measured. It is a bounded, single-window exposure rather than the unbounded loop this PR closes, but it is unmeasured and should not be forgotten.

PR #1276, opened 2026-09-04
fix(runtime): bound archived result evidence readers

ozand:
## Summary

Implements the D4/D11 portion of issue #1173/#1176's residual audit.

- Extends `state_access.artifacts` with explicit directory selection and a payload-key filter.
- Applies the `newest` bound after candidate filtering, so `requests/` and successful/unrelated artifacts cannot evict relevant result evidence.
- Migrates `demand._result_file_defects` to bounded `results/` + `archive/` reads.
- Migrates `usage_evidence._touched_from_results` to bounded `results/` + `archive/` reads requiring `files_changed`.
- Preserves selected source paths in `Window` so touched evidence continues to use result-file mtimes.

## Caller classification

- Results-only: `demand._result_file_defects`, `demand._noop_evidence`, bridge failure/attempt readers, and `existence_index` attempt-title readers.
- Requests-only: bridge/backlog/health queue readers.
- Genuinely multi-directory: `state_access.artifacts` itself, now parameterized rather than forcing every consumer into the pooled `results + requests + archive` view.

## Verification

- Focused: 282 passed (`tests/test_state_access.py`, `tests/test_demand.py`, `tests/test_usage_evidence.py`).
- Full suite: 3002 passed, 17 skipped, 11 warnings, with exactly the four established baseline failures:
  - `tests/test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle`
  - `tests/test_bridge_locking.py::TestAcquireBridgeLock::test_acquires_when_free`
  - `tests/test_bridge_locking.py::TestAcquireBridgeLock::test_second_acquire_in_same_process_is_contended`
  - `tests/test_bridge_locking.py::TestAcquireBridgeLock::test_contended_lock_via_monkeypatched_flock`
- Ruff passed on all changed Python files.
- New tests demonstrate an archived result is observed by both D4 and D11, and that an unreadable selected artifact source produces `Window.status == "unavailable"`.

## Rollout

No host access, deployment, `current` flip, service restart, or merge was performed. Operator review/merge and rollout remain pending.

PR #1283, opened 2026-09-04
docs(adr): close bounded state access audit

ozand:
## Summary

Closes the documentation/audit remainder of #1173 after the implementation PRs merged.

- Records the D4/D11 result/archive migrations and D12's remaining Class-C display limitation.
- Separates proven source/call-site findings from audit judgement for indirect readers.
- Defines the bounded state-access policy over `ledger_window`, `artifacts`, `latest_file`, and `sidecar`.
- Records the retired CycleArchive dead-feature decision and `orphan:#1225` cleanup.
- Names #1280's producer-side truthfulness requirements as an explicit non-goal: the reader contract would not catch a false-success writer.
- Adds an ADR Test Contract, including the unclosed audit-judgement item.

## Verification

- `ruff check nanobot/runtime/state_access.py nanobot/runtime/demand.py nanobot/runtime/usage_evidence.py` — passed.
- Focused contract tests: 48 passed:
  - `tests/test_state_access.py`
  - `tests/test_class_a_windows.py`
  - `tests/test_state_access_hygiene.py`
  - `tests/test_state_path_writers.py`
- No host access, deployment, current-symlink switch, service restart, or merge was performed.

### DIFF
--- a/docs/adr/ADR-002-bounded-state-access.md
+++ b/docs/adr/ADR-002-bounded-state-access.md
@@ -9,7 +9,11 @@ tags: [runtime, state, reliability]
 
 # Status
 
-Accepted as the foundation for the state-reader migrations tracked by #1175–#1179.
+Accepted — implemented for the migrations completed under #1173–#1179.
+
+The dead-feature audit and residual reader decisions for #1173 are recorded
+below. This ADR does not claim that every historical raw reader has been
+migrated; the hygiene allowlist remains a floor, not a compliance certificate.
 
 # Context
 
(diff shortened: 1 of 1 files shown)


ozand:
**Verdict: APPROVE once two citations are corrected — the document tells the truth about its limits; two of its evidence pointers do not exist as written.**

Reviewed as a docs PR: does the ADR distinguish what is proven from what is judged, does it name the boundary it does not cover, and do its claims about the world hold. Checked against `origin/main`, the deployed release, and the PR states — not against the text alone.

## What holds

- **Proven vs judgement is separated, and honestly.** The "Evidence strength is intentionally split" paragraph puts D4/D11 directory roles, the `CycleArchive` consumer and the state-path cleanup under source + call-site + tests, and puts the indirect bridge/existence-index classifications under "current call paths and manual scanners … audit judgement, not complete call-graph proof". The Test Contract table then carries a row for exactly that gap with status `audit judgement / not fully written`, and the Status header says the hygiene allowlist "remains a floor, not a compliance certificate". That is the shape the author promised.
- **#1280 is a named non-goal, and the boundary is described correctly.** "A writer-side integrity bug where an executor failure was serialized as `result_status: completed`, `exit_streak.json` retained `consecutive_failures: 0`, and the request was unconditionally marked handled" is precisely what #1282 fixes, and the three requirements listed (terminal status from the actual executor outcome; durable failure-streak update; handled marker only after truthful terminalization) are the three things that PR changes. The reader contract could not have caught it: `_bridge_status = 'completed'` was written on a path that never read the executor's telemetry. Correctly outside this ADR.
- **Dead-feature / migration claims spot-checked on `origin/main`:**
  - D4 `demand._result_file_defects` → `artifacts(… directories=_RESULT_DIRS)` ✔; D11 `usage_evidence._touched_from_results_with_status` reads the live+archive union ✔ (both from #1276, merged).
  - D12 `state._live_recent_outcomes` still `_LIVE_LEDGER_TAIL_LINES = 200` over the live file ✔ — the ADR's "remains a Class-C display reader … known presentation ambiguity" is exact.
  - `CycleArchive` retirement: #1278 is merged (`56ea94d3`); `CycleArchive` survives on `main` only as one history mention each in `archive.py`/`bridge.py`/`evolution_tree.py`; `orphan:#1225` and its `ORPHAN_ISSUES` reason are gone ✔.
  - Test names cited in the table all exist: `test_artifacts_filters_before_newest_bound_and_supports_directory_selection`, `test_artifacts_unavailable_when_selected_sources_cannot_be_read` (test_state_access.py), `test_archived_result_defect_is_visible` (test_demand.py), `test_nothing_in_the_runtime_reads_cycle_archive_or_stalls` (test_bridge_cycle_branch.py) ✔.

## Two citations to correct before merge

1. **`guard fix 0287e623` does not exist.** Not on `origin/main`, not on this PR's branch (its only commit is `8d173669`), not in any ref I can see. A commit hash in an ADR is a claim a future reader will `git show`; either cite the real commit or drop the phrase. (The row's status column, "dynamic reflection/runtime string-building remain outside static coverage", is accurate and should stay.)
2. **`TestRankingWithoutRewardOrTrigger` is in `tests/test_evolution_tree.py`**, not `tests/test_bridge_cycle_branch.py` as the table says.

## One wording fix (non-blocking)

"the instance-side frozen dashboard is a separately governed retired-history surface" reads as done. The change that labels it — ozand/eeebot-self-evolving#185 — is **open**, not merged; until then the :8080 page still presents the frozen list unlabelled. Suggest "is being labelled as a retired-history surface (eeebot-self-evolving#185, open)".

## For the record, not a finding

- "`cycle_archive.json` is now inert **in this repository**" is correctly scoped: the deployed release is `20260904T023137Z-canonical-76bab401`, which predates #1278, so the host's bridge still constructs `CycleArchive()` until the next deploy. The ADR does not claim otherwise.
- The indirect readers under "audit judgement" are not enumerated. The hygiene allowlist (`tests/test_state_access_hygiene.py::ALLOW`) is the concrete list; naming those entries in the paragraph would let the next reader know exactly which classifications to distrust. Improvement, not a defect.

Nothing invented: with the two citations fixed, this is an accurate closing record of #1173.

ozand:
Orchestrator note on finding 1, because the provenance matters for the fix.

`0287e623` was a real commit — mine, the guard-test strengthening on the #1225 branch, written after the reviewer flagged that the retirement guard was narrower than its claim. It is unreachable now for a reason worth naming rather than treating as a typo: #1278 was **squash**-merged as `56ea94d3` and its branch deleted, so every pre-squash hash on it became a dangling pointer the moment the PR landed. The object still resolves in my local worktree; it will resolve nowhere a future reader looks.

So the finding stands and the fix is right, but the general form is worth carrying: **a commit hash cited in a durable document must be a post-merge hash.** Citing a branch commit in an ADR guarantees a dead pointer, because the branch is deleted as part of merging it. Cite `56ea94d3` (the squash of #1278) here.

Finding 2 is correct as written — `TestRankingWithoutRewardOrTrigger` lives in `tests/test_evolution_tree.py`.

The wording fix on the instance-side dashboard is also right and I would take it rather than treat it as optional: `eeebot-self-evolving#185` is open, the :8080 page still shows the frozen list unlabelled, and an ADR that reads as done when it is not is exactly the failure this ADR is about.

The "for the record" note on the deployed release is a good catch and needs no change: `20260904T023137Z-canonical-76bab401` predates #1278, so the host's bridge does still construct `CycleArchive()` until the next deploy. The ADR's "inert in this repository" scoping is correct precisely because it does not claim otherwise.

On the non-finding about enumerating the audit-judgement readers: take it. The paragraph currently tells a reader that some classifications are judgement without telling them which, and `tests/test_state_access_hygiene.py::ALLOW` already holds the concrete list. Naming them costs one sentence and is the difference between a stated limit and a vague one.

PR #1288, opened 2026-09-04
fix(strategist): cut prompt sections by declared priority, not key position (#1284)

### LINKED ISSUE #1284: Strategist prompt halving cuts scorecard.latest by key position: drops gaps/control_plane, keeps schema_version
## Context (self-contained)

`nanobot/runtime/strategist.py::_halve_section` shrinks the strategist prompt when it exceeds `_MAX_PROMPT_CHARS` (48,000). For a dict section it halves the largest member by **key position** (`dict(list(member.items())[: len(member) // 2])`). That rule is right for `funnel.by_demand_id` (ordered newest-first by design), but `scorecard.latest` has no meaningful key order.

Observed on the host, run `2026-09-04T00:00:21Z` (release `20260904T023137Z-canonical-76bab401`), `decisions.jsonl` row: `"halved": ["scorecard.latest"]`, `prompt_chars: 43391`. `scorecard/latest.json` had 16 keys in this order and size (bytes of JSON):

```
schema_version 14 | computed_at_utc 29 | window_days 1 | loop 478 | cost 121 | quality 93 | value 194 | heldout 91
integrity 64 | feeds 894 | bridge 294 | knowledge_lift 127 | control_plane 2427 | reader_status 559 | gaps_status 10 | gaps 367
```

The first 8 were kept, so the strategist saw three metadata keys (`schema_version`, `computed_at_utc`, `window_days`) and lost `integrity`, `feeds`, `bridge`, `knowledge_lift`, `control_plane`, `reader_status`, `gaps_status`, `gaps`. Partial mitigation exists in the same prompt: `history_7d.series` carried numeric leaves for `feeds.*`, `integrity.*`, `bridge.*`, `knowledge_lift.*`, `reader_status.*`, and the per-gap `futility` section is never halved. What was genuinely lost: the `gaps` table (metric/target/current per open gap) and `control_plane`.

Which section gets halved varies per run (a read-only reconstruction later on 2026-09-04 halves `recent_cycles` instead), and the halving is recorded under `inputs_status.halved`, so this is observable and bounded — a quality defect, not a safety one. Found during the #999 verify-then-close.

## Proposed fix (keep it small)

In `_halve_section`, when the member being halved is a dict whose section is not order-meaningful (i.e. anything other than `funnel`), drop the **largest** members first instead of the second half by position — or, simpler and sufficient here: exempt the small metadata keys and drop `control_plane` first. One function, one test asserting that halving `scorecard.latest` keeps `gaps` and drops `control_plane`.

## Acceptance

- [ ] Test: a 16-key `scorecard.latest` fixture with the live sizes above, prompt forced over the cap, asserts `gaps`, `loop`, `value`, `feeds` survive and `control_plane` is dropped first; `inputs_status.halved` still records the label.
- [ ] `funnel.by_demand_id` halving behaviour unchanged (existing test `test_prompt_stays_within_cap_and_records_halved_sections` still green).
- [ ] Live: the next natural run whose row has `halved` non-empty is quoted with the post-halving `scorecard.latest` keys.

ozand:
Closes #1284.

## Defect

`_halve_section` shrank an over-budget prompt section by keeping the first half of a dict **by key position**. On the 2026-09-04 03:00 run that cut `scorecard.latest` to `schema_version, computed_at_utc, window_days, loop, cost, quality, value, heldout` and dropped `integrity, feeds, bridge, knowledge_lift, control_plane, reader_status, gaps_status, gaps` — three metadata keys worth 44 bytes survived, the gap table the strategist exists to reason about did not. The same function cut `recent_cycles` and `prior_decisions` (oldest-first ledger tails) by keeping their **oldest** half.

## Fix

Position is a proxy for nothing, so each leaf declares its own cheap end in `_CUT_POLICY` (`nanobot/runtime/strategist.py`):

- a tuple names dict keys from cheapest to load-bearing (`scorecard.latest`: metadata, then `control_plane`, `reader_status`, `bridge`, `knowledge_lift`, `heldout`, `integrity`, then `cost`, `quality`, `feeds`, `value`, `loop`, `gaps_status`, `gaps`); keys the policy does not know go before any declared one;
- `head`: rows arrive oldest-first, the head is expendable (`recent_cycles`, `prior_decisions`, `insights.errors`);
- `tail`: the producer ranked the member best-first, the tail is expendable (`funnel.by_demand_id`, `scorecard.history_7d.series`, `lessons`, `insights.cards`).

A path not in the policy is a container and the step descends into its largest member, so the same function serves every halving section. A section with nothing left to cut yields to the next largest instead of falling through to the `{"truncated": true}` payload. `samples` joins `columns` as an axis that is never cut.

Observability: `inputs_status.halved` keeps its section labels; a new `inputs_status.dropped` list records each step as `{"section": "scorecard.latest", "dropped": 8, "keys": [...]}` for dicts or `{"section": "recent_cycles", "dropped": 25, "end": "head"}` for sequences, in the prompt and in the `decisions.jsonl` row.

Not changed: the 48,000 budget, what the strategist reads or writes, `_HALVING_SECTIONS`, the refusal rule.

## Tests (`tests/test_strategist_inputs.py`)

- `test_cut_step_drops_declared_cheap_keys_first_and_names_them` — live 16-key `latest` with host byte sizes; asserts the eight dropped keys by name and that a second step keeps cutting from the cheap end.
- `test_cut_step_order_is_what_pins_the_survivors` — reversed declaration on the same input loses `gaps` and keeps `schema_version`: the pin is the order, the test fails if it is reversed.
- `test_cut_step_drops_undeclared_keys_before_any_declared_one`
- `test_cut_step_takes_the_cheap_end_of_each_sequence` — head/tail per member, axes untouched, `insights.indexes` drops `docs/index.md` first.
- `test_cut_step_yields_when_nothing_is_left`
- `test_every_prompt_leaf_declares_its_cheap_end` — walks the live-shaped archive; any list/str/dict-of-scalars reachable by a step must be in the policy.
- `test_live_shape_over_budget_keeps_gaps_and_drops_control_plane_first` — the 09-04 run replayed; `gaps/gaps_status/loop/value/feeds/quality/cost` survive, `control_plane` and `schema_version` are in the record, record travels with the prompt.
- existing `test_prompt_stays_within_cap_and_records_halved_sections` extended with the `dropped` records; `inputs_status` key-set test updated.

Focused: 73 passed, 2 skipped. Full suite on Windows at `c6053bab`: `5 failed, 3020 passed, 17 skipped` in 21 min. Four are the exact Windows baseline (`test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle` and the three `test_bridge_locking.py::TestAcquireBridgeLock` cases). The fifth, `test_eeebot_dashboard_truth.py::test_queue_reference_keeps_raw_internal_value_until_public_sanitization`, is a path-separator assertion (`'\secret\a.json' != '/secret/a.json'`) added on `main` by #1275 (`632bffbb`, Linux CI green); this branch does not touch the dashboard, so it is a new Windows-only baseline entry, not a regression here.

Next natural run whose row has `halved` non-empty: quote `inputs_status.dropped` and confirm `gaps` is among the surviving `scorecard.latest` keys (or that the cut fell on another section per its declared end).

### DIFF
--- a/nanobot/runtime/strategist.py
+++ b/nanobot/runtime/strategist.py
@@ -28,12 +28,37 @@ _MAX_JSON_BYTES = 256_000
 _MAX_LINES_FILE_BYTES = 256_000
 # Decision reason when the archive view is too empty to advise on (#1182).
 REASON_INPUTS_UNAVAILABLE = "inputs_unavailable"
-# Prompt budget: while the payload is over the cap, the largest of these
-# sections is halved (goals, the tree digest, the futility sidecar and
-# inputs_status are never shrunk); every halving is recorded in inputs_status.
+# Prompt budget (#1182, #1284): while the payload is over the cap, the largest
+# of these sections loses one step (goals, the tree digest, the futility
+# sidecar and inputs_status are never shrunk). Position in a dict is a proxy
+# for nothing, so WHAT a step removes is declared per leaf in _CUT_POLICY,
+# cheapest first:
+#   - a tuple names dict keys from cheapest to load-bearing; undeclared keys go
+#     before any declared one and the load-bearing tail survives longest
+#   - _HEAD_CHEAP: rows arrive oldest-first, the head is expendable
+#   - _TAIL_CHEAP: the producer ranked the member best-first, the tail is expendable
+# A path not in the policy is a container: the step descends into its largest
+# member. Every step is recorded under inputs_status.dropped with the keys or
+# the row count it removed, so the record says what the model did not see.
 _HALVING_SECTIONS = ("lessons", "prior_decisions", "recent_cycles", "insights", "funnel", "scorecard")
 _HALVING_STEPS = 48  # 6 sections x 8 halvings (1/256 each) before the truncated fallback
-_HALVING_KEEP = {"columns"}  # table headers inside a section, not data
+_HALVING_KEEP = {"columns", "samples"}  # table headers and axes inside a section, not data
+_HEAD_CHEAP, _TAIL_CHEAP = "head", "tail"
+_CUT_POLICY: dict[str, tuple[str, ...] | str] = {
+    "lessons": _TAIL_CHEAP,  # insights_input sorts legacy rows newest first
+    "prior_decisions": _HEAD_CHEAP,  # decisions.jsonl tail, oldest first
+    "recent_cycles": _HEAD_CHEAP,  # cycles.jsonl tail, oldest first
+    "insights.cards": _TAIL_CHEAP,  # newest first
+    "insights.errors": _HEAD_CHEAP,  # errors.yaml tail, oldest first
+    "insights.indexes": ("docs/index.md", "memory/index.md"),
+    "funnel.by_demand_id": _TAIL_CHEAP,  # most recently proposed id first
+    "scorecard.history_7d.series": _TAIL_CHEAP,  # series that moved sort first
+    "scorecard.latest": (
+        "schema_version", "computed_at_utc", "window_days",  # metadata, 44 bytes together
+        "control_plane", "reader_status", "bridge", "knowledge_lift", "heldout", "integrity",
+        "cost", "quality", "feeds", "value", "loop", "gaps_status", "gaps",  # what the advice is about
+    ),
+}
 
 def _now() -> str:
     return dt.datetime.now(dt.timezone.utc).isoformat().replace("+00:00", "Z")
@@ -153,6 +178,7 @@ def collect_inputs(state_root: Path, repo_root: Path) -> dict[str, Any]:
             "evolution_tree": tree_meta,
             "recent_cycles": len(recent),
             "halved": [],  # filled by build_strategist_prompt when it shrinks sections
+            "dropped": [],  # what each step removed: keys for dicts, row count and end for sequences
         },
     }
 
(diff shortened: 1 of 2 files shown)


ozand:
## REVIEW VERDICT: APPROVE

I independently re-derived the key behavior of PR #1288 at commit `c6053bab`; the PR is already merged and its CI is green.

- **Generalization:** `_cut_step` is not scorecard-specific. It descends through an unknown nested container to its largest cuttable member, while explicit leaves use `_CUT_POLICY`. `recent_cycles` is explicitly declared with `_HEAD_CHEAP`, and the sequence tests exercise `recent_cycles`, `prior_decisions`, `lessons`, `funnel.by_demand_id`, `insights` members, and `scorecard.history_7d.series`.
- **Undeclared keys:** for tuple-policy dict leaves, `rank.get(name, -1)` sorts undeclared keys before every declared key, so a newly added key is dropped before load-bearing declared data rather than silently entering the survivors.
- **Observability:** each halving iteration appends a record to `inputs_status.dropped`, with exact removed dict keys or sequence count/end. The prompt payload carries the same status object, and the tests assert equality after progressive cuts; repeated cuts remain separately visible.
- **Budget:** `_MAX_PROMPT_CHARS` remains `48_000`; the halving loop and section limits were not raised.
- **Order pin:** the reversed-policy test mutates the declaration, then asserts `gaps`/`loop` are removed and `schema_version` survives. Reversing the intended priority therefore fails the test's expected production outcome; the test is non-vacuous.

I found no additional actionable defect in the requested attack surface. **Approve.**

Issue #51, opened 2026-03-31
[P2-OWNER-UTILITY-001] Add owner-facing utility surfaces after runtime stabilizes


ozand:
## Metadata
- **Task ID:** `P2-OWNER-UTILITY-001`
- **WSJF:** 10.0
- **Workstream:** Memory and owner utility
- **Owner:** agent-product
- **Status:** Active

## User story
- **Link:** `docs/userstory/SUPERMIND_OPERATOR_BOOST.md`
- **Task:** Add owner-facing utility surfaces after runtime stabilizes

## References
- `docs/PLATFORM_EVOLUTION_IDEAS.md`

## Notes
- Downstream of current runtime stabilization.

ozand:
Taking this into active execution now. Delegating a focused audit of existing owner-facing utility surfaces and the smallest missing bounded slice, then implementing it in this same execution thread.

ozand:
Continuing active execution now. Delegating two focused audits for the smallest missing owner-facing utility slice, then implementing the chosen bounded surface in the same execution thread.

ozand:
Roll-out complete.

What landed:
- producer-side control-plane summary now includes a compact `owner_utility` snapshot
- owner utility reports current state, reason, primary action, and compact evidence
- dashboard `/system` now renders an Owner utility card without adding a new route or broad UI surface

Commits:
- eeebot: `e459513` feat: add owner utility snapshot to control plane summary
- eeebot-ops-dashboard: `6c10816` feat: show owner utility in system view

Verification:
- main bounded suite: 107 passed
- dashboard suite: 37 passed
- live `/api/system` contains `owner_utility` and `control_plane`
- live `/system` contains `Owner utility` and `Primary action`

ozand:
Completed and rolled out locally. Owner-facing utility is now surfaced as a compact control-plane summary and dashboard system card.

PR #1293, opened 2026-09-04
fix(usage): a cap-bounded touch read is complete, not partial (#1290)

### LINKED ISSUE #1290: Decay lane permanently blocked: #1273 marks the bounded touch read partial whenever the archive exceeds 50 files
## Context (self-contained)

#1273 (merge `77c6fd9c`, live in release `20260904T212321Z-canonical-632bffbb` since 2026-09-05 00:23 MSK) made `usage_evidence` read touch evidence from `subagents/results` **and** `subagents/archive`, bounded to the newest `_MAX_RESULT_FILES = 50` files, and persist the read's status. A bounded read is reported as `partial`, and `partial` is in `_TOUCH_EVIDENCE_BLOCKING_STATUSES`, so `decay_candidates` returns `[]` (`usage_evidence.py` ~line 1382: `if touch_status in _TOUCH_EVIDENCE_BLOCKING_STATUSES: return []`). Fail-closed on purpose: never decay a script whose touch evidence might sit in the unread part of the archive.

On the host the read is bounded **every time**: `_touched_from_results_with_status` sets `capped = len(entries) > 50`, and the two directories hold 6 + 3,559 JSON files (cleanup keeps the archive, it never prunes it to 50). Observed after the flip:

```
state/usage/last_used.json  scanned_at_utc 2026-09-04T21:55:01Z  touched_results_status: "partial"
```

So the decay lane is off by construction on this host, for as long as the archive holds more than 50 files, i.e. permanently. Same class as #1183 / #1178 / #1166 — a size cap in front of a read that turns the guarded feature into a silent off-switch.

Context that keeps this from being an emergency: decay proposals were already zero for five days before the flip (`decay-*` `proposed` rows per day, last 7 day-files: `2026-08-30: 7`, nothing since). Nothing visible changed at the flip; what changed is that the lane can no longer come back on its own.

## Decision needed

Either the bound must not mark the read `partial` when it is the *only* reason (e.g. treat "50 newest of N" as complete for touch evidence, since touch is a newest-wins signal and a script touched only in the un-read older tail is by definition not recently touched), or the archive must be pruned to the cap, or the lane is accepted as retired and labelled so in `state_paths` per the decommission rule. Recommendation: the first — the cap already keeps the newest files, and decay asks "touched recently?", which the newest 50 answer.

## Acceptance

- [ ] Test: 60 archived results, all older than the touch window, one live result → status is not blocking and `decay_candidates` is non-empty.
- [ ] Test: archive `missing`/`unavailable`/`corrupt` still blocks (unchanged).
- [ ] Live: `last_used.json` `touched_results_status` reads `complete` after deploy with the archive still >50 files, quoted with `scanned_at_utc`.

ozand:
Closes #1290.

## Defect

#1273 (`77c6fd9c`, live since release `20260904T212321Z-canonical-632bffbb`) reads touch evidence from `subagents/results` + `subagents/archive`, newest 50 files, and marks the read `partial` whenever the union exceeds 50. `partial` is in `_TOUCH_EVIDENCE_BLOCKING_STATUSES`, so `stale_artifacts` returns `[]`. The live archive holds 3,559 files, so the read is `partial` on every run and the decay lane is off by construction. Observed on the host: `state/usage/last_used.json` `touched_results_status: "partial"`, `scanned_at_utc 2026-09-04T21:55:01Z`.

**Class: guard cheaper than what it guards** — a size cap in front of a read that turns the guarded feature into a silent off-switch. Same pattern as #1183, #1178, #1166.

## Fix (the authorised arm: newest-50 is complete for touch evidence)

One line of logic in `_touched_from_results_with_status`: the cap no longer sets `partial`. Touch is a newest-wins signal and decay asks "touched recently?"; a file older than the whole window cannot carry a newer touch than anything inside it, so the newest 50 answer the question completely.

The distinction is the whole fix, so it is exact:

| condition | before | after |
|---|---|---|
| union > 50 files, both dirs present, all readable | `partial` (blocks) | `complete` |
| one expected directory missing | `partial` | `partial` (blocks) |
| directory unreadable (`PermissionError`/`OSError`) | `partial`/`unavailable` | unchanged (blocks) |
| unreadable / corrupt / undatable file inside the window | `unavailable`/`corrupt`/`partial` | unchanged (blocks) |
| `subagents/` absent | `missing` | unchanged (blocks) |
| both dirs present and empty | `valid-empty` | unchanged |

Nothing else changes: `_MAX_RESULT_FILES` stays 50, the sort stays `(mtime, directory, name)` descending, `_TOUCH_EVIDENCE_BLOCKING_STATUSES` is untouched, `refresh_usage` still persists the status. Module docstring updated to state the rule and why.

## Tests (`tests/test_usage_evidence.py`)

- `test_capped_reader_is_complete_and_decay_proceeds` (replaces `test_capped_reader_is_partial_and_blocks_decay`): 60 archived results all older than the touch window + one live result → `complete`, `len(touched) == 50`, `stale_artifacts` returns the old script.
- `test_cap_does_not_launder_a_missing_dir`: 60 archived, `results/` absent → `partial`, `stale_artifacts == []`.
- `test_cap_does_not_launder_a_corrupt_file_inside_the_window`: 60 archived + a corrupt newest file → `corrupt`, `stale_artifacts == []`.
- Two ordering tests (`…newest_deterministic_bounded_union`, `…equal_mtime_boundary…`) asserted `partial` incidentally for a capped read with both dirs present; now assert `complete`. Their ordering assertions are unchanged.
- Existing blocking tests unchanged and green: `test_one_missing_result_dir_is_partial`, `test_missing_result_evidence_blocks_decay`, `test_unknown_sidecar_status_blocks_decay`, `test_malformed_result_is_explicit_and_blocks_decay`, `test_decay_refreshes_status_within_watermark`.

Focused: `tests/test_usage_evidence.py` 98 passed. Full Windows suite at `181f16bb`: `5 failed, 3022 passed, 17 skipped` in 21 min — the four named baseline failures (`test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle` and the three `test_bridge_locking.py::TestAcquireBridgeLock` cases) plus `test_eeebot_dashboard_truth.py::test_queue_reference_keeps_raw_internal_value_until_public_sanitization`, the Windows path-separator assertion introduced on `main` by #1275 (`632bffbb`, Linux CI green; also seen on #1288). Not touched by this branch.

## Live check — needs a deploy

After the release carrying this commit is `current`, the next refresh (bridge cycle) must write `state/usage/last_used.json` with `touched_results_status: "complete"` **while `subagents/archive` still holds >50 files** (3,559 today); quote the row with its `scanned_at_utc`. Then the first `decay-*` `proposed` row, if any script qualifies, confirms the lane is back; zero decay proposals after that is a content answer, not a guard answer.

### DIFF
--- a/nanobot/runtime/usage_evidence.py
+++ b/nanobot/runtime/usage_evidence.py
@@ -31,15 +31,14 @@ so the three signals are all things the harness can observe on disk itself:
 Touch evidence scans the union of ``subagents/results/`` and the flat
 ``subagents/archive/`` newest-first, with a deterministic ``(mtime, directory,
 name)`` descending order and a shared bound of :data:`_MAX_RESULT_FILES`.
-The bound alone does not make the read ``partial`` (#1290): touch is a
-newest-wins signal and decay asks "touched recently?", so the newest
-:data:`_MAX_RESULT_FILES` files answer it completely — a script touched only
-in the unread older tail is by definition not recently touched. Any live host
-holds thousands of archived results, so a cap that meant ``partial`` was a
-permanent off-switch for the decay lane. What DOES block: a missing or
-unreadable directory, an unreadable/corrupt file inside the window, or a file
-whose mtime cannot be read — those are reported as ``partial``,
-``unavailable`` or ``corrupt`` explicitly. Class-B consumers such as
+Touch evidence has two explicit bounds (#1290). General refreshes read the
+newest :data:`_MAX_RESULT_FILES` and report ``partial`` when more exist. Decay
+passes its time horizon and reads every result whose mtime is inside that
+window, because a rank-50 prefix cannot answer "touched within 14 days?" on a
+busy host. What blocks either path: a missing or unreadable directory, an
+unreadable/corrupt selected file, or a file whose mtime cannot be read — those
+are reported as ``partial``, ``unavailable`` or ``corrupt`` explicitly.
+Class-B consumers such as
 ``stale_artifacts`` must refuse destructive decay candidates for ``missing``,
 ``unknown``, ``partial``, ``unavailable`` or ``corrupt`` status. Operators
 must provision both expected result directories, including an explicitly
@@ -490,7 +489,11 @@ def _harness_run_signal(script: Path, state_dir: Path, selfevo_repo: Path) -> st
         return None
 
 
-def _touched_from_results_with_status(state_dir: Path) -> tuple[dict[str, str], str]:
+def _touched_from_results_with_status(
+    state_dir: Path,
+    *,
+    since: datetime | None = None,
+) -> tuple[dict[str, str], str]:
     """Read bounded touch evidence from live and archived result artifacts.
 
     The result archiver moves completed payloads from ``results/`` to the flat
@@ -534,14 +537,19 @@ def _touched_from_results_with_status(state_dir: Path) -> tuple[dict[str, str], 
         except OSError:
             return touched, "unavailable"
 
-        # The newest-_MAX_RESULT_FILES prefix IS the complete answer to "touched
-        # recently?" (#1290): files older than the whole window cannot carry a
-        # newer touch than anything inside it. Only evidence that could not be
-        # inspected makes the read partial — a missing or unreadable directory
-        # here (it cannot silently look like a complete empty horizon), or an
-        # unreadable/corrupt/undatable file inside the window, below.
-        status = "partial" if missing_dirs or inaccessible else "complete"
-        for entry in entries[:_MAX_RESULT_FILES]:
+        # A rank bound cannot answer a time-window question. Callers that pass
+        # ``since`` need every result inside that horizon; the general refresh
+        # path remains rank-bounded and reports partial when the cap truncates
+        # its unbounded question.
+        if since is not None:
+            cutoff_ts = since.timestamp()
+            selected = [entry for entry in entries if entry.stat().st_mtime >= cutoff_ts]
+            capped = False
+        else:
+            selected = entries[:_MAX_RESULT_FILES]
+            capped = len(entries) > _MAX_RESULT_FILES
+        status = "partial" if capped or missing_dirs or inaccessible else "complete"
+        for entry in selected:
             try:
                 data = json.loads(entry.read_text(encoding="utf-8"))
             except PermissionError:
@@ -1382,7 +1390,9 @@ def stale_artifacts(
         # A decay decision must not trust a six-hour refresh watermark: the
         # result/archive evidence can change independently of the repo HEAD.
         # Re-read status and merge only the fresh bounded touch map here.
-        fresh_touched, fresh_status = _touched_from_results_with_status(Path(state_dir))
+        fresh_touched, fresh_status = _touched_from_results_with_status(
+            Path(state_dir), since=cutoff,
+        )
         persisted_status = str(usage_data.get("touched_results_status") or "unknown")
         touch_status = fresh_status if fresh_status != "valid-empty" else persisted_status
         if fresh_status not in {"complete", "valid-empty"}:
--- a/nanobot/runtime/usage_evidence.py
+++ b/nanobot/runtime/usage_evidence.py
@@ -31,13 +31,19 @@ so the three signals are all things the harness can observe on disk itself:
 Touch evidence scans the union of ``subagents/results/`` and the flat
 ``subagents/archive/`` newest-first, with a deterministic ``(mtime, directory,
 name)`` descending order and a shared bound of :data:`_MAX_RESULT_FILES`.
-A bounded or incomplete read is reported as ``partial`` (rather than silently
-being treated as complete); missing/unreadable/corrupt input is reported
-explicitly. Class-B consumers such as ``stale_artifacts`` must refuse
- destructive decay candidates for ``missing``, ``unknown``, ``partial``,
-``unavailable`` or ``corrupt`` status. Operators must provision both expected
-result directories, including an explicitly empty archive, before a complete
-empty horizon can be claimed.
+The bound alone does not make the read ``partial`` (#1290): touch is a
+newest-wins signal and decay asks "touched recently?", so the newest
+:data:`_MAX_RESULT_FILES` files answer it completely — a script touched only
+in the unread older tail is by definition not recently touched. Any live host
+holds thousands of archived results, so a cap that meant ``partial`` was a
+permanent off-switch for the decay lane. What DOES block: a missing or
+unreadable directory, an unreadable/corrupt file inside the window, or a file
+whose mtime cannot be read — those are reported as ``partial``,
+``unavailable`` or ``corrupt`` explicitly. Class-B consumers such as
+``stale_artifacts`` must refuse destructive decay candidates for ``missing``,
+``unknown``, ``partial``, ``unavailable`` or ``corrupt`` status. Operators
+must provision both expected result directories, including an explicitly
+empty archive, before a complete empty horizon can be claimed.
 
 systemd/cron execution traces are NOT reachable from the state dir — they
 are deliberately skipped, never faked.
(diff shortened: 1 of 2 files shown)


ozand:
**Verdict: APPROVE IN SUBSTANCE — no blocking findings.**

I reviewed the full diff against the fail-closed guard risks.

- The distinction remains exact: only the cap is removed from the `partial` calculation. Missing expected directories and inaccessible directories still produce `partial`; unreadable/corrupt/undatable files inside the selected window still produce blocking `unavailable`/`corrupt`/`partial` statuses. `_TOUCH_EVIDENCE_BLOCKING_STATUSES` is unchanged, so the guard remains fail-closed for those conditions.
- Tests cover the non-cap paths explicitly: missing result directory remains `partial` and blocks decay; a corrupt newest file produces `corrupt` and blocks decay. Existing blocking tests for missing evidence, unknown sidecar status, malformed result, and refresh behavior remain unchanged. The cap-only ordering tests now assert `complete`.
- The newest-wins argument matches the implementation's actual ordering: candidates are sorted by filesystem `st_mtime`, then directory/name tie-breakers, and the newest 50 are selected. `stale_artifacts` asks whether a file was touched recently, so an artifact outside that newest-mtime window cannot be a newer touch signal than one inside it. The PR's production evidence also shows the host archive is much larger than the cap, which is why the old cap-induced `partial` was a permanent off-switch.
- I found no conversion of all non-complete states into success or any laundering of corrupt/missing evidence.

Focused test evidence in the PR: 98 passed. The PR correctly reports the current Windows baseline as five failures until #1295 merges; the fifth is the pre-existing Windows path-separator test from #1275, not this change.

GitHub does not permit a formal approval from the authenticated PR author, so this is the requested comment verdict rather than an APPROVE review.

ozand:
VERDICT: CHANGES REQUESTED — the status distinction remains fail-closed, but the newest-50 premise is false for the decay question.

## [BLOCKER] Rank-bounded newest 50 is not a complete time-bounded touch horizon

`_touched_from_results_with_status()` sorts result artifacts by filesystem mtime and reads only the newest 50. `stale_artifacts()` asks whether each script was touched inside `older_than_days` (normally 14 days). Those are different windows.

A script can be present only in result #51 while all 51 results are less than 14 days old. With more than 50 cycles in a day on this host, that is routine, not hypothetical. The patch reports this read as `complete`, drops the script's recent touch, and can emit a destructive decay candidate for a script touched yesterday. Being older than 50 newer artifacts does **not** mean being older than the 14-day touch cutoff.

The new positive test avoids this case by deliberately making all 60 archived results 20+ days old. It proves only the safe special case, then generalizes it to every capped read. Add the adversarial case:

1. create 51+ readable artifacts all inside the decay horizon;
2. put `scripts/old_tool.py` only in the 51st-newest artifact;
3. ensure its git/usage evidence would otherwise make it decay-eligible;
4. assert it is not returned as a decay candidate.

That test should fail on this branch.

A safe implementation must make the bound time-aware: for example, continue reading through all artifacts whose mtime is inside the relevant decay horizon, or keep the capped status blocking unless the oldest inspected artifact is already older than the cutoff. The current reader has no cutoff parameter, so it cannot establish completeness for the caller's time question.

## Status-path audit

The patch did preserve the intended fail-closed distinctions outside the cap-only case:

- absent `subagents/` → `missing` → blocks;
- one expected directory missing/inaccessible → `partial` → blocks;
- stat/read failure → `unavailable` → blocks;
- malformed/non-object artifact in the inspected prefix → `corrupt` → blocks;
- undatable inspected artifact → `partial` → blocks;
- unknown persisted status remains blocking.

The added missing-directory and corrupt-file tests would fail if those statuses were globally made non-blocking, and the existing status tests remain green. Focused suite: **98 passed**. CI is green on Python 3.11/3.12/3.13.

The blocker is narrower but fundamental: the one newly non-blocking condition is not actually complete for the 14-day decay decision.

ozand:
The blocker is correct and the reasoning error behind this PR is mine as much as the author's — I authorised this arm, on the newest-wins argument, without checking that "newest 50 by rank" is the same window as "touched within the decay horizon". It is not, and the review states why precisely.

Restating it so the fix targets the right thing: `_touched_from_results_with_status()` bounds by **rank** — the 50 newest artifacts by mtime. `stale_artifacts()` asks a **time** question — was this script touched inside `older_than_days`. On a host doing more than 50 cycles a day, a script can appear only in artifact #51 while all 51 are days old. Marking that read `complete` drops a real recent touch, and the output of this path is a **destructive** decay candidate. So the failure mode of the fix is worse than the failure mode of the bug: the bug switched the lane off, this would switch it on and point it at a script touched yesterday.

The test point is the one I want carried furthest, because it is the third instance of the same pattern tonight. The positive test makes all 60 archived results 20+ days old, which is precisely the case where rank and time coincide — it proves the safe special case and then the change generalises it to every capped read. A fixture that avoids the defect is not evidence about the defect. We have now hit this in #1280 (a retry test whose fixture omitted the `source_artifact` production always supplies), in #1296 (a regression test that appended the ledger row by hand across the broken gap), and here.

For whoever picks up the fix: the reader currently has no cutoff parameter, so it **cannot** establish completeness for the caller's question — that is the actual gap, not the status mapping. Either give it the horizon and let it read until the oldest inspected artifact is older than the cutoff, or keep the capped read blocking unless that condition already holds. And write the adversarial test first: 51+ readable artifacts all inside the horizon, the target script present only in the 51st-newest, asserted not to be returned as a decay candidate. It must fail on the current branch before anything is changed.

What survives from this PR: the status-path audit, and the original finding that the decay lane is permanently off. #1290 stays open. What does not survive is the premise that a rank bound answers a time question.

ozand:
## Takeover revision pushed — time-aware decay horizon

Current head: `c2c711cc`.

The prior newest-50-complete premise is removed. Decision: **restore the decay lane safely by passing its time cutoff into the touch reader and reading every result artifact inside that horizon.** The alternative arm — keep capped reads blocking until archive pruning — was not chosen because it leaves the lane permanently off on the current host.

- General refresh remains newest-50 and reports `partial` when capped.
- `stale_artifacts(older_than_days=N)` passes its exact cutoff and receives all touch evidence inside that horizon.
- `missing`, `unavailable`, `partial`, `corrupt`, and unknown persisted statuses remain blocking.

Adversarial test was written and observed failing before the fix: 51 readable artifacts all inside 14 days, target touch only in #51, persisted evidence otherwise decay-eligible. The old branch emitted the target as destructive decay demand; the new reader sees the #51 touch and excludes it.

Verification:

- Focused usage suite: **101 passed**.
- Full Windows suite: **3032 passed, 17 skipped, 4 failed**, exactly the four established bridge baseline tests.
- Ruff and `git diff --check` passed.

ozand:
VERDICT: APPROVE — the destructive rank/time bug is fixed; the horizon scan is intentionally I/O-bounded by retained state rather than by a file cap, and its current host cost is acceptable but should be stated explicitly.

## Blocker re-check

The previous newest-50-complete premise is gone in `c2c711cc`:

- `_touched_from_results_with_status(..., since=cutoff)` selects every result whose filesystem mtime is at or after the caller's exact cutoff;
- `stale_artifacts(older_than_days=N)` passes the same cutoff it later uses for `last_used` / `last_touched` eligibility;
- the no-`since` refresh path still reads newest 50 and reports `partial` when the union exceeds 50;
- `missing`, `unavailable`, `partial`, `corrupt`, and unknown remain in the blocking path.

The adversarial test has the required shape: 51 artifacts inside 14 days, with the target touch only in the 51st-newest artifact. I restored the old non-blocking rank-50 behavior locally; that test failed by emitting `scripts/old_tool.py` as a decay candidate. Restoring `c2c711cc` makes it pass. This independently confirms the regression would catch loss of the cutoff behavior.

## Time-aware read cost and bound

The decay read has **no independent file-count or byte cap**. Its practical upper bound is all retained result artifacts whose mtime falls inside the caller-provided horizon; because callers supply `older_than_days`, a 90-day caller currently selects the entire retained archive. That is a deliberate correctness-over-I/O trade, not a conventional bounded read.

Read-only measurement on the 2 GB i386 host today, across `results/` + `archive/`:

| horizon | selected files | bytes parsed | JSON parse/read time |
|---|---:|---:|---:|
| 3 days | 652 | 3,637,039 | — |
| 14 days | 2,544 | 14,186,976 | 2.12 s |
| 30 days | 3,601 | 19,401,642 | 2.94 s |
| 90 days | 3,601 | 19,401,642 | 2.96 s |

Enumerating and sorting all 3,601 files cost another **0.77 s**. Measured Python max RSS was about **11.3 MB**. Thus the normal 14-day decay scan is roughly **2.9 s total** today; 30/90 days are roughly **3.7 s total**. This is not free, but it is acceptable for the current retained state and materially safer than dropping recent touch evidence. If archive retention or cycle rate grows substantially, this cost must be revisited rather than describing the path as file-bounded.

## Divergent completeness is coherent

The two callers ask different questions:

- general refresh records a cheap newest-50 snapshot and honestly persists `partial`;
- destructive decay re-reads the exact time horizon and refuses decay on any blocking result.

`stale_artifacts` does not compare a fresh horizon map to the general refresh map. It uses the persisted status only when the fresh horizon is `valid-empty`; otherwise the fresh status governs, and fresh touches are merged monotonically by latest timestamp. I found no consumer that treats the two maps as equivalent or compares their sizes.

## Verification

- focused usage suite on the PR head: **99 passed** in this checkout (PR reports 101 under its environment/test selection);
- adversarial old-behavior demonstration: **failed as expected** with the target emitted;
- working tree restored clean after the demonstration;
- CI is green on Python 3.12 and 3.13; 3.11 was still running when reviewed.

No blocking finding remains. I recommend updating the PR description/doc wording to state the measured unbounded-by-count trade explicitly, but I do not consider that a merge blocker at the measured cost.

PR #1295, opened 2026-09-04
test(dashboard): stop pinning a POSIX separator in the queue-reference test

ozand:
One-line test fix with a disproportionate effect on how the fleet reads its own results.

`test_queue_reference_keeps_raw_internal_value_until_public_sanitization` (added with #1275) asserted:

```python
reference = DASHBOARD.format_stale_request_reference(42.0, Path("/secret/a.json"))
assert reference == ("42.0h", "/secret/a.json")
```

On Windows `str(Path("/secret/a.json"))` is `\secret\a.json`, so this failed locally and passed on the Linux CI:

```
AssertionError: assert ('42.0h', '\secret\a.json') == ('42.0h', '/secret/a.json')
```

The test's subject is that the raw path survives until `sanitize_public_metrics` redacts it — the two assertions that check that are unchanged. The separator spelling was incidental to it.

**Why this is worth its own PR rather than a drive-by.** Every agent brief in this repo quotes the Windows full-suite baseline as exactly four named failures, and treats a fifth as theirs to explain. This test silently made it five. One agent spent a round investigating a "fifth failure" before the real cause was found elsewhere, and the wrong explanation nearly went into a PR body as fact. A baseline that is stated as exact has to actually be exact, or it stops being a useful contract and becomes a source of confident wrong answers.

Verified: `tests/test_eeebot_dashboard_truth.py` is 25 passed on `main` with this change, 24 passed + 1 failed without it.

### DIFF
--- a/tests/test_eeebot_dashboard_truth.py
+++ b/tests/test_eeebot_dashboard_truth.py
@@ -362,11 +362,18 @@ def test_queue_path_redaction_preserves_action_metadata() -> None:
 
 
 def test_queue_reference_keeps_raw_internal_value_until_public_sanitization() -> None:
-    reference = DASHBOARD.format_stale_request_reference(42.0, Path("/secret/a.json"))
-    assert reference == ("42.0h", "/secret/a.json")
+    # The subject is that the raw path survives until sanitization, not how the
+    # platform spells a separator: str(Path("/secret/a.json")) is
+    # "\secret\a.json" on Windows and "/secret/a.json" elsewhere. Pinning the
+    # POSIX form made this a Windows-only failure and silently moved the
+    # documented four-failure Windows baseline to five.
+    raw_path = Path("/secret/a.json")
+    raw_text = str(raw_path)
+    reference = DASHBOARD.format_stale_request_reference(42.0, raw_path)
+    assert reference == ("42.0h", raw_text)
     sanitized = DASHBOARD.sanitize_public_metrics({
-        "oldest_stale_request_path": Path("/secret/a.json"),
-        "oldest_stale_request_path_text": "/secret/a.json",
+        "oldest_stale_request_path": raw_path,
+        "oldest_stale_request_path_text": raw_text,
         "queue_action": f"archive oldest 42.0h @ {reference[1]}",
         "queue_archive_target": f"{reference[1]} (42.0h)",
     })


ozand:
**Verdict: APPROVE.** The test still asserts its subject, the change hides no normalisation, and on Windows the file goes from 24 passed / 1 failed to 25 passed against unchanged `main` code. One low-severity note from the sweep, nothing blocking.

**Does the test still assert what it is for?** Yes. The subject has two halves and both survive. (a) Raw survives the formatter: `format_stale_request_reference` → `format_oldest_stale_request_path`, which is literally `return str(oldest_stale_request_path)` (`scripts/eeebot_dashboard.py:1003-1006`, no redaction, no shortening). Asserting `reference[1] == str(raw_path)` is not tautological — the value under test comes out of the dashboard's formatter, the expectation is computed by `pathlib` — and it would still catch the failure the test exists for: a future formatter that redacts or truncates at this layer. What it would no longer catch is a formatter that switched to `as_posix()`; that is not the subject, and the old assertion only caught it by accident of platform. (b) Redaction still happens at `sanitize_public_metrics`: `_QUEUE_ABSOLUTE_PATH_RE` (`:926`) accepts `[/\\]{1,2}` and drive-letter forms, so `\secret\a.json` is redacted on Windows exactly as `/secret/a.json` is on Linux, and the two `path-redacted` assertions hold on both. Verified empirically, not by reading: PR test file copied over `origin/main` code in a scratch worktree on Windows → `25 passed in 0.46s`.

**Hidden normalisation?** None. The formatter is `str()`; `format_stale_request_reference` composes the age text and that string, nothing else.

**Other POSIX pins in the file** (swept `Path("/`, `== "/`, `/tmp`, `/var`, `as_posix`, backslashes):
- `:71-72`, `:200`, `:469` — `Path("/tmp/...")` used as inputs / `write_snapshot` targets, never compared to a POSIX string; already pass on Windows.
- `:360-361` — the Windows-shaped inputs (`C:\\secret\\a.json`, `\\secret\\a.json`) are asserted redacted; good.
- `:205-206` and `:473-474` — negative assertions `"/secret/" not in rendered` / `"/secret/queue/" not in rendered`. On Windows the raw spelling would be `\secret\`, so these pass vacuously there; the positive `"path-redacted" in combined` at `:478` still proves the redaction, so it is a weaker Windows assertion, not a hole. If you want them platform-honest, compare against `str(Path("/secret/..."))` the same way this PR does. Not required for this PR.

After merge the Windows baseline returns to the four bridge tests; #1288 / #1293 both recorded the fifth with this cause and need no correction.

PR #1296, opened 2026-09-04
fix(runtime): complete hypothesis identity plumbing

### LINKED ISSUE #1285: Lifecycle never marks strategist durable hypotheses touched/answered: demand-id hash does not match hyp-NNNN key
## Context (self-contained)

The strategist (#999) appends HADI hypotheses to `state/hypotheses/durable.json`; the lifecycle tracker in `nanobot/runtime/hypothesis_backlog.py::reconcile` is supposed to mark a hypothesis `answered` when a cycle whose `serves` references it succeeds, which also triggers the #878 verdict (`_apply_verdict`) that evaluates the hypothesis's `insight_criterion`.

For strategist entries this never happens. Observed on the host (release `20260904T023137Z-canonical-76bab401`):

- `durable.json` entry `hyp-0022` "suppress-in-tree-proposals-for-host-metrics-staleness", created `2026-09-03T12:38:11Z`.
- Proposer prompt `2026-09-03T12:46:47Z`: `## Demand … - [hypothesis-cab86c3e9ed8] (hypothesis) suppress-in-tree-proposals-for-host-metrics-staleness`.
- Ledger: `proposed` `2026-09-03T12:47:04Z` cycle-8ef9f15c3647, `serves: "demand hypothesis-cab86c3e9ed8"`; second proposal `13:32:07Z` cycle-a9ed6bd7f8f3, `outcome: success` at `14:01:50Z`.
- `lifecycle.json`: `hyp-0022 {"status": "stale", "cycles_untouched": 50, "stale_at": "2026-09-03T19:41:42Z"}` — stale, not answered, seven hours after it was served and integrated.

Cause, in code: `reconcile` extracts the served ref `hypothesis-cab86c3e9ed8` and matches it with `_ref_matches_candidate(ref, cand)` — `ref == key or ref in key or ref in title`. The durable candidate's key is `hyp-0022` (its `hypothesis_id`) and its title is the plain text; the demand-id hash appears in neither. So `touched_keys`/`answered_keys` never include strategist entries, every one of them reaches `STALE_AFTER_UNTOUCHED_CYCLES = 50` (~7 h at current cadence), and `demand._hypothesis_items` then excludes it (`status == "active"` only). The 09-02 audit's "18 of 20 stale" was this, not non-pickup: at least two of those 20 were served and integrated.

Consequences: (1) lifecycle status for strategist hypotheses is wrong (`stale` where `answered` is true); (2) the HADI `insight_criterion` verdict is never computed for any strategist hypothesis, so the strategist gets no feedback on its own hypotheses in `prior_decisions`; (3) an entry that was served once cannot be served again — which is the intended dedup, so no harm on that axis.

## Proposed fix

Make the demand-id derivation reachable from the candidate: either store the demand id on the durable entry when `demand._hypothesis_items` mints it, or have `_ref_matches_candidate` recompute `hypothesis-<hash>` from the candidate title with the same function `demand` uses. One shared helper, no second hashing rule.

## Acceptance

- [ ] Failing test first: durable entry → `proposed` row with `serves: "demand hypothesis-<hash-of-title>"` → `outcome: success` → `reconcile` marks the entry `answered` and `_apply_verdict` runs.
- [ ] Existing backlog (`hypothesis-llm-proposer-cycle-*`) matching unchanged; existing lifecycle tests green.
- [ ] Live: after deploy, the next strategist hypothesis that gets served shows `answered` (with `verdict`) in `lifecycle.json`, quoted.

ozand:
## Summary

Fixes #1285 by carrying a unique strategist hypothesis identity through demand selection and proposal materialization, then reconciling by exact identity instead of text or generated demand id.

- `_hypothesis_items` carries the source `hypothesis_id` as `hypothesis_ref` on the demand item.
- The assigned-demand path injects that ref into the model proposal after the model response; the model does not need to echo internal bookkeeping.
- `write_request` preserves `hypothesis_ref` in both the request payload and proposed ledger event.
- `hypothesis_backlog.reconcile` uses exact `hypothesis_ref == candidate key` when present; legacy rows without it retain the old matcher.
- Newly minted hypotheses without an explicit id use a short content hash (`title + hypothesis`) rather than the old FIFO `hyp-NNNN` position, so FIFO eviction cannot reuse an identity.
- Added tests for production-path proposal plumbing, strategist-shaped reconciliation, and uniqueness across FIFO eviction.

## State-access deny-set

`nanobot/runtime/hypothesis_backlog.py` is **not** listed in `nanobot/runtime/runtime_deny.py`; this change does not add it because the module is ordinary runtime bookkeeping, not a mutation-surface module.

## Verification

- Focused tests: **416 passed** (`test_hypothesis_backlog.py`, `test_demand.py`, `test_llm_proposer.py`).
- Full suite after rebasing onto current `origin/main` (`2e9937bd`): **3024 passed, 17 skipped, 4 failed**.
- The exact four remaining failures are the established Windows bridge baseline:
- Ruff's changed-runtime checks and `git diff --check` passed.
- No host access, deployment, current switch, service restart, or merge was performed.

### DIFF
--- a/nanobot/runtime/demand.py
+++ b/nanobot/runtime/demand.py
@@ -474,7 +474,6 @@ def _make_item(
     affected_path: str = "",
     vector: str = "",
     direction: str = "",
-    hypothesis_ref: str = "",
 ) -> dict[str, str]:
     summary = (summary or "").strip()[:_MAX_SUMMARY_CHARS]
     return {
@@ -496,7 +495,6 @@ def _make_item(
         # correspondence; see its own docstring). Additive-only: existing
         # callers that omit this arg get "" and are unaffected.
         "direction": (direction or "").strip(),
-        "hypothesis_ref": (hypothesis_ref or "").strip()[:200],
     }
 
 
--- a/nanobot/runtime/demand.py
+++ b/nanobot/runtime/demand.py
@@ -474,6 +474,7 @@ def _make_item(
     affected_path: str = "",
     vector: str = "",
     direction: str = "",
+    hypothesis_ref: str = "",
 ) -> dict[str, str]:
     summary = (summary or "").strip()[:_MAX_SUMMARY_CHARS]
     return {
@@ -495,6 +496,7 @@ def _make_item(
         # correspondence; see its own docstring). Additive-only: existing
         # callers that omit this arg get "" and are unaffected.
         "direction": (direction or "").strip(),
+        "hypothesis_ref": (hypothesis_ref or "").strip()[:200],
     }
 
 
@@ -1670,7 +1672,10 @@ def _hypothesis_items(
                 continue
             seen.add(title)
             evidence = str(entry.get("evidence") or entry.get("metric") or entry.get("data_to_collect") or entry.get("insight_criterion") or entry.get("acceptance") or "")
-            items.append(_make_item("hypothesis", title, evidence))
+            hypothesis_ref = str(entry.get("hypothesis_id") or "").strip()
+            demand_item = _make_item("hypothesis", title, evidence)
+            demand_item["hypothesis_ref"] = hypothesis_ref
+            items.append(demand_item)
         # #1219: ``research/hypotheses.json`` is no longer a source here. Its
         # writer (``cycle_planning._write_research_feed``) was deleted with the
         # planner (#924) and the file froze on 2026-08-22; durable + backlog
--- a/nanobot/runtime/hypothesis_backlog.py
+++ b/nanobot/runtime/hypothesis_backlog.py
@@ -668,6 +668,8 @@ def append_hypotheses(state_dir: Path | str, new_entries: list[dict[str, Any]]) 
         entries = []
 
     # Map existing titles / keys to avoid duplicate additions
+    import hashlib
+
     existing_titles = {
         str(e.get("title") or e.get("task_title") or e.get("hypothesis") or "").strip().lower()
         for e in entries
@@ -687,8 +689,13 @@ def append_hypotheses(state_dir: Path | str, new_entries: list[dict[str, Any]]) 
         if lookup_key in existing_titles:
             continue
 
+        hypothesis_id = str(item.get("hypothesis_id") or "").strip()
+        if not hypothesis_id:
+            hypothesis_id = "hyp-" + hashlib.sha256(
+                f"{title}\n{hypothesis_text}".encode("utf-8")
+            ).hexdigest()[:16]
         entry_record: dict[str, Any] = {
-            "hypothesis_id": str(item.get("hypothesis_id") or f"hyp-{len(entries) + 1:04d}"),
+            "hypothesis_id": hypothesis_id,
             "task_title": title or hypothesis_text[:80],
             "title": title or hypothesis_text[:80],
             "hypothesis": hypothesis_text,
--- a/nanobot/runtime/hypothesis_backlog.py
+++ b/nanobot/runtime/hypothesis_backlog.py
@@ -358,10 +358,15 @@ def reconcile(state_dir: Path, *, now: datetime | None = None) -> None:
         answered_keys: dict[str, str] = {}
         for cid, prow in proposed_by_cycle.items():
             ref = _serves_hypothesis_ref(str(prow.get("serves") or ""))
-            if not ref:
+            explicit_ref = str(prow.get("hypothesis_ref") or "").strip()
+            if not ref and not explicit_ref:
                 continue
             for cand in candidates:
-                if not _ref_matches_candidate(ref, cand):
+                if explicit_ref:
+                    matches = explicit_ref == cand["key"]
+                else:
+                    matches = _ref_matches_candidate(ref, cand)
+                if not matches:
                     continue
                 touched_keys.add(cand["key"])
                 outcome_row = outcome_by_cycle.get(cid)
--- a/nanobot/runtime/hypothesis_backlog.py
+++ b/nanobot/runtime/hypothesis_backlog.py
@@ -55,6 +55,7 @@ verification gate).
 """
 from __future__ import annotations
 
+import hashlib
 import json
 import os
 import re
@@ -667,9 +668,29 @@ def append_hypotheses(state_dir: Path | str, new_entries: list[dict[str, Any]]) 
         backlog_data = {"schema": "hypothesis-durable-v1", "entries": []}
         entries = []
 
-    # Map existing titles / keys to avoid duplicate additions
-    import hashlib
+    # Repair duplicate FIFO-era ids before exact identity links can act on
+    # them. Unique legacy ids stay unchanged; each duplicate receives the
+    # same content identity used for new entries. Lifecycle rows under the old
+    # id remain historical rather than being guessed onto one duplicate.
+    id_counts: dict[str, int] = {}
+    for entry in entries:
+        if isinstance(entry, dict):
+            hypothesis_id = str(entry.get("hypothesis_id") or "").strip()
+            if hypothesis_id:
+                id_counts[hypothesis_id] = id_counts.get(hypothesis_id, 0) + 1
+    for entry in entries:
+        if not isinstance(entry, dict):
+            continue
+        hypothesis_id = str(entry.get("hypothesis_id") or "").strip()
+        if id_counts.get(hypothesis_id, 0) <= 1:
+            continue
+        title = str(entry.get("title") or entry.get("task_title") or "").strip()
+        hypothesis_text = str(entry.get("hypothesis") or "").strip()
+        entry["hypothesis_id"] = "hyp-" + hashlib.sha256(
+            f"{title}\n{hypothesis_text}".encode("utf-8")
+        ).hexdigest()[:16]
 
+    # Map existing titles / keys to avoid duplicate additions
     existing_titles = {
         str(e.get("title") or e.get("task_title") or e.get("hypothesis") or "").strip().lower()
         for e in entries
--- a/nanobot/runtime/llm_proposer.py
+++ b/nanobot/runtime/llm_proposer.py
@@ -2585,6 +2585,7 @@ def write_request(
     rationale = str(proposal.get("rationale") or "").strip()
     target_path = str(proposal.get("target_path") or "").strip()
     serves = str(proposal.get("serves") or "").strip()
+    hypothesis_ref = str(proposal.get("hypothesis_ref") or "").strip()
 
     improvements_dir = state_dir / "improvements"
     improvements_dir.mkdir(parents=True, exist_ok=True)
@@ -2661,6 +2662,7 @@ def write_request(
         "target_path": target_path,
         "serves": serves,
         "source_artifact": "llm_proposer",
+        **({"hypothesis_ref": hypothesis_ref} if hypothesis_ref else {}),
     }
     # #760: demand traceability — when serves is 'demand <id>', the proposed
     # row also carries the id, so proposal→demand-item is queryable.
--- a/nanobot/runtime/llm_proposer.py
+++ b/nanobot/runtime/llm_proposer.py
@@ -2847,6 +2847,18 @@ def maybe_propose(state_dir: Path, selfevo_repo: Path | None) -> str | None:
         def _proposal_demand_id(p: Any) -> str:
             return _demand_id_from_serves(p.get("serves")) if isinstance(p, dict) else ""
 
+        def _stamp_assigned_hypothesis_ref(p: Any) -> None:
+            if not assigned or not demand_items or not isinstance(demand_items[0], dict):
+                return
+            if not isinstance(p, dict):
+                return
+            assigned_item = demand_items[0]
+            if _proposal_demand_id(p) != str(assigned_item.get("id") or "").strip():
+                return
+            hypothesis_ref = str(assigned_item.get("hypothesis_ref") or "").strip()
+            if hypothesis_ref:
+                p["hypothesis_ref"] = hypothesis_ref
+
         def _is_noop_reply(p: Any) -> bool:
             # #760 roll-out fix: the weak host model emits no_valuable_task
             # as the string "true" (or 1) rather than a JSON boolean; such a
@@ -2869,8 +2881,7 @@ def maybe_propose(state_dir: Path, selfevo_repo: Path | None) -> str | None:
 
         proposal = _call_propose()
         calls_made += 1
-        if assigned and demand_items and isinstance(demand_items[0], dict) and isinstance(proposal, dict):
-            proposal["hypothesis_ref"] = str(demand_items[0].get("hypothesis_ref") or "").strip()
+        _stamp_assigned_hypothesis_ref(proposal)
 
         if allow_no_op and _is_noop_reply(proposal):
             _record_noop_skip(state_dir, _noop_skip_reason(str(proposal.get("reason") or "")))
@@ -2880,8 +2891,7 @@ def maybe_propose(state_dir: Path, selfevo_repo: Path | None) -> str | None:
         if not ok and calls_made < _MAX_LLM_CALLS:
             proposal = _call_propose(rejection_reason=reason)
             calls_made += 1
-            if assigned and demand_items and isinstance(demand_items[0], dict) and isinstance(proposal, dict):
-                proposal["hypothesis_ref"] = str(demand_items[0].get("hypothesis_ref") or "").strip()
+            _stamp_assigned_hypothesis_ref(proposal)
             if allow_no_op and _is_noop_reply(proposal):
                 _record_noop_skip(state_dir, _noop_skip_reason(str(proposal.get("reason") or "")))
                 return None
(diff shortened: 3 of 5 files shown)


ozand:
**Verdict: REQUEST CHANGES — the explicit link is never populated on the production path, so the fix is inert live; and the identifier it links on is not unique.** The reconcile side is exactly what the constraint asked for (exact key equality, no substring widening); the mint→row plumbing has a gap in the middle, and the test writes the ledger row by hand across precisely that gap.

### 1. `hypothesis_ref` is read from `proposal`, but nothing ever puts it there

- `demand._hypothesis_items` stamps `hypothesis_ref` on the **demand item** (`demand.py:1675-1678`). Good — that is mint time.
- `llm_proposer.write_request` reads `proposal.get("hypothesis_ref")` (`:2588`) and writes it into `proposed_event` (`:2665`).
- `proposal` is the model's JSON: `_call_propose()` → `propose()`; `serves` is whatever the model returned (validated by `_validate_serves`, `demand_id` then derived from it). At `bb9eb84e`, `hypothesis_ref` occurs in `llm_proposer.py` **only** at lines 2588 and 2665, both inside `write_request`. There is no `proposal[...] =` / `update()` from `demand_items[0]` anywhere (grep for `demand_items[0]` finds only `assigned_id` reads at `:1560`, `:2826`, `:2864`), and the prompt renders `id / summary / evidence / affected_path` (`:1057-1085`), not `hypothesis_ref`, so the model cannot echo it either.
- Result on the host: `hypothesis_ref` is always `""`, the `**({...} if hypothesis_ref else {})` spread adds nothing to the row, `reconcile` sees `explicit_ref == ""` and takes the old `_ref_matches_candidate` branch — the exact behaviour #1285 describes. Strategist entries keep going `stale`.

The regression test proves the reconcile half only: it calls `cycle_ledger.append_event(..., "hypothesis_ref": "hyp-0022")` directly. I confirmed it fails on `origin/main` (`1 failed, 37 passed`) and passes on the branch (`414 passed` across the three focused files), so it is a real test of `reconcile` — but it does not exercise the path that is broken in production.

**Fix, one place:** after `_call_propose()` returns a dict (both calls, ~`:2869` and the sizing re-call), when `assigned and demand_items and isinstance(demand_items[0], dict)`, set `proposal["hypothesis_ref"] = str(demand_items[0].get("hypothesis_ref") or "")` — the same block that already derives `assigned_id` from `demand_items[0]` (`:2863-2864`) is the right home, since that is where the cycle is known to be pinned to one demand item. Then `write_request` works as written. Add the test that fails today: an assigned hypothesis demand item with `hypothesis_ref: "hyp-0022"` → the `proposed` ledger row carries `hypothesis_ref == "hyp-0022"` (drive it through the proposer with a stubbed LLM, not by appending the row).

### 2. `hypothesis_id` is not unique, so exact-key linking joins the wrong entry

`append_hypotheses` mints `hypothesis_id = f"hyp-{len(entries) + 1:04d}"` (`hypothesis_backlog.py:~686`) while `durable.json` is a FIFO capped at `DURABLE_MAX_ENTRIES = 20`. Once the file is full, every run re-mints the same numbers. Live `durable.json` right now (20 entries):

```
duplicate ids: {'hyp-0021': 3, 'hyp-0022': 2}
  hyp-0021 2026-09-02T00:00:28 | Pre-Proposer Suppression of Plateaued AST Refactors
  hyp-0021 2026-09-03T12:38:11 | exercise-unconfirmed-scripts-via-test-harness
  hyp-0022 2026-09-03T12:38:11 | suppress-in-tree-proposals-for-host-metrics-staleness
  hyp-0021 2026-09-04T00:00:47 | Pre-filtering exhausted demand IDs in candidate selection ...
  hyp-0022 2026-09-04T00:00:47 | Turn-bounded direct-edit gate dissolves inspection-versus-completion ...
lifecycle hyp-0021 -> stale | Pre-Proposer Suppression of Plateaued AST Refactors
lifecycle hyp-0022 -> stale | suppress-in-tree-proposals-for-host-metrics-staleness
```

`_candidate_key` uses `hypothesis_id`, so `lifecycle.json` already collapses three distinct hypotheses onto `hyp-0021`. With this PR, a `proposed` row carrying `hypothesis_ref: hyp-0022` would mark **whichever entry currently holds that id** `answered` — today that is the 09-04 "Turn-bounded direct-edit gate" entry, not the 09-03 host-metrics one that was actually served. Pre-existing defect, but this PR is the first thing that acts on the id as an identity, so it has to be fixed here or the link is wrong by construction. Smallest fix: mint from a monotonic counter persisted in `durable.json` (e.g. `next_id`), or suffix the id with a short hash of the title; and a test that appends 21+ entries and asserts all ids distinct.

### 3. Entries that predate the fix

Rows without `hypothesis_ref` fall back to the existing matcher — no new heuristic, behaviour unchanged, and the PR body states it ("Legacy requests without the explicit field retain the existing text matcher"). That is the right choice and it is stated. Consequence worth one line in the body: every strategist entry served before deploy stays `stale` for good (the 20 durable entries today are already 18/20 stale, so nothing is lost).

### 4. Smaller

- PR body: "`write_request` preserves `hypothesis_ref` in the request payload and proposed ledger event" — only the ledger event got it; the request JSON `payload` dict (`:2600-2655`) does not. Either add it or drop the claim.
- `_make_item` gained a `hypothesis_ref` kwarg that the only caller ignores (it sets the key afterwards). Pass it through the kwarg or drop the kwarg.
- `reconcile`: when `explicit_ref` is present it is authoritative and exact, with no fallback to text — correct, and exactly the constraint. Keep that shape.

### Verified

- New test fails on `main` (1/38), passes on branch; focused `test_hypothesis_backlog.py test_demand.py test_llm_proposer.py` → 414 passed at `bb9eb84e` on Windows.
- Live `durable.json`: all 20 entries carry `hypothesis_id`, so once item 1 is plumbed the field will be populated — but see item 2 for what it is populated with.
- Host read-only throughout.

ozand:
**Verdict: REQUEST CHANGES (small) — both blocking defects from the first review are closed on the production path; what remains is that the seven live entries already sharing an id are neither handled nor mentioned, and the ref is stamped without checking the model served the assigned item.** Everything else is right and I would merge once those two are addressed or explicitly declined in the body.

### Closed: `hypothesis_ref` now travels the real path

`maybe_propose` stamps `proposal["hypothesis_ref"]` from `demand_items[0]` after every `_call_propose()` return under assignment (`llm_proposer.py:2872, 2883, 2921`), `write_request` carries it into both the request payload (`:2652`) and the `proposed` row (`:2666`), and `reconcile` joins by exact `explicit_ref == cand["key"]`. `test_assigned_hypothesis_ref_reaches_proposed_ledger` drives `maybe_propose` with a stubbed `propose` and reads the ledger row back — the path that was missing, not a hand-appended row. Verified: the three new tests fail on `origin/main` (`3 failed, 231 deselected`) and the focused files pass at `06616a08` on Windows (`416 passed`). The body's Windows baseline of four after the `2e9937bd` rebase is correct.

### Closed: new ids are unique, and the reason is stated

`hyp-<sha256(title\nhypothesis)[:16]>` for entries that arrive without an id; `test_generated_hypothesis_ids_remain_unique_after_fifo_eviction` appends 22 and gets 20 distinct. Body says why ("so FIFO eviction cannot reuse an identity"). Run against a copy of the live `durable.json`, two strategist-shaped appends mint `hyp-1df508921dd943bc` and `hyp-1fb8fe5795f6a27d` and evict `hyp-0008`/`hyp-0009` — correct.

### Open 1: the entries that already share an id

The live file has grown since my first review: **`hyp-0021` ×4, `hyp-0022` ×3** (the 03:00 MSK run on 09-05 minted two more under the old rule). The PR leaves them as they are and the body does not mention them. What that does, simulated on the live copy with this branch's code:

```
proposed row hypothesis_ref=hyp-0022 + outcome success → reconcile
lifecycle hyp-0022 -> status answered, title 'suppress-in-tree-proposals-for-host-metrics-staleness' (the 09-03 entry)
durable entries carrying hyp-0022:
  2026-09-03  suppress-in-tree-proposals-for-host-metrics-staleness
  2026-09-04  Turn-bounded direct-edit gate dissolves inspection-versus-completion …
  2026-09-05  Pre-seed standard library module cache to dissolve …
```

`demand._hypothesis_items` reads lifecycle status by `hypothesis_id`, so one served cycle marks all three `answered` and drops them from demand; the 09-04 and 09-05 hypotheses are never offered. That is exactly "silently re-matched" for the seven entries, for roughly ten days until FIFO eviction clears them. Two acceptable resolutions, either is fine, but one must be chosen and written down:

- **Handle:** in `append_hypotheses` (or a one-shot on load), re-mint any `hypothesis_id` that is shared by more than one entry using the same content hash, and leave unique legacy ids alone. Lifecycle keys for the re-minted entries start fresh; the old `hyp-0021/0022` lifecycle rows stay as history. ~10 lines plus a test seeded with two entries sharing `hyp-0022`.
- **Decline, stated:** add to the body that the seven pre-existing duplicates are accepted as mis-linkable until they age out of the 20-entry FIFO (~2026-09-15 at the current 2/day), with the interference above spelled out. That is a legitimate call for a ten-day window; it just has to be a call.

### Open 2: the ref is stamped whether or not the model served the assigned item

`proposal["hypothesis_ref"]` is set whenever `assigned` is true. Nothing checks that the model's `serves` actually names the assigned demand — `_validate_serves` only checks the prefix, and there is no comparison of `_proposal_demand_id(proposal)` against `demand_items[0]["id"]` anywhere in the file (grep). A model that answers an assignment with `serves: "demand priority-…"` or another hypothesis id produces a `proposed` row that still carries the assigned hypothesis's ref, and its success then marks that hypothesis `answered` by a cycle that did not serve it. One-line guard at each stamping site (or in a small helper, since the block is now copied three times): stamp only when `_proposal_demand_id(proposal) == str(demand_items[0].get("id") or "")`. Add the negative test: assigned item + proposal serving a different id → row has no `hypothesis_ref`.

### Nits

- The three identical stamping blocks want a `_stamp_hypothesis_ref(proposal)` closure next to `_proposal_demand_id`.
- `import hashlib` inside `append_hypotheses` — move to module imports.
- `_make_item` no longer takes `hypothesis_ref`; the item is patched after construction. Fine, just noting the earlier kwarg is gone.

### Not exercised here

No host run; the "handled or stated" question above is decided from the live file copied read-only at 2026-09-05 ~03:20 MSK.

ozand:
## Remaining re-check items addressed

Current head: `0142f258`.

1. **Pre-existing duplicate IDs:** handled, not deferred. On each `append_hypotheses` pass, only IDs duplicated in the current durable FIFO are reminted using the same content hash as new entries. Unique legacy IDs remain unchanged. Old lifecycle records under the shared ID remain historical rather than being guessed onto one of several duplicate hypotheses. A regression seeds two different entries with `hyp-0022` and asserts both are reminted to distinct IDs.
2. **Truthful assignment link:** `hypothesis_ref` is now stamped only when `_proposal_demand_id(proposal)` exactly equals the assigned item's demand ID. A model proposal serving a different demand receives no ref. The repeated stamping blocks are consolidated into one helper. A negative production-path test asserts a wrong-`serves` proposal writes no `hypothesis_ref`.

Focused verification: **418 passed**. Runtime Ruff and `git diff --check` passed.

ozand:
**Verdict: REQUEST CHANGES (one item) — remint is correct and idempotent, the serves guard is right, but "leave the old lifecycle rows historical" does not only strand history: it resurrects already-served hypotheses as active candidates, because the reminted entry has no lifecycle row and no row means `active`.** The rows carry a `title`, so the fix is exact, not a guess. Everything below is measured on a copy of the live `durable.json` / `lifecycle.json` / `backlog.json` (2026-09-05 ~03:20 MSK) with this branch's code at `0142f258`; focused files `418 passed` on Windows.

### Closed from the previous round

- **Remint** rewrites only ids duplicated in the current FIFO; unique legacy ids (`hyp-0009` … `hyp-0020`) are untouched. Live shape: `hyp-0021` ×4 and `hyp-0022` ×3 became seven distinct `hyp-<sha256[:16]>` ids.
- **Idempotent, no churn.** A second production-shaped call (strategist entries, all deduplicated → `appended 0`) leaves the file byte-identical. Only the empty-list call (`append_hypotheses(state, [])`, the test's shape) rewrites `updated_at`; production has a single caller (`strategist._apply`) and never passes `[]`. Fine as is; worth one line in the docstring.
- **Serves guard**: `_stamp_assigned_hypothesis_ref` stamps only when `_proposal_demand_id(p) == assigned id`; negative test drives `maybe_propose` with `serves: "demand priority-other"` and the row has no `hypothesis_ref`. Right.

### Open: orphaned lifecycle rows resurrect served hypotheses

What the code does with rows whose key no longer exists in `durable.json`: `reconcile` iterates candidates only, so an orphan row is never touched again — not re-evaluated, not aged, never deleted. That part is "ignored cleanly". The problem is the other side of the remint: the entry's **new** key has no row, and `top_candidates` / `demand._is_active` treat no-row as `active`.

Measured, after one remint + `reconcile`:

```
lifecycle rows 108 → 116 ; orphan rows 93 → 96 (hyp-0021, hyp-0022 both left 'stale' — historical, as intended)
lifecycle_counts.active 75 → 83   (+8: every reminted entry gets a fresh 'active' row)
top_candidates now offers again:
  'Pre-Proposer Suppression of Plateaued AST Refactors'          — served 2026-09-02 as hypothesis-1c61690d4b8e, cycle success
  'suppress-in-tree-proposals-for-host-metrics-staleness'        — served 2026-09-03 as hypothesis-cab86c3e9ed8, integrated 14:01Z
```

So after deploy, the first strategist run remints, and the two strategist hypotheses that were actually executed re-enter the hypothesis lane as fresh active candidates. The proposer's title-based duplicate guards may or may not stop the re-proposal; the lifecycle, which is the mechanism this PR exists to make truthful, says they were never served.

**Fix (exact, small):** lifecycle rows store `title`. When reminting an entry, if a lifecycle row exists under the old id **and** its `title` equals this entry's title, move that row to the new key (keep `status`, `answered_evidence`, `stale_at`, `verdict*`); if the old row's title matches none of the duplicates, leave it historical as you do now. That is a match on the row's own recorded identity, not a guess among candidates — on the live file it moves `hyp-0022`'s row (title `suppress-in-tree-proposals-for-host-metrics-staleness`) to the reminted host-metrics entry and `hyp-0021`'s row to the Plateaued-AST entry, and leaves nothing ambiguous. Test: two entries sharing `hyp-0022`, lifecycle row `hyp-0022` with `status: answered, title: <first entry's title>` → after `append_hypotheses`, the first entry's new key is `answered`, the second's has no row / is `active`, and `top_candidates` does not list the first.

### Pre-existing, not this PR, but you asked whether anything counts orphans

`lifecycle_counts` counts every row regardless of whether its key is still a candidate. On the live file **75 of the 108 rows are orphans with `status: active`** — keys from FIFO-evicted durable entries and regenerated `backlog.json` snapshots that were never touched again after their candidate vanished. `scorecard._control_plane_snapshot` reads that as `active: 75` (visibility only, never a target). The remint adds +8 to it on day one and, without the title migration, more each time a served entry is re-minted and re-served. Worth its own small issue: either `reconcile` should retire rows whose key is no longer a candidate (`status: orphaned` or drop), or `lifecycle_counts` should count candidates only.

### Nits

- `hypothesis_ref` on the demand item is set after `_make_item`; fine.
- The PR body should state the remint's effect on `lifecycle_counts` after deploy, so the +8 jump in `control_plane.hypotheses.active` is not read as a regression.

ozand:
## FINAL RE-CHECK VERDICT: APPROVE — ready to merge

Re-checked head `0142f258` on the three requested axes.

### 1. Legacy lifecycle orphans are inert for decisions

After duplicate durable IDs are reminted, the old lifecycle entry remains under the now-absent shared ID. It does not match any current candidate key, so it cannot suppress, answer, stale, or resurface a current hypothesis through `top_candidates`, `demand._hypothesis_items`, or reconciliation.

`lifecycle_counts()` does still count qualifying status/verdict fields from orphan entries, but that snapshot is explicitly visibility-only and is not fed into fitness, targets, or gaps. In the concrete stale-orphan case, `stale` is not one of the statuses counted by `lifecycle_counts`, so it does not inflate even the reported active/answered totals. The historical row remains readable without affecting current candidates.

### 2. Remint is idempotent

I ran two `append_hypotheses(state, [])` passes over a durable file containing two different entries sharing `hyp-0022`:

```text
first pass ids  = ['hyp-afe581022a7a37ee', 'hyp-b1dc6c35e646dd67']
second pass ids = ['hyp-afe581022a7a37ee', 'hyp-b1dc6c35e646dd67']
stable = True
```

After the first pass each ID count is one, so the second pass skips reminting. The old stale lifecycle row stayed historical, while both newly keyed candidates surfaced normally.

### 3. Wrong-`serves` negative test is non-vacuous

I temporarily removed only the `_proposal_demand_id(proposal) == assigned_id` guard and ran:

```text
tests/test_llm_proposer.py::TestDemandDrivenMode::test_assigned_hypothesis_ref_not_stamped_for_different_serves
```

Observed failure:

```text
AssertionError: assert 'hypothesis_ref' not in {
  'demand_id': 'priority-other',
  'hypothesis_ref': 'hyp-0022',
  ...
}
```

The source was restored after the mutation probe. This directly proves the negative test fails if the serves-equality check is removed; it is not relying on an unrelated dedup test.

No blocking findings remain. **Approve.**

PR #1298, opened 2026-09-05
fix(reflector): record parse failures and llm telemetry

### LINKED ISSUE #1291: Reflector: >50% of attempts/day fail as 'malformed reflector output' since 08-29, and its LLM calls are never recorded (unit ReadWritePaths excludes state/llm_calls)
## Context (self-contained)

The reflector (`nanobot/runtime/reflector.py`, `eeebot-reflector.timer`, every 30 min, #1007) asks the model for a strict-JSON reflection per completed cycle and journals one row per attempt to `state/reflector/reflections.jsonl` (archives under `reflector/archive/`). An attempt whose output fails `_parse_output` is journaled as `status: error, error: "ValueError: malformed reflector output"` with `summary: "Reflector error; cycle will be retried."` — and retried on the next run, so a cycle can burn several calls.

Found during the post-deploy verification of release `20260904T212321Z-canonical-632bffbb` (2026-09-05). `reflector.py` is untouched by that release; this is not a regression of it.

## Evidence (host, read-only, 2026-09-05 01:00 MSK)

Journal rows by day, all files (`reflections.jsonl` + `archive/*.gz`):

```
2026-08-29  error 37   ok  3   consumed 44  skipped_pruned  8
2026-08-30  error 68   ok 52   consumed 11  skipped_pruned 10
2026-08-31  error 52   ok 51                skipped_pruned 12
2026-09-01  error 69   ok 31                skipped_pruned  2
2026-09-02  error 70   ok 57                skipped_pruned 17
2026-09-03  error 83   ok 53                skipped_pruned  8
2026-09-04  error 53   ok 29                skipped_pruned  2   (to 21:55Z)
```

More than half of every day's attempts since 08-29 end in error. Error strings over the same period:

```
425  ValueError: malformed reflector output
  3  RateLimitError: Error code: 429 ...
  3  InternalServerError: Error code: 500 ...
  1  APITimeoutError: Request timed out.
```

So 425 of 432 errors are the model's output failing validation, not transport. The unit summary line shows it too: `23:53:20 MSK {"candidates":3,"processed":3,"errors":0}` → `00:23:21 {"candidates":2,"processed":1,"errors":1,"consecutive_errors":1}` → `00:53:58 {"candidates":3,"processed":0,"errors":3,"consecutive_errors":3}`. Failing cycle ids on the last two runs: `cycle-1387465b520e` (21:23:21Z and again 21:53:32Z), `cycle-09fc14d55cf5` (21:53:47Z), `cycle-85a26dea5953` (21:53:58Z). Nothing surfaces this: the summary line goes to the journal only, the scorecard has no reflector dimension, and `consecutive_errors` resets on any success.

## One shape or several? Cannot be told from the record — and why

`_parse_output` returns `None` (→ the single `malformed` string) for at least five distinct causes: non-JSON text, fenced JSON, `cycle_id` mismatch, a finding/recommendation `kind` outside `_VALID_FINDINGS`/`_VALID_RECOMMENDATIONS`, or a missing `summary`/`findings`/`recommendations`. The journal row keeps only the exception string, not the response.

The response should be in `state/llm_calls/` — `_default_llm` calls `record_llm_call` and `record_llm_prompt` under `call_context(cycle_id, "reflector")`. It is not: **zero rows with `component: reflector` in any `llm_calls/*.jsonl` since 2026-07-07** (today's file has `bridge 851, proposer 116, strategist 1, curator 1`). Cause, verified on the unit:

```
eeebot-reflector.service   ProtectSystem=strict   ReadWritePaths=/var/lib/eeepc-agent/self-evolving-agent/state/reflector
eeebot-strategist.service  ProtectSystem=strict   ReadWritePaths=/var/lib/eeepc-agent/self-evolving-agent/state
```

The reflector may write only under `state/reflector`; the telemetry append to `state/llm_calls/<day>.jsonl` hits a read-only filesystem and `record_llm_call` swallows it (`except Exception: pass`, by design). Every reflector LLM call for two months has gone unrecorded — cost, tokens, finish_reason and the response text alike — which is also why the 2026-09-02 #999 audit could not find reflector prompts. Same class as [unobservable-guard]: a writer whose failure leaves no trace reads identically to one that never ran.

## What settles it

1. Widen the unit: `ReadWritePaths=/var/lib/eeepc-agent/self-evolving-agent/state` (as the strategist unit already has), in `host/eeepc/systemd/eeebot-reflector.service`. After one run the `llm_calls/prompts/<day>.jsonl` records carry `finish_reason` and the response head for each attempt.
2. Make the journal row self-sufficient regardless: record *which* `_parse_output` check failed (a short reason string, e.g. `not_json`, `fenced`, `cycle_id_mismatch`, `bad_kind:<kind>`, `missing_key:<key>`) and the first ~200 chars of the response in the error row. That turns "malformed" into a countable shape.

With either in place, one day of data answers the question: one dominant reason (e.g. `finish_reason: length` at `max_tokens=1600`, or a recurring invalid `kind`) is a prompt/parser fix; a spread across reasons is model quality and a different issue. Do not diagnose past that; the current record cannot support more.

## Acceptance

- [ ] Unit file grants write access to `state` (or at least `state/llm_calls`); after deploy, `grep -c '"component": "reflector"' state/llm_calls/<today>.jsonl` > 0, quoted with the row.
- [ ] Error rows carry a failure reason and response head; test per reason.
- [ ] One day of post-fix rows quoted: distribution of reasons and `finish_reason`, and the follow-up issue (prompt/parser vs model) named from it.

ozand:
## Summary

Fixes #1291. Reflector telemetry can write to the state tree with existing systemd hardening preserved. Complete JSON fences are repaired and counted as `fenced_json`; malformed response shapes and schema causes now have distinct reasons. Error rows contain structured `parse_reason`, `response_chars`, bounded `response_head` (200) and `response_tail` (80). Transport errors use `not_attempted` with empty response diagnostics. Plain flag initialization and consistent detail access replace the previous nits.

`json_truncated` deliberately denotes a text-shape heuristic (starts with `{` or `[` and fails JSON parsing), not proof of token exhaustion. `finish_reason` remains in LLM telemetry; plumbing it through the content-return contract is unnecessary for this bounded change. Actual dominant production cause remains unverified until post-rollout data exists.

## Independently executed response probe

The parent executed `probe_1298.py` against this worktree. All 15 shapes below have distinct reasons; the probe asserts uniqueness.

| Response shape | Result | parse_reason |
|---|---|---|
| fenced-valid | accepted | `fenced_json` |
| fenced-invalid | rejected | `fenced_not_json` |
| fenced-unclosed | rejected | `fenced_unclosed` |
| closing-fence-then-prose | rejected | `fenced_trailing_text` |
| prose-then-fence | rejected | `prose_then_fence` |
| bare-truncated | rejected | `json_truncated` |
| prose-only | rejected | `not_json` |
| non-object | rejected | `not_object` |
| missing-key | rejected | `missing_or_invalid:summary` |
| cycle-id-missing | rejected | `cycle_id_missing` |
| cycle-id-wrong | rejected | `cycle_id_mismatch` |
| valid-finding-kind-empty-detail | rejected | `invalid_finding:empty_detail` |
| valid-recommendation-kind-empty-detail | rejected | `invalid_recommendation:empty_detail` |
| bad-finding-kind | rejected | `invalid_finding:bad_kind:unknown` |
| bad-recommendation-kind | rejected | `invalid_recommendation:bad_kind:unknown` |

## Verification

- Base verified: `a64470e5`; revision: `2917a84f`.
- `python -m pytest tests/test_reflector.py -q`: **26 passed**.
- `python -m pytest tests/ -q`: **3049 passed, 17 skipped, exactly 4 failed** in 1542.23 seconds. Executed with explicit `cd /t/Code/.worktrees/eeebot-1291`, redirected to `full-final-1298.log`; no pytest pipeline.
- Expected Windows failures only:
- Ruff for both Python files and `git diff --check`: passed.
- Coverage includes the requested response taxonomy, both item validators, missing/mismatched cycle ID, structured journal diagnostics, consecutive malformed attempts, transport exception diagnostics, repaired-fence count visibility, and existing watermark/unit tests.
- Previous five-failure reports are superseded: they were not valid verification of this worktree.
- Latest CI is pending; prior-head green is not evidence for this revision.
- No deployment, restart, host mutation, or merge. Main advanced to `ec8721a8` during verification; tested base remains the requested `a64470e5`. PR three-dot diff contains only four intended files.

### DIFF
--- a/docs/specs/observability/spec.md
+++ b/docs/specs/observability/spec.md
@@ -50,6 +50,24 @@ remains the explanatory reference (signal map, anti-patterns, walkthroughs).
   A change that cannot be observed after the fact from state SHALL NOT ship until
   it has that observability.
 
+### Reflector response diagnostics (#1291)
+
+Reflector error rows in `state/reflector/reflections.jsonl` carry a structured
+`parse_reason`, raw `response_chars`, and bounded `response_head` (200 characters)
+and `response_tail` (80 characters). Transport failures have `not_attempted` and
+empty response diagnostics, never a previous attempt's values.
+
+Malformed text distinguishes `fenced_unclosed`, `fenced_trailing_text`,
+`prose_then_fence`, `json_truncated`, and `not_json`. Complete plain/JSON fences
+are repaired; successes retain `fenced_json`, while invalid enclosed JSON gets
+`fenced_not_json`. `json_truncated` is a text-shape heuristic, not proof of a token
+limit. The telemetry recorder retains the authoritative `finish_reason`; the
+LLM content-return contract is unchanged. Schema diagnostics distinguish missing
+cycle IDs, mismatches, required fields, invalid kinds, and empty details.
+
+The reflector unit permits writes to the state tree, matching the strategist,
+so `state/llm_calls` recording is possible while systemd hardening remains intact.
+
 ## Scenarios
 
 ### Scenario: operator answers all seven questions from state
--- a/host/eeepc/systemd/eeebot-reflector.service
+++ b/host/eeepc/systemd/eeebot-reflector.service
@@ -21,4 +21,4 @@ NoNewPrivileges=true
 PrivateTmp=true
 ProtectHome=false
 ProtectSystem=strict
-ReadWritePaths=/var/lib/eeepc-agent/self-evolving-agent/state/reflector
+ReadWritePaths=/var/lib/eeepc-agent/self-evolving-agent/state
--- a/nanobot/runtime/reflector.py
+++ b/nanobot/runtime/reflector.py
@@ -287,26 +287,32 @@ def _messages(cycle_id: str, transcript: dict[str, Any], ledger: list[dict[str, 
     return [{"role": "system", "content": system}, {"role": "user", "content": user}]
 
 
-def _parse_output(value: Any, cycle_id: str) -> dict[str, Any] | None:
+def _parse_output(value: Any, cycle_id: str) -> tuple[dict[str, Any] | None, str]:
     if isinstance(value, str):
         try:
             value = json.loads(value)
         except Exception:
-            return None
-    if not isinstance(value, dict) or str(value.get("cycle_id") or "") != cycle_id:
-        return None
-    if not isinstance(value.get("summary"), str) or not isinstance(value.get("findings"), list) or not isinstance(value.get("recommendations"), list):
-        return None
+            return None, "not_json"
+    if not isinstance(value, dict):
+        return None, "not_object"
+    if str(value.get("cycle_id") or "") != cycle_id:
+        return None, "cycle_id_mismatch"
+    if not isinstance(value.get("summary"), str):
+        return None, "missing_or_invalid:summary"
+    if not isinstance(value.get("findings"), list):
+        return None, "missing_or_invalid:findings"
+    if not isinstance(value.get("recommendations"), list):
+        return None, "missing_or_invalid:recommendations"
     findings = []
     for item in value["findings"]:
         if not isinstance(item, dict) or item.get("kind") not in _VALID_FINDINGS or not str(item.get("detail") or "").strip():
-            return None
+            return None, f"invalid_finding:{item.get('kind') if isinstance(item, dict) else 'not_object'}"
         findings.append({"kind": item["kind"], "detail": str(item["detail"])[:1000]})
     recommendations = []
     for item in value["recommendations"][:_MAX_RECOMMENDATIONS]:
         if not isinstance(item, dict) or item.get("kind") not in _VALID_RECOMMENDATIONS or not str(item.get("detail") or "").strip():
-            return None
-        recommendations.append({"kind": item["kind"], "detail": str(item["detail"])[:1000], "evidence": str(item.get("evidence") or "")[:1000]})
+            return None, f"invalid_recommendation:{item.get('kind') if isinstance(item, dict) else 'not_object'}"
+        recommendations.append({"kind": item["kind"], "detail": str(item.get("detail") or "")[:1000], "evidence": str(item.get("evidence") or "")[:1000]})
     result = {
         "cycle_id": cycle_id,
         "summary": value["summary"][:2000],
@@ -316,7 +322,7 @@ def _parse_output(value: Any, cycle_id: str) -> dict[str, Any] | None:
     }
     if isinstance(value.get("mermaid"), str):
         result["mermaid"] = value["mermaid"][:4000]
-    return result
+    return result, "ok"
 
 
 def _default_llm(messages: list[dict[str, str]], model: str, cycle_id: str) -> str:
@@ -399,15 +405,15 @@ def run_reflector(
             messages = _messages(cycle_id, transcript, context_rows, _journal_tail(state_dir))
             model = resolve_model("reflector", strip_openai=True)
             response = llm(messages, model) if llm else _default_llm(messages, model, cycle_id)
-            parsed = _parse_output(response, cycle_id)
+            parsed, parse_reason = _parse_output(response, cycle_id)
             if parsed is None:
-                raise ValueError("malformed reflector output")
+                raise ValueError(f"malformed reflector output: {parse_reason}")
             _append_journal(state_dir, {**parsed, "timestamp": _now()})
             _save_watermark(state_dir, cycle_id)
             result["processed"] += 1
             consecutive_errs = 0
         except Exception as exc:
-            _append_journal(state_dir, {"cycle_id": cycle_id, "timestamp": _now(), "summary": "Reflector error; cycle will be retried.", "findings": [], "recommendations": [], "followed_previous": [], "status": "error", "error": f"{type(exc).__name__}: {exc}"[:500]})
+            _append_journal(state_dir, {"cycle_id": cycle_id, "timestamp": _now(), "summary": "Reflector error; cycle will be retried.", "findings": [], "recommendations": [], "followed_previous": [], "status": "error", "error": f"{type(exc).__name__}: {exc}"[:500], "response_head": str(response)[:200] if "response" in locals() else ""})
             result["errors"] += 1
             consecutive_errs += 1
             result["consecutive_errors"] = consecutive_errs
--- a/nanobot/runtime/reflector.py
+++ b/nanobot/runtime/reflector.py
@@ -298,6 +298,7 @@ def _parse_output(value: Any, cycle_id: str) -> tuple[dict[str, Any] | None, str
                     value = json.loads(value)
                 except Exception:
                     return None, "fenced_not_json"
+                parsed_from_fence = True
             else:
                 try:
                     value = json.loads(raw)
@@ -337,7 +338,7 @@ def _parse_output(value: Any, cycle_id: str) -> tuple[dict[str, Any] | None, str
     }
     if isinstance(value.get("mermaid"), str):
         result["mermaid"] = value["mermaid"][:4000]
-    return result, "ok"
+    return result, "fenced_json" if locals().get("parsed_from_fence") else "ok"
 
 
 def _default_llm(messages: list[dict[str, str]], model: str, cycle_id: str) -> str:
@@ -416,6 +417,7 @@ def run_reflector(
         context_rows = [row for row in rows if str(row.get("cycle_id") or "") == cycle_id]
         if cycle_id in proposed:
             context_rows.insert(0, proposed[cycle_id])
+        response: Any = None
         try:
             messages = _messages(cycle_id, transcript, context_rows, _journal_tail(state_dir))
             model = resolve_model("reflector", strip_openai=True)
@@ -423,12 +425,16 @@ def run_reflector(
             parsed, parse_reason = _parse_output(response, cycle_id)
             if parsed is None:
                 raise ValueError(f"malformed reflector output: {parse_reason}")
-            _append_journal(state_dir, {**parsed, "timestamp": _now()})
+            _append_journal(state_dir, {
+                **parsed,
+                "timestamp": _now(),
+                **({"parse_reason": parse_reason} if parse_reason != "ok" else {}),
+            })
             _save_watermark(state_dir, cycle_id)
             result["processed"] += 1
             consecutive_errs = 0
         except Exception as exc:
-            _append_journal(state_dir, {"cycle_id": cycle_id, "timestamp": _now(), "summary": "Reflector error; cycle will be retried.", "findings": [], "recommendations": [], "followed_previous": [], "status": "error", "error": f"{type(exc).__name__}: {exc}"[:500], "response_head": "".join(ch for ch in str(response)[:200] if ch >= " " or ch in "\t\n\r") if "response" in locals() else ""})
+            _append_journal(state_dir, {"cycle_id": cycle_id, "timestamp": _now(), "summary": "Reflector error; cycle will be retried.", "findings": [], "recommendations": [], "followed_previous": [], "status": "error", "error": f"{type(exc).__name__}: {exc}"[:500], "response_head": "".join(ch for ch in str(response)[:200] if ch >= " " or ch in "\t\n\r") if response is not None else ""})
             result["errors"] += 1
             consecutive_errs += 1
             result["consecutive_errors"] = consecutive_errs
--- a/nanobot/runtime/reflector.py
+++ b/nanobot/runtime/reflector.py
@@ -289,10 +289,25 @@ def _messages(cycle_id: str, transcript: dict[str, Any], ledger: list[dict[str, 
 
 def _parse_output(value: Any, cycle_id: str) -> tuple[dict[str, Any] | None, str]:
     if isinstance(value, str):
-        try:
-            value = json.loads(value)
-        except Exception:
-            return None, "not_json"
+        raw = value.strip()
+        if raw.startswith("```"):
+            first_line, _, remainder = raw.partition("\n")
+            if first_line.strip().lower() in {"```", "```json"} and remainder.rstrip().endswith("```"):
+                value = remainder.rstrip()[:-3].strip()
+                try:
+                    value = json.loads(value)
+                except Exception:
+                    return None, "fenced_not_json"
+            else:
+                try:
+                    value = json.loads(raw)
+                except Exception:
+                    return None, "not_json"
+        else:
+            try:
+                value = json.loads(raw)
+            except Exception:
+                return None, "not_json"
     if not isinstance(value, dict):
         return None, "not_object"
     if str(value.get("cycle_id") or "") != cycle_id:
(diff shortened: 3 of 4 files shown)


ozand:
VERDICT: CHANGES REQUESTED — the observability direction is right, but the parse-reason contract is incomplete and the response capture is not sufficiently self-contained/safe.

## Findings

### [BLOCKER] `_parse_output` does not distinguish all validation causes

The new reasons are useful for:

- `not_json`;
- `not_object`;
- `cycle_id_mismatch`;
- missing/invalid `summary`, `findings`, or `recommendations`;
- invalid finding kind;
- invalid recommendation kind.

However, the issue explicitly identifies fenced JSON as a distinct failure cause. A fenced response still goes through `json.loads()` and is recorded only as `not_json`, collapsing it with arbitrary non-JSON text. The acceptance requires the reason set to distinguish the distinct `_parse_output` causes, so fenced JSON needs its own reason (and test), or the implementation must explicitly document why it is intentionally grouped.

Also, the currently shown focused tests cover only `not_json`, cycle mismatch, and invalid finding kind; they do not pin the missing-key, invalid recommendation, non-object, or fenced cases.

### [MAJOR] `response_head` is bounded but not proven safe for journal content

`str(response)[:200]` is bounded, which is good, but the journal stores the raw response head. The PR does not test control-character/newline handling or explicitly sanitize the captured text. A model response can therefore inject newlines/control characters into a JSONL journal line's string value (JSON encoding will escape it when written, so this is not a raw JSONL-breakout, but it still makes the journal content potentially noisy/unsafe for terminal or downstream display). Add a test/contract for the intended sanitization or explicitly justify raw-but-JSON-encoded capture and bound it accordingly.

### [MAJOR] No test proves the widened service boundary is the exact intended scope

The unit change is narrowly scoped to:

```text
ReadWritePaths=/var/lib/eeepc-agent/self-evolving-agent/state
```

This matches the strategist unit and does not relax `ProtectSystem=strict`, `NoNewPrivileges`, `PrivateTmp`, or `ProtectHome`. That part is correct. But the PR has no test or static assertion guarding the unit hardening invariants against accidental weakening. Since this is a security-boundary change, add focused unit-file assertions that `ProtectSystem=strict`, `NoNewPrivileges=true`, `PrivateTmp=true`, `ProtectHome=false`, and the exact state-tree `ReadWritePaths` remain present.

## Positive verification

- `ReadWritePaths` is widened only from `state/reflector` to the state tree; no other hardening directive changed.
- Telemetry call path remains under `call_context(cycle_id, "reflector")`, with `record_llm_call` and `record_llm_prompt`; the unit boundary now permits the target state subtree.
- `response_head` is capped at 200 characters.
- Focused reflector tests on the PR branch: **11 passed**.
- `git diff --check`: passed.

Do not merge until the reason taxonomy and security-boundary tests are made explicit and focused tests pass for the actual acceptance surface.

ozand:
Review blockers addressed on the rebased branch:

- Fenced JSON is now deliberately repaired: a complete ```json fence is stripped and its contents parsed. A valid repaired response is journalnante withfangled `fenced_json`кому?

ozand:
Correction/superseding the malformed preceding comment:

Review blockers are addressed on the branch rebased onto current `origin/main` (`3dbaa3a4`):

- Fenced JSON is deliberately repaired. A complete plain/`json` Markdown fence is stripped and the enclosed value parsed.
- A successfully repaired response is journaled with parse reason `fenced_json`, so the fencing rate remains countable instead of disappearing into success.
- Invalid fenced content is recorded separately as `fenced_not_json`; arbitrary non-JSON prose remains `not_json`.
- Tests now cover fenced valid JSON, fenced invalid JSON, non-object output, missing required key, invalid finding kind, invalid recommendation kind, and cycle-id mismatch.
- The service-unit test asserts the exact intended write boundary (`/var/lib/eeepc-agent/self-evolving-agent/state`) and asserts the narrower old path is absent while `ProtectSystem=strict`, `NoNewPrivileges=true`, and `PrivateTmp=true` remain.
- `response_head` stays a bounded 200-character diagnostic; control characters are removed in one expression before JSON encoding, without adding a sanitization subsystem.

Focused reflector suite: 14 passed.

ozand:
**Verdict: REQUEST CHANGES (small, one place) — the unit change is exactly right, the fence repair is safe and stays countable, but the reason set still collapses the two most likely production causes (max_tokens truncation, bare or inside a fence) into `not_json`, and the error row lacks the fields that would separate them. One day of data will therefore not answer "one dominant cause or a spread" from the journal alone.** Everything below was run against the branch at `441a9363`: 14 reflector tests pass; the parse function was probed with 14 response shapes.

### 1. Unit — a security boundary, read from the file

The diff is one line: `ReadWritePaths=/var/lib/eeepc-agent/self-evolving-agent/state/reflector` → `…/state`. `ProtectSystem=strict`, `NoNewPrivileges=true`, `PrivateTmp=true` unchanged; `ProtectHome=false` was already there; `ExecStartPre` mkdir/chown unchanged. The grant is the state tree and nothing wider — byte-identical to `eeebot-strategist.service`'s line, which is the precedent. It is wider than the minimum the recorder needs (`state/llm_calls`, `state/llm_calls/prompts`, and `state/action_index`, which `record_llm_prompt` rotates via `build_action_index`), so if you want the narrowest grant, list those three plus `state/reflector`; matching the strategist is a defensible choice, say which you intend. The unit test pins the three hardening directives and the exact new line. Good.

### 2. Fence repair — hides nothing

- Repair applies only when the first line is exactly ```` ``` ```` or ```` ```json ```` (case-insensitive, stripped) and the text ends with ```` ``` ````. The inner text then goes through the *same* validation as bare JSON, so nothing the old path correctly rejected on content is now accepted; only the fence is removed.
- A repaired success is journaled with `parse_reason: "fenced_json"` and the row otherwise looks like a clean parse (`processed += 1`). Nothing counts it as `ok`; the success row without `parse_reason` is the only "clean" shape. Countable: `grep -c '"parse_reason": "fenced_json"'`.
- Fenced-but-invalid → `fenced_not_json`. Verified.
- One loss: a fenced response whose *content* then fails validation reports only the content reason (`cycle_id_mismatch`, `invalid_finding:…`) — the fence fact is dropped. Minor; a `fenced:` prefix or a second field would keep it.

### 3. Reason set vs the actual failure branches — where it still collapses

Probe results on the branch (first token = accepted/rejected, second = reason):

```
OK  ok                          bare valid
OK  fenced_json                 ```json … ```   (also ```JSON, and "```json " with trailing space)
REJ fenced_not_json             ```json not-json ```
REJ not_object / cycle_id_mismatch / missing_or_invalid:<key> / invalid_*:not_object   — as designed
REJ not_json   <- fenced, UNCLOSED (cut by max_tokens inside the fence)
REJ not_json   <- fenced, closing fence then trailing prose ("Hope this helps!")
REJ not_json   <- prose, then a valid fence
REJ not_json   <- bare JSON truncated (cut by max_tokens)
REJ not_json   <- prose only
REJ cycle_id_mismatch           <- cycle_id missing entirely (same code as wrong id)
REJ invalid_finding:error_pattern        <- kind VALID, detail EMPTY   ← reads as "kind error_pattern is invalid"
REJ invalid_recommendation:skill_candidate <- kind VALID, detail EMPTY  ← same misattribution
```

Five distinct shapes land in `not_json`. Two of them are the prime suspects for 425 errors on a model called with `max_tokens=1600`: an unterminated fence and a bare object cut mid-way — both are truncation, and both are indistinguishable in the row from "the model wrote prose". Prose-before-fence is the third common shape (chatty model) and is also repairable, but this PR does not attempt it; fine to leave unrepaired, not fine to leave uncounted.

`invalid_finding:<kind>` when the kind is valid and `detail` is empty is not a collapse but a mislabel: a reader will conclude the model invents kinds when it is actually emitting empty details.

### 4. Will one day of journal rows be decisive on their own?

Not yet. `response_head` (200 chars, control characters stripped — good) shows how a response *starts*; truncation shows at the *end*. To classify `not_json` without opening `llm_calls` (which rotates), the error row needs either sub-reasons or the raw facts. Smallest change that makes the day decisive, all inside `_parse_output` / the except branch:

- Sub-classify `not_json` from the text itself: `fenced_unclosed` (starts with ```` ``` ````, no closing fence), `fenced_trailing_text` (closing fence followed by non-whitespace), `prose_then_fence` (fence not at start), `json_truncated` (starts with `{` or `[`, `json.loads` fails), `not_json` (everything else). Five string checks, no new dependencies.
- Split `invalid_finding` / `invalid_recommendation` into `…:bad_kind:<kind>` and `…:empty_detail`; and `cycle_id_missing` vs `cycle_id_mismatch`.
- Put the reason in its own key on **error** rows too (`parse_reason`), not only inside `error: "ValueError: malformed reflector output: …"`. Today the success path has the key and the error path has a substring; counting should not need a regex over a free-text field.
- Record `response_chars: len(response)` and a `response_tail` of the last ~80 chars alongside `response_head`. With `chars` near the 1600-token ceiling and a tail that ends mid-token, truncation is provable from the row alone. `finish_reason` would be better still, but `_default_llm` returns only `content`; adding it means changing that return or stashing it on a module attribute — your call whether that is worth it now that `llm_calls` will actually record it (unit fix), given the rotation caveat.

With those, the post-deploy question — "dominant cause or spread" — is one `Counter` over `parse_reason` in `reflections.jsonl`, no second file.

### Nits

- `return result, "fenced_json" if locals().get("parsed_from_fence") else "ok"` — `locals().get` for a flag that may be undefined; initialise `parsed_from_fence = False` at the top and return it plainly.
- Recommendations use `str(item.get("detail") or "")` where findings use `str(item["detail"])`; both are validated non-empty just above, so harmless — unify.
- PR body says the five-failure Windows baseline; #1295 has merged, so after a rebase it should read four. Body also lists tests for "malformed JSON, cycle mismatch, invalid finding kind" — the branch has more than that (fenced valid/invalid, not_object, missing key, invalid recommendation, unit); update it so the reviewer of record sees what is actually covered.

### Verified

- `tests/test_reflector.py` at `441a9363`: 14 passed (Windows).
- Downstream reader of `reflections.jsonl` (`knowledge_curator`) reads by key and ignores extras; the new `parse_reason` / `response_head` keys are safe for it.
- Host not touched; this PR's unit change needs a `daemon-reload` at deploy, which the release script already does for unit files.

ozand:
| Response shape | Result | parse_reason |
|---|---|---|
| fenced-valid | accepted | `fenced_json` |
| fenced-invalid | rejected | `fenced_not_json` |
| fenced-unclosed | rejected | `fenced_unclosed` |
| closing-fence-then-prose | rejected | `fenced_trailing_text` |
| prose-then-fence | rejected | `prose_then_fence` |
| bare-truncated | rejected | `json_truncated` |
| prose-only | rejected | `not_json` |
| non-object | rejected | `not_object` |
| missing-key | rejected | `missing_or_invalid:summary` |
| cycle-id-missing | rejected | `cycle_id_missing` |
| cycle-id-wrong | rejected | `cycle_id_mismatch` |
| valid-finding-kind-empty-detail | rejected | `invalid_finding:empty_detail` |
| valid-recommendation-kind-empty-detail | rejected | `invalid_recommendation:empty_detail` |
| bad-finding-kind | rejected | `invalid_finding:bad_kind:unknown` |
| bad-recommendation-kind | rejected | `invalid_recommendation:bad_kind:unknown` |

ozand:
**Verdict: APPROVE — it holds. Re-derived the table myself against `2917a84f`: every one of the 14 shapes I used to find the collapse now lands on its own reason, the error row stands alone (own `parse_reason` key, `response_chars`, `response_tail`), and one `Counter` over `reflections.jsonl` answers "dominant cause or spread". Four residual bundles surfaced by extra adversarial shapes are listed below as non-blocking follow-ups; each is resolvable from fields already in the same row.**

### 1. The table, re-derived (28 shapes, my probe, not the author's)

My original 14 — the collapse finding — now all distinct:

```
OK  ok                               bare valid
OK  fenced_json                      ```json … ```   (also ```JSON; "```json " with trailing space)
REJ fenced_unclosed                  ```json … cut by max_tokens inside the fence
REJ fenced_trailing_text             ```json … ``` Hope this helps!
REJ prose_then_fence                 Here is the reflection: ```json … ```
REJ json_truncated                   {"cycle_id": … cut
REJ not_json                         prose only
REJ cycle_id_mismatch                fenced, valid syntax, wrong cycle_id
REJ invalid_finding:empty_detail     kind valid, detail ""            ← was invalid_finding:error_pattern
REJ invalid_finding:not_object       finding is a string
REJ cycle_id_missing                 no cycle_id key                  ← was cycle_id_mismatch
REJ invalid_recommendation:empty_detail
```

Plus `invalid_finding:bad_kind:process`, `invalid_recommendation:bad_kind:nope`, `invalid_finding:bad_kind:None`, `fenced_not_json` (```yaml with JSON inside), `not_object` (array), `missing_or_invalid:summary` / `:findings`. The **empty_detail split fires on the right branch** — `bad_kind` is only reached when the kind is actually invalid and reports the offending value — which was the actively wrong label before.

### 2. The row stands alone — driven through `run_reflector`, not read

Eight end-to-end rows, one per shape, `_seed` from the PR's own test module:

```
fenced valid     processed=1 status=ok     parse_reason=fenced_json                (success row carries the repair fact)
unclosed         status=error parse_reason=fenced_unclosed             chars=68  tail='```json\n{"cycle_id": "c1", …e'
truncated        status=error parse_reason=json_truncated              chars=80  tail='…"findings": [{"kind": "error_pattern", "deta'
empty_detail     status=error parse_reason=invalid_finding:empty_detail chars=138
bad_kind         status=error parse_reason=invalid_finding:bad_kind:process
prose            status=error parse_reason=not_json                    chars=31  tail='I cannot reflect on this cycle.'
fenced bad cycle status=error parse_reason=cycle_id_mismatch           chars=154 tail='…}\n```'
llm raised       status=error parse_reason=not_attempted               chars=0   error='RuntimeError: Error code: 500 …'
Counter({'fenced_json': 1, 'fenced_unclosed': 1, 'json_truncated': 1, 'invalid_finding:empty_detail': 1,
         'invalid_finding:bad_kind:process': 1, 'not_json': 1, 'cycle_id_mismatch': 1, 'not_attempted': 1})
```

Every error row has `parse_reason` as its own key, `response_chars`, `response_head` (200) and `response_tail` (80), control characters stripped. Truncation is provable from the row alone: `json_truncated` / `fenced_unclosed` plus a tail that ends mid-token plus `response_chars` near the 1,600-token ceiling — no `llm_calls` needed. Transport failures are `not_attempted` with `response_chars: 0`, so they cannot be mistaken for a parse shape. Yes: one `Counter(row.get("parse_reason", "ok"))` over the file answers the question.

### 3. Fence repair

`fenced_json` is journaled on the success row and nowhere else, so a repaired response is countable and never reads as a clean parse. **Fenced-then-invalid content still reports the bare content reason** (`cycle_id_mismatch` above, no fence fact) — not addressed, and not stated in the body as a choice. Minor, as before; say so in the body or prefix the reason. Not blocking.

### 4. Residual bundles from the extra shapes (non-blocking; each disambiguable from the row)

| reason | shapes sharing it | how the row still tells them apart | cheap refinement |
|---|---|---|---|
| `json_truncated` | bare JSON cut by max_tokens **and** bare JSON followed by prose (`{…}\nHope this helps!`) | tail ends mid-token vs ends in prose | `json.JSONDecoder().raw_decode(raw)`: success with a remainder → `json_then_trailing_text` (and it is repairable, like the fence) |
| `not_json` | prose only, prose then **bare** JSON, empty string | `response_chars: 0`; tail is JSON vs prose | `empty_response` for chars 0; `raw_decode` from the first `{` for prose-then-JSON |
| `fenced_trailing_text` | closing fence then prose **and** a *valid* fenced response whose inner JSON contains ` ``` ` in a string | tail ends with ```` ``` ```` in the valid case | close on the **last** fence (`rfind`) instead of the first — one token; the valid response is currently rejected |
| `prose_then_fence` | prose then fence **and** a BOM before the fence | head starts with `﻿` | strip `﻿`; rare from a model |

None of these is one of the shapes the incident data pointed at, and none hides a class from the day-one count: the Counter will name `json_truncated`, `fenced_unclosed`, `not_json` and the two `empty_detail` labels as top-level buckets, and the tail column resolves the sub-shapes without a second file. Worth a follow-up issue once the first day's distribution says which bucket matters.

### 5. Unit and tests

Unit diff unchanged since the first review: one line, `ReadWritePaths` `…/state/reflector` → `…/state`, matching the strategist unit; hardening directives untouched. `tests/test_reflector.py` at head: 26 passed. Branch is on `a64470e5`; `main` has moved (#1304), no overlap in files — a clean rebase or a merge commit, reviewer's choice.

One day of rows after deploy will answer the question this was for.

PR #1301, opened 2026-09-05
fix(dashboard): complete semantic color follow-ups (#1299)

### LINKED ISSUE #1299: Dashboard status colours: four small follow-ups measured during the #1287 review
## Context (self-contained)

#1277 shipped as #1287 (`08114a75`): the dashboard now carries semantic status colours and `data-status` / `data-authoritative` / `data-context-only` attributes, so an `unavailable` source no longer renders identically to a healthy one.

Four small things were measured during that PR's review and deliberately left out of it. None is a defect in what shipped; all are cheap, and each would otherwise be rediscovered later by someone re-deriving the same measurements.

## The four

**1. The `.status-badge[data-status=…]` rule block is emitted twice** — `scripts/eeebot_dashboard.py:2317-2323` and `:2324-2338`. The second is the complete one and wins. Drop the first.

**2. `filled.setdefault("reward_trend", [])` in the test helper is inert.** `render_html` calls `sanitize_public_metrics` first, which sets `sanitized["reward_trend"] = []` (`:1187`), so the fixture value never reaches `_build_html_context` — verified with a spy during review, even with `reward_trend=[("cycle-x", 1.0)]` set. Harmless, but it implies a contract the code does not have. Remove it, or keep it with a comment saying why it cannot matter.

**3. The trend-badge tokens have no foreground rule.** `--trend-good/neutral/bad` (`#166534`, `#1d4ed8`, `#b91c1c`) inherit their text colour. Against `--text #f8fafc` they measure 6.8 / 6.4 / 6.2 and pass AA; against `--muted #94a3b8` they measure 2.8 / 2.6 / 2.5 and fail. This is currently moot on the public page — `sanitize_public_metrics` strips `reward_trend`, so trend badges never render there — which is exactly why it will be missed if trend badges ever come back. Give them an explicit foreground token and add the pair to `test_status_tokens_meet_wcag_aa_contrast`.

**4. `context-only` maps to the offline hue** (`:2335`), so it is colour-identical to `unavailable`; only the attributes distinguish them. That may well be intended — both mean "not authoritative" — but it is currently implicit. Either say so in a comment or give it its own token.

## Also worth knowing, not a task

The critical and offline hues are 224° apart (`#7f1d1d` at 0°, `#1e3a8a` at 224°) but share lightness, background-against-background contrast 1.03. So the distinction #1173 asked for is carried by hue and by the badge text, not by lightness — which is fine for a sighted reader on a colour display and is exactly why the `data-status` attribute matters for everything else. Do not "fix" the lightness without a reason; changing it would disturb the four AA pairs that currently pass.

## Acceptance Criteria

- [ ] The duplicate rule block is gone and the rendered page is unchanged — assert the rendered CSS contains the `data-status` rule exactly once.
- [ ] Item 2 resolved either way, with the reason in the code if it stays.
- [ ] Trend tokens have an explicit foreground and are covered by the contrast test at the 4.5:1 floor.
- [ ] Item 4 is either commented as intended or given its own token.
- [ ] The deploy gate's `/` status and size checks still pass; the page must stay above the 1024-byte floor.

ozand:
Closes #1299.

## Summary

- remove the duplicate emitted status/trend badge CSS rule block and assert each rendered selector occurs exactly once;
- remove the inert `reward_trend` fixture default: public sanitization clears that field before HTML context construction;
- explicitly use the existing `--text` foreground for all trend badges and cover all three background/foreground pairs at the WCAG AA 4.5:1 floor;
- document that `context-only` intentionally shares the offline palette because both are non-authoritative and remain distinguished by badge text and semantic attributes.

The critical/offline hues and lightness from #1287 are unchanged.

## Verification

- focused dashboard truth suite: `29 passed`;
- `ruff check tests/test_eeebot_dashboard_truth.py`: passed;
- `git diff --check`: passed;
- static HTML export: 18,070 bytes, above the 1,024-byte deploy floor;
- full Windows pytest, output redirected to `full-pytest-1299.log`: `4 failed, 3025 passed, 17 skipped`.

The four failures are exactly the established Windows baseline:

1. `tests/test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle`
2. `tests/test_bridge_locking.py::TestAcquireBridgeLock::test_acquires_when_free`
3. `tests/test_bridge_locking.py::TestAcquireBridgeLock::test_second_acquire_in_same_process_is_contended`
4. `tests/test_bridge_locking.py::TestAcquireBridgeLock::test_contended_lock_via_monkeypatched_flock`

Branch was created from and rebased onto current `origin/main` (`3dbaa3a4`, which includes #1287 `08114a75`). No deployment or host mutation was performed.

### DIFF
--- a/scripts/eeebot_dashboard.py
+++ b/scripts/eeebot_dashboard.py
@@ -2314,13 +2314,6 @@ _HTML_TEMPLATE = """<!DOCTYPE html>
             font-size: 12px;
             font-weight: 600;
         }}
-        .status-badge[data-status="fresh"], .status-badge[data-status="valid"], .status-badge[data-status="nominal"] {{ background: var(--state-nominal); color: var(--state-nominal-fg); }}
-        .status-badge[data-status="stale"], .status-badge[data-status="caution"] {{ background: var(--state-caution); color: var(--state-caution-fg); }}
-        .status-badge[data-status="malformed"], .status-badge[data-status="error"], .status-badge[data-status="failed"], .status-badge[data-status="critical"] {{ background: var(--state-critical); color: var(--state-critical-fg); }}
-        .status-badge[data-status="unavailable"], .status-badge[data-status="offline"] {{ background: var(--state-offline); color: var(--state-offline-fg); }}
-        .trend-badge.trend-good {{ background: var(--trend-good); }}
-        .trend-badge.trend-neutral {{ background: var(--trend-neutral); }}
-        .trend-badge.trend-bad {{ background: var(--trend-bad); }}
         .status-badge[data-status="fresh"],
         .status-badge[data-status="valid"],
         .status-badge[data-status="nominal"] {{ background: var(--state-nominal); color: var(--state-nominal-fg); }}
@@ -2332,10 +2325,11 @@ _HTML_TEMPLATE = """<!DOCTYPE html>
         .status-badge[data-status="critical"] {{ background: var(--state-critical); color: var(--state-critical-fg); }}
         .status-badge[data-status="unavailable"],
         .status-badge[data-status="offline"] {{ background: var(--state-offline); color: var(--state-offline-fg); }}
+        /* context-only intentionally shares the offline palette: both convey non-authoritative state, distinguished by attributes */
         .status-badge[data-status="context-only"] {{ background: var(--state-offline); color: var(--state-offline-fg); }}
-        .trend-badge.trend-good {{ background: var(--trend-good); }}
-        .trend-badge.trend-neutral {{ background: var(--trend-neutral); }}
-        .trend-badge.trend-bad {{ background: var(--trend-bad); }}
+        .trend-badge.trend-good {{ background: var(--trend-good); color: var(--text); }}
+        .trend-badge.trend-neutral {{ background: var(--trend-neutral); color: var(--text); }}
+        .trend-badge.trend-bad {{ background: var(--trend-bad); color: var(--text); }}
         .path {{
             font-family: monospace;
             font-size: 12px;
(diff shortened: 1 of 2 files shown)


ozand:
**Verdict: APPROVE.** All four items closed, verified by running rather than by reading the diff. I did not write this and did not write #1287, so this is an independent check.

**Rendered page, not source** — the duplicate rule block was the item most likely to look fixed while still emitting twice, so I counted occurrences in the output of `render_html`:

```
[data-status="unavailable"]     1
[data-status="context-only"]    1
.trend-badge.trend-good         1
inline style="background:       False
page bytes                      17498
```

Each rule appears exactly once and no inline background survives. 17,498 bytes is comfortably above the deploy gate's 1,024-byte floor.

**Contrast now covers seven pairs, and all clear AA:**

```
#065f46 / #6ee7b7   5.04   #92400e / #fcd34d   4.92
#7f1d1d / #fecaca   6.93   #1e3a8a / #bfdbfe   7.29
#166534 / #f8fafc   6.81   #1d4ed8 / #f8fafc   6.41
#b91c1c / #f8fafc   6.18
```

The three trend pairs are the ones that mattered: they had no foreground rule and inherited, and the whole point was that they are currently unreachable — `sanitize_public_metrics` strips `reward_trend`, so nothing renders them and nothing would have caught them coming back. They now have an explicit `color: var(--text)` and are pinned by the test.

**`context-only` sharing the offline palette is now stated** rather than implicit, with the reason: both convey non-authoritative state and are distinguished by attributes. That was the right call over inventing a fifth hue.

**The inert `reward_trend` default is gone.** That one was mine, from #1287, and it implied a contract the code does not have.

`tests/test_eeebot_dashboard_truth.py`: 29 passed. CI green on all three Pythons.

One observation, not blocking and not a request: two status pairs sit at 5.04 and 4.92 against a 4.5 floor. That is real headroom but not much, so a future palette nudge could push `#92400e / #fcd34d` under without anyone intending it. The test will catch it — which is exactly what it is for — but worth knowing that caution is the tightest pair before someone adjusts it.

Issue #1193, opened 2026-09-02
AGENTS.md is operator-owned: remove the loop's write path (allowlist + proposer prompt), refuse with a ledger-visible reason


ozand:
Parent: #1188 (policy decision, comment of 2026-09-02). Independent of #1191 (runtime cap defect) — this stops the file growing, #1191 stops growth from silently evicting memory and skills; both are needed, neither waits for the other. Touches `gate.py` / `llm_proposer.py` only; `demand.py` change is one annotation. Do not add a size cap on `AGENTS.md` (see #1183, #1178, #1166).

## Problem

`AGENTS.md` is the instruction the loop reads on every cycle and the cheapest artifact it can write: the one root path allowlisted past the prefix check (`gate.py` `_ALLOWED_EXACT_PATHS` / `_GATE_BASENAME_ALLOWLIST`, `llm_proposer.py` `_ALLOWED_EXACT_PATHS`), named as an acceptable target in the proposer prompt three times (`llm_proposer.py:263`, `:304`, `:1413`), doc-only so it cannot fail a test. Measured on the host (UTC): 20 integrated reflection-lane cycles in five days changed only `AGENTS.md`, 20 distinct demand ids, +380/−6 lines; 91% of the file was written 2026-08-27 → 2026-09-01; the file crossed the 24,000-char system-prompt budget at 2026-09-01 16:34 UTC (19:34 MSK) and every cycle since runs without memory or skills (#1191). The guidance written in that window had no measurable effect on the behaviour it targets (zero-edit share 22% on 08-28 → 45–50% on 08-31…09-02). The #1090 doc-only budget never engaged against these cycles (rotation-blind counter + `tests/test_agents_structure.py` co-changes reclassifying them as code-bearing). Full evidence on #1188.

## Decision (binding for this issue)

`AGENTS.md` is operator-owned. The loop keeps its three designed self-teaching channels — `skills/` (procedural how-to, progressively loaded), `lessons/` (fed to the proposer on 272 of 273 recent proposals via `lessons_context`), `memory/` — all of which stay writable and already carry the same themes (skills `early-turn-staging`, `fast_path_redundancy_skip`, `targeted-test-discovery` were created 2026-08-28…08-31 alongside the equivalent `AGENTS.md` sections).

## Scope

[...]

PR #1302, opened 2026-09-05
fix(agent): never let the prompt cap pick surviving instructions by position (#1300)

### LINKED ISSUE #1300: System-prompt cap silently drops the tail of instance AGENTS.md (11,688 of 22,972 chars, 11 sections) in every executor prompt since 2026-09-01
## Context (self-contained)

The executor's system prompt is assembled by `nanobot/agent/context.py::ContextBuilder.build_system_prompt` in the order identity → bootstrap (`AGENTS.md` only, `BOOTSTRAP_FILES = ["AGENTS.md"]`) → active skills → skills catalogue → memory, then the bridge appends the operator charter and loop identity as `system_context` (`bridge.py:2666`). The builder caps its part at `MAX_SYSTEM_PROMPT_CHARS = 24000` and, since #1191, trims **bootstrap first, by whole lines from the end** (`_fit_system_prompt`, `_trim_lines`). The only trace of a trim is `logger.warning("System prompt cap dropped content: bootstrap=N chars")` — journal only.

The instance loop keeps appending guidance sections to its own `AGENTS.md` (8 `docs:` commits 2026-08-29 → 09-01). The file grew from 2,062 chars (`20cfd387`, 2026-08-24) to 22,972 chars (`4dfc6ee4`, 2026-09-01 19:34 MSK). Since that commit, **every executor prompt drops the last 11,688 chars of AGENTS.md — 51% of the file, eleven whole sections.**

## Evidence (host, read-only, 2026-09-05)

Newest real prompt dump (`SELFEVO_DUMP_PROMPTS=1` is on; `state/prompts/cycle-0a776b606626.system.txt`, 2026-09-04T23:59:54Z, 28,385 chars):

```
section offsets: '## AGENTS.md' 1453 · '# Skills' 12758 · '# Memory' 19941 · '# Immutable operator charter' 23978 · '# Loop agent identity' 26746
AGENTS.md section as delivered: 11,298 chars, 210 lines  (file on disk: 22,972 chars, 417 lines)
headings KEPT (last 6):  Turn budget checkpoints · Early turn direct file inspection and editing · Early skip evaluation ·
                         Target git log history inspection · Bounded reading & search · Command output bounding
headings DROPPED (all):  Composite context inspection · Early implementation transition · Demand and reflection lookup paths ·
                         Working knowledge · Lesson markdown schema (required headers) · Canonical skill path layout · Script contract ·
                         Secret verification scoping · Targeted test validation (legacy glob suites) ·
                         Pure documentation changes (test-bypass workflow) · Standard test runner (unittest over pytest)
first dropped text: "4. **Truncation is a signal, not an error.** …"
```

Read-only rebuild with the deployed `ContextBuilder` (loop profile, loop-excluded skills) on the live workspace: `section chars: identity 1446, bootstrap 22986, skills_catalogue 6951, memory 4030 | joined 35434 | cap 24000` → `WARNING System prompt cap dropped content: bootstrap=11688 chars`. Journal (this boot only, from 2026-09-04 12:44): 63 identical warnings on 09-04/09-05 — one per executor spawn. Nothing in `state/` records it; the journal on this host holds about one day.

Per-commit reconstruction (AGENTS.md size, skills count and MEMORY.md size at each sha; estimate of the kept bootstrap = cap − other sections) puts the threshold crossing at `4dfc6ee4` 2026-09-01 19:34: kept ≈16,104 chars vs the file's 22,972. The 09-02 executor prompts recorded `prompt_tokens ≈ 10,000` at seq 1, consistent with a capped ~24 K-char system prompt plus a ~7 K-char task.

## What it silently removed — one measured casualty

The #957 knowledge-base pointer lives under `## Working knowledge` at line 291 (char 16,341):

```
- `docs/` is the reference library: read `docs/index.md` before mutating this
  file or creating/improving a skill (`docs/agents/` covers AGENTS.md practice,
  `docs/skills/` covers skill authoring).
```

While it was inside the kept window (2026-08-24 22:36 → 2026-09-01 19:34), **6 of 10 skill-creating cycles read the KB**, exact `read_file` paths from the durable prompt records: `docs/index.md`, `docs/skills/index.md`, `docs/skills/anatomy.md`, `docs/skills/create-vs-improve.md`, `docs/skills/evaluate.md`, `docs/skills/stdlib-scripts.md` (cycles `37d1214b8c85`, `b47428b1f3e6`, `5fcf9a4965b0`, `0416827ab4cc`, `fb57aefb1997`, `049146789e69`). After the cut: **0 of 8** (all eight skill-creating cycles on 2026-09-02). #1294 ("the KB is inert, 26 for 26") was reading the action index, whose `normalize_action` collapses those paths to `read:docs/*.md`; the exact paths are only in `llm_calls/prompts/` (14-day retention). Details on #1294.

The other ten dropped sections — test runner policy, script contract, lesson schema, secret-scan scoping, skill path layout — are rules the loop wrote for its executor that the executor has not seen for four days. Nothing measures their compliance either.

## Why this is not #1294 and not a prompt-text fix

#1294 asks whether the pointer is acted on. This issue is that a growth mechanism the loop controls (appending to AGENTS.md) silently deletes standing instructions from the bottom up, with no state-side trace. Making the pointer "louder" or appending anything to AGENTS.md lands in the dropped tail. Not proposing prompt content here — the arm is the operator's; what the evidence rules out is any remedy that adds text below line ~210 of the current file.

## Acceptance

- [ ] The drop is recorded durably per cycle (e.g. `dropped_sections`/`dropped_chars` in the cycle's `started` ledger row or the prompt dump), not only as a journal WARNING.
- [ ] Operator decision on the mechanism: raise/split the cap, trim a different section first, bound AGENTS.md growth (the loop's own `docs:` appends), or move standing rules out of bootstrap — recorded on this issue.
- [ ] Live check after whichever change: a prompt dump whose AGENTS.md section equals the file on disk (or a recorded, intended drop), and `## Working knowledge` present in it.
- [ ] Related: #1294 (KB pointer), #1191 (bootstrap-first trim), #957 (KB), #1193 (AGENTS.md self-edit surface).

ozand:
Closes #1300. Same class and same remedy shape as #1284 (PR #1288): a bound that chose what survives by position is replaced by a declared choice, and the choice is recorded.

**Ships with its data:** the declaration lives in the instance repo — **ozand/eeebot-self-evolving#186** — and must be merged (and fetched by the bridge) **before** this PR is deployed. Deployed the other way round, the first cycle fails `system_prompt_overflow`, exits 4, and the deploy gate rolls the release back: by design, harmless, wasted.

## Defect

`ContextBuilder._fit_system_prompt` caps the builder's part of the executor system prompt at 24,000 chars and trimmed the bootstrap (`AGENTS.md`) first, whole lines from the end. On 2026-09-01 19:34 the instance `AGENTS.md` reached 22,972 chars and the total crossed (`22986 + 1446 + 6951 + 4030 = 35434`). Every executor prompt since carried ~210 of 417 lines and lost eleven `##` sections, `## Working knowledge` among them; the only trace was a journal `WARNING`. Measured effect on one casualty: KB uptake 6-of-10 skill cycles with the pointer present, 0-of-8 without. **Trimming from the end is the defect, not the cap.**

## Fix — mechanism (this PR)

- Loop profile is **strict** (`build_system_prompt(..., strict=None)` → `loop_profile`). The only content the cap may remove is a bootstrap `## ` section whose body contains **`<!-- prompt-fit: droppable -->`** (`ContextBuilder.DROPPABLE_MARKER`), removed **whole, largest first**. Every unmarked section is critical.
- **Decision, stated:** the cap never drops a critical section. If critical content still does not fit, `build_system_prompt` raises `SystemPromptOverflowError(over_by, cap, sections, dropped)`. No line-trimming in strict mode. Interactive sessions keep the pre-#1300 behaviour (stated).
- Both profiles leave `builder.last_fit`; `NANOBOT_SYSTEM_PROMPT_MAX_CHARS` overrides the cap (malformed → default, never unbounded). **No cap override is shipped in the bridge unit or preset, deliberately**: the lever this PR chooses is the declaration, not a bigger budget.
- Bridge: on `SystemPromptOverflowError` no subagent is spawned, result `blocked` / `rollback.reason = system_prompt_overflow`, ledger `outcome: failed`, request left pending (no `handled_` marker, no retry counter), process exits **`EXIT_SYSTEM_PROMPT_OVERFLOW = 4`** → `exit_streak` → bridge health → deploy gate (#1304 keeps 4 fatal at activation; only 3 is transport).
- Every spawn appends a ledger row `{"phase": "system_prompt", "cycle_id", "chars", "cap", "dropped": [...]}`; an overflow appends it with `overflow: true, over_by, sections`. Consumers filter by named phases, so the row is invisible to them.

## Fix — data (ozand/eeebot-self-evolving#186), and the judgement behind it

The reviewer's finding was correct and is what this revision answers: at `1372ddd8` the live `AGENTS.md` has zero markers, so strict mode raises (`exceeds cap by 1243` in the reviewer's environment; by ~11,400 with the host's catalogue and memory). #186 marks **13 loop-authored efficiency sections** droppable (turn budgeting, what to read first, output bounding, test-runner tips — losing one costs turns) and leaves **14 critical**: the five operator sections from #175 (`Identity`, `Charter`, `Cycle contract`, `Working knowledge`, `Script contract`) and the loop-authored rules the gate or the repo's safety depend on (`Forbidden operational paths`, `Staging protocol`, `Execution protocol`, `Cycle termination`, `Immediate skip protocol`, `Out-of-scope handoff`, `Lesson markdown schema`, `Canonical skill path layout`) plus a new `## Prompt budget (droppable sections)` explaining the marker without containing it. Authorship was the first cut; each loop-authored section was then placed by what breaks if the executor never sees it.

Verified with this branch's builder on the marked file: live sizes → **OK at 23,291 / 24,000, 10 droppable sections removed, every critical one kept** (`## Working knowledge` included); reviewer's environment → OK, 2 dropped.

**Marker survival / new sections — decided, not assumed.** The marker is a comment line inside the section body and survives appends and rewrites of the body; only deleting the section removes it. **New sections arrive unmarked, and unmarked means critical** — the correctness-safe default (nothing the operator wrote can vanish) and the overflow-dangerous one. The counterweight is in the instance's own structural tests (#186): a `## ` section that is neither on `CRITICAL_SECTIONS` nor marked fails `test_every_section_is_critical_or_marked_droppable`; a marked critical section fails `test_critical_sections_are_never_marked_droppable`; a critical set that exceeds the live budget fails `test_critical_sections_fit_the_executor_prompt_budget`. A loop-authored addition cannot silently grow the critical set; and if it does slip through, the failure is a loud exit-4 cycle, not a silent cut. (AGENTS.md has been an operator-owned path for the loop since #1193; its last loop commit is 2026-09-01.) This repo pins the literal: `test_marker_literal_is_the_one_the_instance_repo_carries`.

## Tests

`tests/test_context_prompt_fit.py` (9): marker literal pinned; split round-trips every character; strict drops only declared sections, largest first, and keeps a critical section that sits *after* the droppables; strict raises with `sections`/`dropped`/`over_by`; strict never line-trims; non-strict keeps the interactive behaviour; nothing dropped under the cap; env override + malformed fallback; subagent builder strict and exposes `last_prompt_fit`.
`tests/test_bridge_system_prompt_overflow.py` (2), end to end on the #1280 harness: overflow → not spawned, `blocked`/`system_prompt_overflow`, ledger `failed`, `system_prompt` row with `overflow: true`, `rc == EXIT_SYSTEM_PROMPT_OVERFLOW != EXIT_EXECUTOR_LLM_ERROR`, streak failure, no marker, no retry counter; fitting prompt → `rc 0` and a `system_prompt` row listing the dropped section.
Existing `test_context_prompt_budget.py` (#1191) and `test_bridge_executor_llm_error.py` unchanged, green. Rebased onto `ec8721a8` (#1304 merged). Full Windows suite at the final head `3aecf71a` (on `ec8721a8`): `4 failed, 3069 passed, 17 skipped` in 22.5 min — exactly the baseline by name (`test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle` and the three `test_bridge_locking.py::TestAcquireBridgeLock` cases).

## Deployability — the question, answered before merge

Today's live instance `AGENTS.md` carries **zero** droppable markers. Deployed as-is, the first cycle raises `SystemPromptOverflowError`, exits `EXIT_SYSTEM_PROMPT_OVERFLOW = 4`, and #1304's classifier treats 4 as a genuine activation failure and rolls the release back — correctly. Three ways out were on the table; the choice and the reasons:

**Chosen — 1. Markers first, through the instance repo's gated path; this PR merged but not deployed until the host has them.** The declaration is **ozand/eeebot-self-evolving#186** (open). It marks the **13 loop-authored efficiency sections** droppable — turn budgeting, what to read first, output bounding, test-runner tips: losing any of them costs turns, not correctness — and keeps **14 critical**: the five operator sections from #175 (`Identity`, `Charter`, `Cycle contract`, `Working knowledge`, `Script contract`) and the loop-authored rules the gate or the repo's safety depend on (`Forbidden operational paths`, `Staging protocol`, `Execution protocol`, `Cycle termination`, `Immediate skip protocol`, `Out-of-scope handoff`, `Lesson markdown schema`, `Canonical skill path layout`), plus a new `## Prompt budget` section that explains the marker without containing it. Authorship was the first cut; each loop-authored section was then placed by what breaks if the executor never sees it — a gate failure or an unsafe edit → critical; extra turns → droppable. Verified with this branch's builder on the marked file: live sizes → 23,291 / 24,000, 10 sections dropped, every critical one kept; the reviewer's environment → 2 dropped. New sections arrive unmarked (critical) by default; #186's structural tests fail any section that is neither critical-listed nor marked, so the critical set cannot grow silently.

**Rejected — 2. Set `NANOBOT_SYSTEM_PROMPT_MAX_CHARS` high enough to fit today's content.** It turns the refusal into a guard that fires only after the *next* few thousand chars of growth, and nothing re-tightens it: the cap would sit at a number chosen to match one day's file, and the first person to hit it again would raise it again. The override stays available as an operator decision recorded on #1300; it is not the default and this PR ships none.

**Rejected — 3. Classify exit 4 as non-fatal at activation, like 3.** No. Exit 3 is "this one cycle could not reach the gateway" — a transient with a per-cycle base rate, not a property of the release. Exit 4 is "this release, with this instance file, cannot build the prompt it needs" — it will fail every cycle until the data or the code changes, which is exactly what an activation check exists to catch. Agreeing with the reviewer here; the classifier's `failed` for 4 is pinned by test.

**So that the next deployer does not learn this from a journal:**

- The exit-4 rollback now names itself in the deploy output. `lib_bridge_exit.sh` gains `BRIDGE_EXIT_SYSTEM_PROMPT_OVERFLOW=4` and `describe_bridge_exit_status`; the activation `die` line prints `EXIT_SYSTEM_PROMPT_OVERFLOW: the executor prompt cannot hold every critical AGENTS.md section (#1300) — the instance AGENTS.md must carry '<!-- prompt-fit: droppable -->' markers (ozand/eeebot-self-evolving#186) before this release can run`. Driven by test: status 4 still rolls back, and the CRITICAL line carries the remedy.
- The runbook gets **Step 8b**: the one-line host check (`grep -c 'prompt-fit: droppable' …/AGENTS.md` → expect 13) and the order of operations — merge #186, wait one bridge cycle (the bridge fetches the instance `origin/main` at cycle start), re-check, then deploy.
- A pre-flip preflight that runs the new builder against the live workspace was considered and not built: it would duplicate the bridge's workspace path, interpreter and excluded-skill list in bash, and a conservative version (no exclusions) would falsely refuse today's marked file (margin 709 chars vs ~2.2 K of excluded catalogue). The self-naming rollback plus the documented check covers the failure the reviewer is worried about without a second copy of the bridge's configuration.

## Live check owed after deploy (in this order: #186 merged → bridge has fetched → deploy this)

First cycle on the release: a `system_prompt` ledger row (`chars`, `cap`, `dropped` — expect ~10 declared sections), a prompt dump whose `## AGENTS.md` section equals the file minus exactly those sections, `## Working knowledge` present. If #186 was not in place: `outcome: failed, reason: system_prompt_overflow`, `exit_streak.last_exit_status: 4`, deploy rolled back — the fix working, not failing.

### DIFF
--- a/docs/EEEPC_DEPLOY_VERIFY_ROLLBACK_RUNBOOK.md
+++ b/docs/EEEPC_DEPLOY_VERIFY_ROLLBACK_RUNBOOK.md
@@ -149,6 +149,30 @@ Expected:
 
 This proves the verification release can read the same live proof fields coherently.
 
+### Step 8b — precondition for any release at or after #1300 (strict executor prompt)
+
+Since eeebot #1300 the executor's system prompt is built strict: only instance
+`AGENTS.md` sections marked `<!-- prompt-fit: droppable -->` may be dropped
+under the 24,000-char cap, and a prompt that cannot hold every unmarked
+section fails the cycle with exit status 4 (`EXIT_SYSTEM_PROMPT_OVERFLOW`).
+`deploy_release.sh` classifies 4 as a genuine activation failure and rolls the
+release back, naming the cause in its output.
+
+Before deploying such a release, confirm the host's instance workspace already
+carries the markers (ozand/eeebot-self-evolving#186 or later):
+
+```bash
+ssh ozand@eeepc-lan "sudo grep -c 'prompt-fit: droppable' \
+  /var/lib/eeepc-agent/self-evolving-agent/eeebot-self-evolving/AGENTS.md"
+# expected: 13 (or more); 0 means the deploy will roll itself back
+```
+
+Order of operations when both land together: merge the instance markers PR,
+wait one bridge cycle (the bridge fetches the instance `origin/main` at cycle
+start), re-run the check above, then deploy. The alternative lever —
+`NANOBOT_SYSTEM_PROMPT_MAX_CHARS` in the bridge env chain — is an operator
+decision recorded on #1300, not a default.
+
 ## Activation Workflow (optional)
 
 Only do this if the verification release must become the active pinned runtime.
--- a/host/eeepc/scripts/deploy_release.sh
+++ b/host/eeepc/scripts/deploy_release.sh
@@ -664,7 +664,9 @@ if [ "$VERIFY_ONLY" -eq 0 ]; then
     transport)
       echo "[remote] bridge first post-flip cycle exited EXIT_EXECUTOR_LLM_ERROR ($BRIDGE_EXIT_EXECUTOR_LLM_ERROR): transport failure recorded as blocked, not an activation failure (#1303); release stays active" ;;
     *)
-      die "bridge activation failed (restart rc=$BRIDGE_RESTART_RC Result=$BRIDGE_RESULT ExecMainStatus=$BRIDGE_EXIT_STATUS)" ;;
+      # #1300: a known status names itself and its remedy here, so the deployer
+      # reads the cause in the deploy output instead of in the bridge journal.
+      die "bridge activation failed (restart rc=$BRIDGE_RESTART_RC Result=$BRIDGE_RESULT ExecMainStatus=$BRIDGE_EXIT_STATUS) $(describe_bridge_exit_status "$BRIDGE_EXIT_STATUS")" ;;
   esac
 fi
 # --- #1303 bridge activation end ---
--- a/host/eeepc/scripts/lib_bridge_exit.sh
+++ b/host/eeepc/scripts/lib_bridge_exit.sh
@@ -16,6 +16,23 @@
 # Keep the value equal to nanobot/runtime/bridge.py::EXIT_EXECUTOR_LLM_ERROR;
 # tests/test_deploy_activation_exit_class.py pins the two together.
 BRIDGE_EXIT_EXECUTOR_LLM_ERROR=3
+# #1300: the strict ContextBuilder could not fit every critical AGENTS.md
+# section under the system-prompt cap. A property of the release plus the
+# instance's AGENTS.md, not a transient: stays a genuine activation failure
+# and rolls back — but the rollback must name itself and its remedy.
+BRIDGE_EXIT_SYSTEM_PROMPT_OVERFLOW=4
+
+# describe_bridge_exit_status <ExecMainStatus>
+# One line naming a known status and what to do about it; empty for others.
+describe_bridge_exit_status() {
+  case "${1:-}" in
+    "$BRIDGE_EXIT_EXECUTOR_LLM_ERROR")
+      echo "EXIT_EXECUTOR_LLM_ERROR: the executor could not reach its model this cycle (#1280)" ;;
+    "$BRIDGE_EXIT_SYSTEM_PROMPT_OVERFLOW")
+      echo "EXIT_SYSTEM_PROMPT_OVERFLOW: the executor prompt cannot hold every critical AGENTS.md section (#1300) — the instance AGENTS.md must carry '<!-- prompt-fit: droppable -->' markers (ozand/eeebot-self-evolving#186) before this release can run; details in the cycle's system_prompt ledger row and the bridge journal" ;;
+    *) echo "" ;;
+  esac
+}
 
 # classify_bridge_run <systemctl-restart-rc> <Result> <ExecMainStatus>
 # Prints exactly one word:
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -2,6 +2,7 @@
 
 import base64
 import mimetypes
+import os
 import platform
 from pathlib import Path
 from typing import Any
@@ -15,6 +16,24 @@ from nanobot.agent.skills import SkillsLoader
 from nanobot.utils.helpers import build_assistant_message, detect_image_mime
 
 
+class SystemPromptOverflowError(RuntimeError):
+    """The strict builder cannot fit every critical section under the cap (#1300).
+
+    Raised instead of silently trimming: a prompt missing standing
+    instructions is under-specified, and the caller (the bridge) must treat
+    the cycle as failed and say so on a surface that is watched.
+    """
+
+    def __init__(self, *, over_by: int, cap: int, sections: dict[str, int], dropped: list[dict[str, Any]]):
+        self.over_by, self.cap, self.sections, self.dropped = over_by, cap, sections, dropped
+        detail = ", ".join(f"{name}={chars}" for name, chars in sections.items())
+        super().__init__(
+            f"system prompt exceeds cap by {over_by} chars (cap {cap}; sections {detail}; "
+            f"droppable sections already removed: {len(dropped)}) — mark bootstrap sections "
+            f"'{ContextBuilder.DROPPABLE_MARKER}' or raise {ContextBuilder.SYSTEM_PROMPT_CAP_ENV}"
+        )
+
+
 class ContextBuilder:
     """Builds the context (system prompt + messages) for the agent."""
 
@@ -23,31 +42,52 @@ class ContextBuilder:
     BOOTSTRAP_FILES = ["AGENTS.md"]
     _RUNTIME_CONTEXT_TAG = "[Runtime Context — metadata only, not instructions]"
     MAX_SYSTEM_PROMPT_CHARS = 24000
+    #: Operator override of the cap (positive int). The cap is legitimate;
+    #: the budget is the operator's to set, never the builder's to enforce by
+    #: silently choosing which instructions survive.
+    SYSTEM_PROMPT_CAP_ENV = "NANOBOT_SYSTEM_PROMPT_MAX_CHARS"
+    #: A bootstrap ``## `` section containing this marker (anywhere in its
+    #: body) is one the operator allows the cap to drop, whole. Every other
+    #: section is critical and is never dropped in the strict (loop) profile.
+    DROPPABLE_MARKER = "<!-- prompt-fit: droppable -->"
     MAX_MEDIA_BYTES = 2 * 1024 * 1024
 
     def __init__(self, workspace: Path):
         self.workspace = workspace
         self.memory = MemoryStore(workspace)
         self.skills = SkillsLoader(workspace)
+        #: What the last build kept and dropped: ``{"cap", "chars", "strict",
+        #: "dropped": [{"section", "chars", "how"}]}``. Callers record it.
+        self.last_fit: dict[str, Any] | None = None
 
     def build_system_prompt(
         self,
         skill_names: list[str] | None = None,
         excluded_skill_names: list[str] | None = None,
         loop_profile: bool = False,
+        strict: bool | None = None,
     ) -> str:
         """Build the system prompt from identity, bootstrap files, skills, and memory.
 
         Prompt ordering is identity, bootstrap, active skills, skills
-        catalogue, then memory. Under the cap, bootstrap is trimmed first so
-        growth at the bootstrap end cannot evict the later skills and memory
-        sections; any resulting loss is reported by the builder.
+        catalogue, then memory. Under the cap (#1300):
+
+        * ``strict`` (default for the loop profile): only bootstrap ``## ``
+          sections carrying :data:`DROPPABLE_MARKER` may be dropped, whole,
+          largest first. If the rest still does not fit, raise
+          :class:`SystemPromptOverflowError` — never choose survivors by position.
+        * non-strict (interactive sessions): the pre-#1300 behaviour, bootstrap
+          trimmed first at complete-line boundaries, loss logged.
+
+        Either way :attr:`last_fit` records what was dropped.
 
         *excluded_skill_names* is an optional list of skill names to omit from
         the summary (used by the self-evolving loop subagent to suppress
         operator-only builtins such as weather/tmux/clawhub).  Has no effect
         on normal interactive sessions.
         """
+        if strict is None:
+            strict = loop_profile
         sections = [("identity", self._get_identity(loop_profile=loop_profile))]
 
         bootstrap = self._load_bootstrap_files()
@@ -77,7 +117,7 @@ Skills with available="false" need dependencies installed first - you can try in
         memory = self.memory.get_memory_context(loop=loop_profile)
         if memory:
             sections.append(("memory", f"# Memory\n\n{memory}"))
-        return self._fit_system_prompt(sections)
+        return self._fit_system_prompt(sections, strict=strict)
 
     @staticmethod
     def _join_sections(sections: list[tuple[str, str]]) -> str:
@@ -99,17 +139,28 @@ Skills with available="false" need dependencies installed first - you can try in
             used += len(line)
         return "".join(kept)
 
+    def _cap(self) -> int:
+        """The cap: :data:`SYSTEM_PROMPT_CAP_ENV` when it is a positive int, else the class default."""
+        raw = os.environ.get(self.SYSTEM_PROMPT_CAP_ENV, "").strip()
+        try:
+            value = int(raw) if raw else 0
+        except ValueError:
+            value = 0
+        return value if value > 0 else self.MAX_SYSTEM_PROMPT_CHARS
+
     def _trim_section_to_fit(
         self,
         sections: list[tuple[str, str]],
         name: str,
+        cap: int | None = None,
     ) -> tuple[list[tuple[str, str]], int]:
         """Trim one section to the cap and return its dropped character count."""
+        cap = self._cap() if cap is None else cap
         index = next((i for i, (section_name, _) in enumerate(sections) if section_name == name), None)
         if index is None:
             return sections, 0
         current = sections[index][1]
-        if len(self._join_sections(sections)) <= self.MAX_SYSTEM_PROMPT_CHARS:
+        if len(self._join_sections(sections)) <= cap:
             return sections, 0
 
         low, high = 0, len(current)
@@ -118,7 +169,7 @@ Skills with available="false" need dependencies installed first - you can try in
             middle = (low + high) // 2
             candidate = self._trim_lines(current, middle)
             trial = sections[:index] + [(name, candidate)] + sections[index + 1:]
-            if len(self._join_sections(trial)) <= self.MAX_SYSTEM_PROMPT_CHARS:
+            if len(self._join_sections(trial)) <= cap:
                 best = candidate
                 low = middle + 1
             else:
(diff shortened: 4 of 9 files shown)


ozand:
VERDICT: CHANGES REQUESTED — strict overflow handling is correctly loud and durable, but this branch is not deployable against today's instance: the first and every subsequent executor cycle will fail until the markers or cap override ship with it.

## [BLOCKER] The mechanism ships without the data/config required for the live prompt to fit

From the code, not the PR description:

- `SubagentManager._build_subagent_prompt()` always calls `build_system_prompt(..., loop_profile=True)`.
- `ContextBuilder.build_system_prompt()` resolves `strict=None` to `strict = loop_profile`, so the bridge's effective profile is strict.
- In strict mode only bootstrap `## ` sections containing `<!-- prompt-fit: droppable -->` can be removed.
- This PR adds no marker outside tests (`rg` finds only the marker constant), does not modify the instance repository, and does not add a cap override to the bridge unit/config.
- The live instance `AGENTS.md` is **417 lines / 23,449 bytes** and contains **zero markers**.
- None of the effective bridge environment files defines `NANOBOT_SYSTEM_PROMPT_MAX_CHARS`.

I copied the live `AGENTS.md` read-only and built it with this branch's real `ContextBuilder`, the bridge's excluded-skill list, `loop_profile=True`, and the default cap. It raised:

```text
SystemPromptOverflowError
system prompt exceeds cap by 1243 chars
cap=24000
sections: identity=1394, bootstrap=22986, skills_catalogue=849
last_fit: strict=True, dropped=[]
```

That reproduction used an otherwise minimal workspace, so the live instance memory/skills can only maintain or increase the overflow. The historical 35,434-char measurement gives the larger production estimate, but the decisive fact is simpler: **even the minimal strict reconstruction raises today with zero droppable content.**

A loud failure is better than silent instruction loss, but deploying a release known to fail every cycle is not safe. Ship one of the declared operator choices in the same rollout-ready change:

1. instance `AGENTS.md` markers whose measured total makes the real loop prompt fit while preserving required sections; or
2. an explicit reviewed bridge-unit cap override, with the resulting full prompt measured against the executor model context budget.

The marker/config change may live in the appropriate instance/host repository, but PR #1302 must not be merged as deployable until the coupled rollout artifact exists and the release gate can exercise it.

## Exception path audit — correct

If overflow fires, it is caught around the pre-spawn prompt build. No subagent is spawned. The bridge records:

- result status `blocked` with `rollback.reason=system_prompt_overflow`;
- ledger `phase=system_prompt`, `overflow=true`, cap, overshoot, section sizes, and dropped list;
- ledger outcome `failed`, reason `system_prompt_overflow`;
- exit code 4, which the `__main__` recorder persists into `bridge/exit_streak.json`;
- no handled marker and no LLM retry counter.

Thus the process does not disappear without a cycle record. The focused end-to-end tests cover this and pass.

## Marker persistence

The reflection lane can no longer mutate root `AGENTS.md`: #1193 made it operator-owned in both proposer and gate. Therefore an operator-added marker is stable against normal loop edits. The old several-appends-per-day premise no longer applies after #1193; any marker loss would be an operator/repository change and should be reviewed as such.

## Visibility audit

The new `system_prompt` ledger row is a substantial improvement over a journal-only WARNING and makes exact drops queryable. Overflow also reaches the failed outcome, blocked result, exit streak, systemd failure, and deploy health gate.

However, I do not find the claimed bridge health dimension in the current dashboard: `scripts/eeebot_dashboard.py::_build_health_dimensions()` exposes queue, cleanup, host probe/coverage, reward, gate, CPU, memory, and disk, but not `bridge/exit_streak`. The scorecard records the bridge section, and the deploy gate watches the streak, but the operator dashboard does not surface it. That is a non-blocking observability gap relative to the PR wording; the deployment blocker above already prevents rollout.

## Tests

- focused prompt-fit + bridge-overflow suites: **10 passed**;
- CI: green on Python 3.11, 3.12, and 3.13;
- `git diff --check`: passed.

ozand:
## Blocker addressed — data now ships with the mechanism (`7c8f67c1`, rebased on `a64470e5`)

The finding stands exactly as written: strict mode, zero markers in the live `AGENTS.md`, no cap override → `SystemPromptOverflowError` on the first cycle. The mechanism was right; it shipped without its declaration.

**Declaration:** ozand/eeebot-self-evolving#186 — through the instance repo's normal PR path, not a direct commit. It marks 13 loop-authored efficiency sections `<!-- prompt-fit: droppable -->` and leaves 14 critical (the five operator sections from #175 plus the loop-authored rules the gate or the repo's safety depend on), adds a short `## Prompt budget` section that explains the marker without containing it, and preserves the file's CRLF line endings. Rationale per section is in that PR's body and in the updated body here.

**Verified with this branch's builder on the marked file:** live host sizes (identity 1,446 / catalogue 6,951 / memory 4,030) → OK at 23,291 / 24,000 chars, 10 droppable sections removed, every critical section kept including `## Working knowledge`; your environment's sizes (catalogue 849, no memory) → OK, 2 dropped.

**Marker survival / new sections — decided:** unmarked-by-default, i.e. new sections are critical (correctness-safe). The overflow-safe counterweight is in the instance's structural tests (#186): a section that is neither critical-listed nor marked fails, a marked critical section fails, and a critical set exceeding the live budget fails. This repo pins the literal (`test_marker_literal_is_the_one_the_instance_repo_carries`) so the two cannot drift apart silently.

**No cap override is added to the unit or preset**, deliberately: the declaration is the lever. **Sequencing:** #186 merges and is fetched by the bridge before this deploys; the other order fails the first cycle with exit 4 and the gate rolls back — the fix working, not failing.

Only the test pin was added on this branch (`7c8f67c1`); the mechanism commit is unchanged apart from the rebase. Focused files rerun green (13 + 2 + 1).

ozand:
Deployability question answered in the body (section **Deployability — the question, answered before merge**), and enforced at `3aecf71a`: option 1 (markers first via ozand/eeebot-self-evolving#186; this PR merged, not deployed until the host carries them), option 2 rejected (nothing re-tightens the cap), option 3 rejected (exit 4 is release-plus-data, not a transient — agreeing with the classifier and its reviewer). New on the branch: the exit-4 rollback names itself and the remedy in the deploy output (`describe_bridge_exit_status`, driven by test: status 4 still rolls back), and runbook Step 8b carries the one-line host check and the order of operations. Rebased onto `ec8721a8`; 36 focused tests green.

ozand:
Full Windows suite on the final head `3aecf71a` (on `ec8721a8`): `4 failed, 3069 passed, 17 skipped` — exactly the four bridge baseline tests by name. Body updated to match.

ozand:
Verdict: sound — approve.

Reviewed `3aecf71a` on `fix/1300-prompt-fit`, cross-checked against the instance PR `ozand/eeebot-self-evolving#186`. Four things were asked; all four checked out against code and tests, not against the author's description of them.

## 1. Order-of-operations if #1302 deploys before #186

Verified by driving the actual `deploy_release.sh` activation stanza (`tests/test_deploy_activation_exit_class.py::test_activation_overflow_rollback_names_itself_and_the_remedy`, run here, passing): `ExecMainStatus=4` → `classify_bridge_run` returns `failed` (not `transport`, which is reserved for status 3 only) → `die "... $(describe_bridge_exit_status 4)"` → `die` returns 1 → under `set -eE` the `ERR` trap fires → `rollback_remote()` restores `PREV_RELEASE_PATH` and restarts both services. The message and the rollback are the same code path, not two independent claims — the trap fires regardless of what `die`'s argument says, so a self-descriptive message can't ship without the rollback that produces it. `test_activation_genuine_failure_still_rolls_back` separately confirms status 4 is in the "still fatal" set alongside 1/203/signal-9/unreadable, and `test_gate_other_nonzero_exits_still_roll_back` confirms the second rollback point (the post-restart journal gate) also treats status 4 as roll-back-worthy, not transport. Deploying #1302 before #186 lands produces exactly what the author claims: a self-diagnosing, harmless, automatically-reverted failed deploy — confirmed by execution, not by reading the shell.

## 2. What #186 marks droppable — substance check

Read all 13 sections #186 marks `<!-- prompt-fit: droppable -->` in full (not the PR's own summary of them): turn budget checkpoints, early-turn inspection order, early-skip evaluation, git-log-before-edit, bounded reading, command-output bounding, composite-inspection batching, bounded history inspection, demand/reflection lookup paths, secret-scan scoping, targeted-test-validation, doc-only test bypass, unittest-over-pytest. Every one is turn/token economy guidance — how to spend a bounded budget — not a correctness or safety rule. Cross-checked the one case that looks border-line: "Demand and reflection lookup paths" (droppable) ends with "these paths are read-only for you: never edit or stage anything under `state/`" — a real safety constraint. But it is a restatement, not the sole instance: "Forbidden operational paths (control plane)" (unmarked/critical) already states `state/` is forbidden to stage/edit/delete and explicitly cross-references "see the lookup table above." Losing the droppable section loses a table of *where to look things up*, not the only place the *do-not-edit* rule lives. Cross-referenced the 26 pre-existing `## ` headings in the live instance `AGENTS.md` (fetched from `ozand/eeebot-self-evolving@main`, current) against the PR's 14 critical + 13 droppable lists: every existing heading is classified exactly once, no gaps, no duplicates, no critical rule that only exists inside a droppable section. Defensible.

## 3. Marker survival against loop self-edits

Confirmed the mechanism: the marker is a literal HTML comment line inside a section body (`_split_bootstrap_sections` in `context.py` splits on `## ` boundaries and preserves every character — pinned by `test_split_keeps_every_character_and_names_sections`), so appending to or rewriting within a section's existing body leaves the marker intact; only deleting the whole section removes it. New sections arrive **unmarked by default**, and unmarked is the critical set (`_drop_droppable_bootstrap_sections` only removes what carries the exact marker string; everything else is untouchable in strict mode — pinned by `test_strict_never_trims_lines_inside_a_section` and `test_strict_refuses_when_critical_sections_do_not_fit`). That default is correctness-safe (nothing an operator wrote can vanish without the marker being added deliberately) and overflow-dangerous (a growing unmarked file can overflow the cap and halt cycles) — exactly the tradeoff the task asked me to name, not paper over. #186 supplies the counterweight on the instance side, not this PR: `tests/test_agents_structure.py::test_every_section_is_critical_or_marked_droppable` fails closed on any new `## ` heading that is neither on the operator-maintained `CRITICAL_SECTIONS` list nor marked droppable, and `test_critical_sections_fit_the_executor_prompt_budget` fails if the critical set alone would overflow the live budget. That means a loop-authored addition cannot silently grow the critical set past the cap; it can only land unmarked and then fail the instance repo's own test suite, which is a build-time defense, not a runtime one — worth naming as the actual shape of the safeguard rather than "unmarked is safe."

## 4. Ledger row on the normal path, not only on drops

Read the call site directly, not just the drop-branch: `bridge.py`'s spawn path calls `mgr._build_subagent_prompt()` (which always populates `builder.last_fit`, including `dropped: []` when nothing was dropped — `ContextBuilder._fit_system_prompt` sets `self.last_fit` on every call, both the early-return under-cap branch and the strict/non-strict fit branches), then unconditionally checks `isinstance(_prompt_fit, dict)` and writes the `phase: system_prompt` ledger row with `chars`/`cap`/`dropped` — no branch on whether `dropped` is non-empty. Confirmed by `test_fitting_prompt_journals_what_the_cap_dropped` (a case where the prompt fits after dropping one section — still journaled) and independently by re-reading `_fit_system_prompt`'s under-cap early return, which sets `fit = {"cap":..., "chars":..., "strict":..., "dropped": []}` before returning — a cycle that never gets near the cap still produces a queryable row with an empty `dropped` list. This answers the actual concern: the row is written on every spawn attempt regardless of outcome, so it can tell you the cap is near well before it is ever hit, not only after.

## Verification performed

- `python -m pytest tests/test_deploy_activation_exit_class.py tests/test_context_prompt_fit.py tests/test_bridge_system_prompt_overflow.py -v` — 36/36 passed.
- Full suite: `python -m pytest tests/ -q` — 3069 passed, 17 skipped, 4 failed; the four failures are exactly the pre-existing Windows-host baseline (`test_bridge_cycle_tags.py::test_fail_open_tag_failure_never_breaks_a_green_cycle`, three in `test_bridge_locking.py`), unrelated to this change and present on `main` before it.
- `ruff check` on the five touched files: one `F821` in `subagent.py` (pre-existing, confirmed present in `origin/main`'s copy of the file, outside this PR's diff hunk) and pre-existing `I001` import-order findings in `context.py`/`bridge.py` (confirmed identical against `origin/main`). No new lint findings from this diff.
- Read all 13 droppable and 14 critical instance sections in `ozand/eeebot-self-evolving#186` in full, and diffed the classification against the 26 headings in the live `main` `AGENTS.md` — complete, non-overlapping coverage.

## Residual, stated rather than silently accepted

The safety-net for "unmarked-by-default growing the critical set" lives entirely in the instance repo's own test suite (#186), not in this PR's code. That is a reasonable split of responsibility (the operator owns what AGENTS.md says; the harness owns how the cap behaves), but it means #1302 alone does not prevent the instance repo from reintroducing an overflow later — only #186's tests do, and only for as long as they stay in sync with the harness's live section sizes (`identity 1446`, `skills_catalogue 6951`, `memory 4030` are hardcoded into `test_critical_sections_fit_the_executor_prompt_budget`'s budget arithmetic). If those live sizes drift meaningfully, that test's margin narrows or disappears silently until the next drift-triggered failure — worth a cheap follow-up (assert those constants against `ContextBuilder`'s actual current output rather than hardcoding them) but not a blocker for this PR.

PR #1304, opened 2026-09-05
fix(deploy): a transport-failed first cycle is not a broken release (#1303)

### LINKED ISSUE #1303: The #1280 honest exit code makes any transient gateway 404 roll back a deploy
## What happened

Deploy of `a64470e5` on 2026-09-05 06:38 MSK activated cleanly — release extracted, symlink flipped, units synced, dashboard restarted and gated green (`/` HTTP 200, 19,023 bytes) — and was then rolled back to `20260904T212321Z-canonical-632bffbb` because the bridge's start job failed.

The bridge did not fail. One cycle's LLM call did:

```
Spawned subagent [488b5fe4]: selfevo-goal-boo
Subagent [488b5fe4] failed: LLM execution failed: Error calling LLM:
  litellm.NotFoundError: GeminiException - {"detail":"Not Found"}
executor_llm_error: request left pending for retry (1/3)
bridge-result: wrote blocked result to result-llm-proposer-cycle-e539c3594e42.json
bridge-error: recorded structured error to lessons/errors.yaml
eeepc-self-evolving-subagent-bridge.service: Main process exited, code=exited, status=3/NOTIMPLEMENTED
Failed to start eeepc-self-evolving-subagent-bridge.service
```

Every line of that is #1280 (PR #1282) working exactly as designed: the transport failure is recorded as `blocked` with `rollback.reason = executor_llm_error` instead of the old lie of `completed`, the request is left pending with a retry counter, and the process exits `EXIT_EXECUTOR_LLM_ERROR` so `crash_record.record_exit` can advance the failure streak.

The exit code is the point of that fix. It is also, to systemd, a `Type=oneshot` start-job failure — and `deploy_release.sh` treats a failed activation as a broken release.

## Why this matters more than one lost deploy

The `GeminiException - Not Found` class is transient and known: 6 of 93 spawns in an earlier sample, each failing in well under a minute. It is not a property of the release.

So as of #1282 merging, **any deploy whose first post-flip cycle happens to draw a gateway 404 gets rolled back**, regardless of what is in it. The window is roughly one cycle, the base rate is a few percent per cycle, and the failure is silent about its own cause: the deploy log says the bridge failed to start, which reads as "this release is broken".

The immediate cost is already paid: the rollback returned the host to `632bffbb`, which predates #1282, so the loop is once again recording executor failures as `completed`. The fix for the lying is currently un-deployable by the mechanism the fix itself trips.

## The shape of it

Two correct decisions collide. A cycle that could not reach its model **should** exit non-zero — that is what makes the streak and the crash record honest. An activation check **should** fail a release whose service cannot start. Neither is wrong on its own; what is missing is that the deploy cannot distinguish "this release cannot run" from "this one cycle could not reach the gateway".

## What to decide

Pick one deliberately and record why:

1. **Distinguish the exit code at the activation check.** `deploy_release.sh` already knows `EXIT_EXECUTOR_LLM_ERROR`'s meaning is "transport, not release". Treat that specific code as a non-fatal activation outcome, and require a *different* failure — or a repeat — before rolling back. Narrow, and keeps the streak semantics intact.
2. **Retry the activation once before rolling back.** A transient 404 will not usually repeat on the next cycle; a broken release will. Costs one cycle of wall clock and is agnostic to the exit code's meaning.
3. **Separate "the process ran" from "the cycle succeeded"** — the bridge exits 0 for "I ran and recorded an outcome", and the failure streak reads the recorded outcome rather than the exit status. Cleanest conceptually, largest change, and it moves the signal #1280 deliberately put on the exit code.

Do not solve it by making the bridge exit 0 on executor LLM errors. That would restore the exact lie #1280 removed, in a new place.

## Acceptance Criteria

- [ ] A deploy whose first post-flip cycle fails with `EXIT_EXECUTOR_LLM_ERROR` completes instead of rolling back — demonstrated, not asserted, by driving the activation path with that exit code.
- [ ] A deploy whose bridge genuinely cannot start (import error, missing venv, bad unit) still rolls back — asserted by a separate case, because the whole risk of this change is weakening the check that exists to catch real breakage.
- [ ] The chosen arm is recorded in `deploy_release.sh` with the reason, next to the check it changes.
- [ ] Whatever is chosen preserves #1280's semantics: the failure streak still advances on a transport failure, and no cycle records `completed` for a call that never returned.
- [ ] Deployed, and the deploy that carries this change is itself the live verification.

## Note

`a64470e5` is not deployed as a result of this. It carries #1276, #1279, #1278, #1283, #1282, #1275, #1287, #1296, #1297, #1301 and #1293. Redeploying is the obvious next step once this is fixed, and it will be the first real test.

ozand:
Closes #1303. Opened on the author's behalf — deploys are blocked and the branch was ready with an independent review already posted on the issue.

## What broke

Deploying `a64470e5` at 06:38 MSK activated cleanly, then rolled back. The bridge's first post-flip cycle drew `litellm.NotFoundError: GeminiException - {"detail":"Not Found"}` and every line after that was #1282 working as designed — recorded `blocked` with `rollback.reason = executor_llm_error`, request left pending at retry 1/3, process exited `EXIT_EXECUTOR_LLM_ERROR` so the failure streak stays honest.

To systemd that is a `Type=oneshot` start-job failure. `systemctl restart` at `deploy_release.sh:634` runs under `set -eE` with `trap rollback_remote ERR` still installed, so the non-zero return fired the trap and flipped `current` back. The release carrying #1280 became un-deployable by the mechanism its own fix trips, and the host returned to recording executor failures as `completed`.

## The fix, and why this arm

Arm 1 of the three on the issue: distinguish the exit status at the activation check. The restart's failure is captured (`|| BRIDGE_RESTART_RC=$?`) so it no longer reaches the trap; the script then reads `Result` and `ExecMainStatus` and passes both to `classify_bridge_run` in a new `lib_bridge_exit.sh`, sourced remotely from `$RELEASE_DIR` so the classification travels with the release that defines the code.

Two corrections during design are worth recording, because both of my original framings were wrong:

- Arm 1 as first described was not implementable — the script never sees `ExecMainStatus` at that point and had to be made to query it.
- **Arm 2 (retry the activation once) is refuted by the incident itself.** All three attempts on that pending request drew the same 404, at 06:40, 06:43 and 06:48, and it was retired at 3/3. A single retry would also have failed.

Arm 3 (exit 0 and read the recorded outcome) was rejected because it moves the signal #1280 deliberately put on the exit status.

## Independent review

Posted on #1303 by an agent that did not write this and had established the diagnosis. Verdict APPROVE, with the two things that actually mattered checked by execution:

- **Bypassing the trap did not un-cover anything.** Every non-transport tuple reaches an explicit `die` while the trap is still installed; it is cleared only after the classification block. Verified for exit 1, exit 203, prompt-overflow exit 4, signal termination and an unreadable `ExecMainStatus` — the rollback callback fires in all five, and does not fire for the transport tuple.
- **A signal cannot alias exit 3.** `classify_bridge_run` requires `Result=exit-code` as well as status 3: `Result=signal, ExecMainStatus=3` classifies `failed`, as does signal 9.

Activation tests: 23 passed, driving the real stanza under `set -eEuo pipefail` rather than asserting on source. Existing deploy suite: 33 passed.

## Verification owed

The deploy that carries this change is its own live test: it must complete rather than roll back if its first cycle draws a 404, and must still roll back on a genuinely broken release. I will run it and report.

### DIFF
--- a/host/eeepc/scripts/deploy_release.sh
+++ b/host/eeepc/scripts/deploy_release.sh
@@ -19,6 +19,9 @@
 set -euo pipefail
 
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
+# #1303: the bridge's exit-status classes, shared with the remote activation
+# check (which sources the same file from the extracted release).
+. "$(dirname "${BASH_SOURCE[0]}")/lib_bridge_exit.sh"
 HOST="eeepc"
 DRY_RUN=0
 REF=""
@@ -630,9 +633,41 @@ fi
 
 # Ensure bridge service is restarted correctly. Keep the activation rollback trap
 # installed through this restart so a bridge failure restores the previous release.
+#
+# --- #1303 bridge activation begin ---
+# The bridge is Type=oneshot, so `systemctl restart` blocks until its first
+# post-flip cycle ends and returns that cycle's exit status. Since #1280 a cycle
+# whose model call never returned exits EXIT_EXECUTOR_LLM_ERROR (3) so the
+# failure streak stays honest; on 2026-09-05 06:38 a transient gateway 404 in
+# that first cycle turned into "Failed to start … status=3", the ERR trap rolled
+# `a64470e5` back, and the release carrying #1280 became un-deployable by the
+# mechanism its own fix trips.
+#
+# Arm 1 of #1303, deliberately: distinguish the exit status here. Status 3 is
+# "this cycle could not reach the gateway" — the process imported, started, ran
+# a cycle and recorded it `blocked` — not "this release cannot run"; every other
+# non-zero status or signal still is, and still rolls back. Not arm 2 (retry the
+# activation once): it costs a full cycle and a second model call, and the
+# health gate below would still need this same distinction for the journal and
+# streak it reads. Not arm 3 (exit 0 and read the outcome instead): that moves
+# the signal #1280 put on the exit status, and the streak semantics stay exactly
+# as they are. The lib is the single home of the value, pinned to
+# bridge.EXIT_EXECUTOR_LLM_ERROR by test.
 if [ "$VERIFY_ONLY" -eq 0 ]; then
-  sudo systemctl restart eeepc-self-evolving-subagent-bridge.service
+  . "$RELEASE_DIR/host/eeepc/scripts/lib_bridge_exit.sh"
+  BRIDGE_RESTART_RC=0
+  sudo systemctl restart eeepc-self-evolving-subagent-bridge.service || BRIDGE_RESTART_RC=$?
+  BRIDGE_RESULT="$(systemctl show -p Result --value eeepc-self-evolving-subagent-bridge.service 2>/dev/null || true)"
+  BRIDGE_EXIT_STATUS="$(systemctl show -p ExecMainStatus --value eeepc-self-evolving-subagent-bridge.service 2>/dev/null || true)"
+  case "$(classify_bridge_run "$BRIDGE_RESTART_RC" "$BRIDGE_RESULT" "$BRIDGE_EXIT_STATUS")" in
+    ok) ;;
+    transport)
+      echo "[remote] bridge first post-flip cycle exited EXIT_EXECUTOR_LLM_ERROR ($BRIDGE_EXIT_EXECUTOR_LLM_ERROR): transport failure recorded as blocked, not an activation failure (#1303); release stays active" ;;
+    *)
+      die "bridge activation failed (restart rc=$BRIDGE_RESTART_RC Result=$BRIDGE_RESULT ExecMainStatus=$BRIDGE_EXIT_STATUS)" ;;
+  esac
 fi
+# --- #1303 bridge activation end ---
 trap - ERR
 
 REMOTE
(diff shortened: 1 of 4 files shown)


ozand:
Full local suite on the branch, run by me:

```
4 failed, 3056 passed, 17 skipped, 38 warnings in 684.66s

```

Exactly the four-item Windows baseline, by list not by count. No fifth. CI green on 3.11/3.12/3.13.

Merging. The deploy immediately after this is the live test of the change, and I will report what it does rather than assume — including the case that would mean this fix is wrong: a genuinely broken release completing instead of rolling back.

ozand:
## Author's notes for PR #1304 (branch `fix/1303-activation-exit-class` @ `bf1338c7`)

## What tripped

`host/eeepc/scripts/deploy_release.sh:634` restarts the `Type=oneshot` bridge inside the remote activation block while `trap rollback_remote ERR` (#1266) is still armed. `systemctl restart` waits for the start job, so it returns the first post-flip cycle's exit status. Since #1280 a cycle whose model call never returned exits `EXIT_EXECUTOR_LLM_ERROR = 3`; on 2026-09-05 06:38 a transient gateway 404 in that cycle made the restart return non-zero, the trap flipped `current` back to `632bffbb`, and the release carrying #1280 became un-deployable by the mechanism its own fix trips. The post-deploy health gate never entered into the incident — but see below for why it is changed too.

## Which correction changed the choice

**Correction 2** — the exit status was never in hand at the trapped command. The script only saw a non-zero `systemctl restart`. So arm 1 could not be "branch on the code you already have"; it had to become "capture the restart rc, then query `systemctl show -p Result` / `-p ExecMainStatus`, and classify". That is what is implemented. **Correction 3** ruled arm 2 out on the incident's own journal (three retries of the same pending request, same NotFound, 06:40/06:43/06:48). Arm 3 (bridge exits 0, streak reads the recorded outcome) was considered and not taken: it moves the signal #1280 deliberately put on the exit status, it needs `crash_record` and the gate to learn a new source anyway, and the deploy script already reasons about exit statuses (`code=killed, status=(4|6|8|11)`), so teaching it one more code is in kind. `bridge.py` is untouched: the bridge still exits 3 and the streak still advances.

## The change

**`host/eeepc/scripts/lib_bridge_exit.sh` (new, sourced, 30 lines):** `BRIDGE_EXIT_EXECUTOR_LLM_ERROR=3` and `classify_bridge_run <restart-rc> <Result> <ExecMainStatus>` → `ok | transport | failed`. `transport` only when `Result=exit-code` **and** `ExecMainStatus=3`; an unreadable status, a signal, or any other code is `failed`. The constant is pinned to `bridge.EXIT_EXECUTOR_LLM_ERROR` by test.

**Remote activation (the tripping point), `deploy_release.sh` between `# --- #1303 bridge activation begin/end ---`:** sources the lib from the extracted release (`$RELEASE_DIR/host/eeepc/scripts/…`; the release is a full `git archive`, the block already uses that path for `migrate_goal_priorities.py`), runs the restart with `|| BRIDGE_RESTART_RC=$?` so the trap does not fire on the command itself, queries the two properties, and: `ok` → continue; `transport` → log and continue with the release active; anything else → `die`, which under `set -eE` runs the ERR trap → `rollback_remote`. The reason for arm 1 over 2 and 3 is recorded in the comment above the stanza, next to the check it changes.

**Local health gate — the second tripwire.** Once the trap stops rolling back on status 3, the gate's `MAIN_PID_EXIT` grep (`status=[1-9]`) and its exit-streak check would roll the same release back a poll later. Demonstrated, not inferred: the `origin/main` script run through the test harness with the exact line systemd logged (`Main process exited, code=exited, status=3/NOTIMPLEMENTED`) → `FAIL: Bridge process crashed`, rollback, rc 1. So: the journal grep gets `| grep -vE 'code=exited, status=3/'` inside the same remote pipeline (the test replay evaluates that pipeline, so the filter is exercised, not asserted), and a status-3 line is logged once as a transport failure; the streak FAIL branch treats `last_exit_status == 3` as transport (log once, keep polling for a clean run) and, independently, only counts a failure while `consecutive_failures != 0` — the recorder resets it on the next clean run, and transport-then-clean is now the expected post-flip sequence. A real crash in the window is still caught by the journal grep regardless of the streak. Verdicts unchanged: a transport failure alone ends in `UNKNOWN` or `NO-CRASH`, never `CLEAN-EXIT`; a following clean run is `CLEAN-EXIT`.

**What is not weakened:** status 1 (import error), 203 (missing venv/interpreter), 4 (#1302's system-prompt overflow), any signal, or a restart failure with no readable status all still roll back at activation; the same codes still roll back at the gate; a streak failure with any other exit status still rolls back.

## Tests — both acceptance cases driven, not asserted

`tests/test_deploy_activation_exit_class.py` (19):
- `test_lib_constant_is_pinned_to_the_bridge_exit_code` — `3 == bridge.EXIT_EXECUTOR_LLM_ERROR`.
- `test_classify_bridge_run` ×8 — every class, including status-3-without-`exit-code` and empty status → `failed`.
- **Activation, executed from the script's own text** (extracted between the markers) under `set -eEuo pipefail` with the real `die → return 1 → ERR trap → rollback_remote` wiring and a scripted `systemctl`: `test_activation_transport_exit_keeps_the_release` (rc 3 / exit-code / 3 → stanza completes, no rollback); `test_activation_genuine_failure_still_rolls_back` ×5 (1, 203, 4, signal, empty status → `CRITICAL: bridge activation failed`, rollback fired, stanza aborted); `test_activation_clean_exit_continues`.
- **Gate, through the full script with the existing harness:** status-3 journal line → rc 0, no rollback, transport logged, `UNKNOWN`; status 1/4/203 lines → rollback ×3; streak `last_exit_status: 3` → no rollback, waits; streak `last_exit_status: 4` → rollback; transport line + streak reset by a later success → `CLEAN-EXIT`.
- `tests/test_deploy_release.py` fixture copies the lib alongside the script (it copies the script alone into a temp repo); all 36 existing deploy tests unchanged and green, including `test_g` (status 1 rolls back) and `test_o` (non-transport streak failure rolls back).

Full Windows suite at `bf1338c7` (on `a64470e5`): `4 failed, 3056 passed, 17 skipped` in 25 min — exactly the baseline by name (`test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle` and the three `test_bridge_locking.py::TestAcquireBridgeLock` cases).

## Live verification — the deploy that carries this is the test

Redeploy `a64470e5`+this. Expected on the host: activation log line `bridge first post-flip cycle exited EXIT_EXECUTOR_LLM_ERROR (3)…release stays active` **only if** the first cycle draws a 404; otherwise the ordinary path. Either way `current` must point at the new release after the gate, and `exit_streak.json` must still show `last_exit_status: 3` / `consecutive_failures ≥ 1` if a transport failure occurred — the streak advancing is #1280 still working. Quote both on #1303.

PR #1308, opened 2026-09-05
fix(curator): retain complete prompt input batches (#1307)

### LINKED ISSUE #1307: Curator tears JSON input at 48K and silently drops the newest lesson tail
## Problem

The knowledge curator builds its `NEW LESSONS` input as JSON and then applies a raw character slice:

```python
body = json.dumps(lessons, ensure_ascii=False, separators=(",", ":"))
body = body[:MAX_INPUT_CHARS]  # 48,000
```

This can produce invalid JSON and silently remove the tail of the selected lesson batch. It is the non-executor prompt-loss finding from #1305.

## Live evidence (2026-09-05, read-only)

Using the deployed release's own `lessons_after`, `_read_index`, and `_messages` against the live instance repo/state:

```text
selected lessons: 40
serialized NEW LESSONS: 49,617 chars
cap: 48,000 chars
slice applied: yes
json.loads(sliced_body): JSONDecodeError
failure: Unterminated string at char 47,980
```

The cut lands inside the `approach` string of the newest selected tail item. `_messages` sends that torn JSON as prompt text. There is no input-truncation field in curator status/decisions/errors, so operators cannot distinguish a complete lesson batch from a sheared one.

The 2026-09-04 curator run still reported `status: ok`, `processed: 40`, `writes: 2`; successful output therefore does not prove complete input. Existing errors classify malformed **output**, not malformed/truncated input.

## Why it matters

`lessons_after` selects up to 40 chronological post-watermark entries. Raw slicing can partially or wholly remove the newest entries while the curator advances its watermark through the batch. Knowledge can be permanently skipped without a reader-visible unavailable/partial signal.

## Scope

Fix only curator input construction and visibility. Do not broaden #1300 (executor `ContextBuilder`) or #1291 (reflector output parsing).

## Acceptance criteria

- [ ] `NEW LESSONS` remains valid JSON under the 48,000-character budget.
- [ ] Trimming occurs at complete lesson-item boundaries with a declared ordering; no string/object is torn.
- [ ] A test with the live failure shape (40 items, serialized size >48K, newest tail crossing the cap) fails before the fix and proves the retained/dropped item set after it.
- [ ] Curator status or decision diagnostics record whether input was complete/partial and which/count of items were omitted.
- [ ] Watermark advancement cannot silently skip omitted lesson items; either process only retained items or leave omitted items eligible for the next run.
- [ ] Focused and full tests pass within the repository's known platform baseline.

Source: #1305 non-executor prompt-cap audit.

ozand:
Closes #1307.

## Defect

The curator serialized up to 40 lessons and then sliced the JSON string at 48,000 characters. The live 2026-09-05 batch serialized to 49,617 characters; the slice ended inside an `approach` string and was invalid JSON. The run nevertheless reported `status: ok, processed: 40`, and advanced the watermark through all 40 entries.

## Fix

- Fit the lesson batch at complete JSON item boundaries. `_messages` now rejects an unfitted oversized batch instead of slicing it.
- Retain the **oldest contiguous prefix** and defer the newest suffix. This ordering is deliberate: `lessons_after` is chronological and the watermark is one scalar cursor, so only an oldest prefix can be acknowledged without permanently skipping omitted entries. The next run starts at the deferred suffix.
- Report input completeness directly: `input_status`, `selected`, `omitted`, bounded `omitted_ids`, and `selection: oldest-prefix-newest-deferred`.
- `processed` counts only complete items actually sent to the model.
- The watermark advances only through the retained prefix after staging succeeds.

No cap increase was made. No `ContextBuilder` or reflector code was changed.

## Regression evidence

The live-shape test creates 40 complete lesson objects whose serialized body exceeds 48K and proves valid JSON, complete-item retention, prefix-only watermark advancement, and exact deferred-suffix pickup on the next run.

Manual fixture: raw 54,871 chars; sent 46,639 valid JSON chars; selected 34; deferred 6 (`L34`..`L39`); `input_status=partial`.

## Verification

- focused curator tests: **24 passed**;
- Ruff and `git diff --check`: passed;
- full Windows pytest (redirected): **4 failed, 3087 passed, 17 skipped**.

Failures are exactly the established baseline: one `test_bridge_cycle_tags` case and the three `TestAcquireBridgeLock` cases. Branch is based on current `origin/main` (`5a222e78`, including #1298 `8cd5217b`). Host was read-only; no deployment or runtime mutation.

### DIFF
--- a/nanobot/runtime/knowledge_curator.py
+++ b/nanobot/runtime/knowledge_curator.py
@@ -604,9 +604,39 @@ def _parse_output(value: Any) -> list[dict[str, Any]] | None:
     return clean
 
 
+def fit_lessons_to_input_budget(
+    lessons: list[dict[str, Any]],
+    max_chars: int = MAX_INPUT_CHARS,
+) -> tuple[list[dict[str, Any]], dict[str, Any]]:
+    """Keep the oldest complete prefix that fits the JSON budget (#1307).
+
+    ``lessons_after`` is chronological and the watermark is one scalar cursor,
+    so only a contiguous oldest prefix may be acknowledged safely. Newer items
+    are deferred, not dropped: advancing through a newest-only suffix would
+    permanently skip the omitted older entries. Each candidate is serialized
+    whole, so the model never receives torn JSON.
+    """
+    retained: list[dict[str, Any]] = []
+    for item in lessons:
+        candidate = [*retained, item]
+        body = json.dumps(candidate, ensure_ascii=False, separators=(",", ":"))
+        if len(body) > max_chars:
+            break
+        retained.append(item)
+    omitted = lessons[len(retained):]
+    return retained, {
+        "input_status": "complete" if not omitted else "partial",
+        "selected": len(retained),
+        "omitted": len(omitted),
+        "omitted_ids": [_entry_key(item) for item in omitted[:10]],
+        "selection": "oldest-prefix-newest-deferred",
+    }
+
+
 def _messages(lessons: list[dict[str, Any]], index: str, facts: str) -> list[dict[str, str]]:
     body = json.dumps(lessons, ensure_ascii=False, separators=(",", ":"))
-    body = body[:MAX_INPUT_CHARS]
+    if len(body) > MAX_INPUT_CHARS:
+        raise ValueError("curator lesson batch was not fitted to complete items")
     system = (
         "You are the eeebot knowledge curator. Return ONLY a JSON array. "
         "Each item must be one of: "
@@ -1692,6 +1722,26 @@ def run_curation(
             "last_success_ts": _now(), **result["stages"],
         })
         return result
+    # The scalar watermark can safely acknowledge only a contiguous oldest
+    # prefix. Newer items that do not fit are explicitly deferred and remain
+    # eligible after the watermark advances through the selected prefix.
+    entries, input_fit = fit_lessons_to_input_budget(entries, MAX_INPUT_CHARS)
+    if not entries:
+        curation_stage = {
+            "status": "partial",
+            "processed": 0,
+            "writes": 0,
+            "staged": [],
+            **input_fit,
+        }
+        result = {"ok": True, "processed": 0, "writes": 0, "staged": [], "stages": {
+            "reflector_mint": reflector_stage,
+            "curation": curation_stage,
+        }}
+        _atomic_json(state_dir / "curator" / "status.json", {
+            "last_success_ts": _now(), **result["stages"],
+        })
+        return result
     try:
         model = resolve_model("curator", strip_openai=True)
         messages = _messages(entries, _read_index(workspace), "")
@@ -1745,10 +1795,15 @@ def run_curation(
         staged_paths = [e["path"] for e in staged]
         unsupported = sum(1 for e in staged if e.get("overlap_flag"))
         curation_stage = {
-            "status": "ok", "processed": len(entries), "writes": writes, "staged": staged_paths,
+            "status": "ok" if input_fit["input_status"] == "complete" else "partial",
+            "processed": len(entries),
+            "writes": writes,
+            "staged": staged_paths,
+            **input_fit,
         }
         result_dict: dict[str, Any] = {
             "ok": True, "processed": len(entries), "writes": writes, "staged": staged_paths,
+            "input_status": input_fit["input_status"], "omitted": input_fit["omitted"],
             "stages": {"reflector_mint": reflector_stage, "curation": curation_stage},
         }
         _atomic_json(state_dir / "curator" / "status.json", {
--- a/tests/test_issue_1216_curator_signal.py
+++ b/tests/test_issue_1216_curator_signal.py
@@ -92,6 +92,8 @@ def test_run_curation_reports_fact_stage_success(tmp_path: Path) -> None:
     assert result["stages"]["curation"] == {
         "status": "ok", "processed": 1, "writes": 1,
         "staged": ["memory/facts/insight.md"],
+        "input_status": "complete", "selected": 1, "omitted": 0,
+        "omitted_ids": [], "selection": "oldest-prefix-newest-deferred",
     }
 
 
--- a/tests/test_knowledge_curator.py
+++ b/tests/test_knowledge_curator.py
@@ -7,7 +7,9 @@ from pathlib import Path
 from nanobot.runtime.knowledge_curator import (
     _ACTION_INDEX_SEGMENTS,
     _fact_path,
+    _messages,
     clear_staged_manifest,
+    fit_lessons_to_input_budget,
     lessons_after,
     load_staged_manifest,
     migrate_loose_lessons,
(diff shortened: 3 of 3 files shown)


ozand:
Verdict: sound — approve.

Reviewed `5f0146b3` on `fix/1307-curator-input-boundary` against `origin/main` (`5a222e78`). All four requested checks were driven, not read.

## 1. The watermark — driven, not read

`run_curation` reassigns `entries, input_fit = fit_lessons_to_input_budget(entries, MAX_INPUT_CHARS)` *before* the watermark is computed later as `_entry_key(entries[-1])` — so the watermark is always keyed off the last item actually in the retained (sent-to-model) list, never off the original pre-fit batch. Confirmed by execution, not inspection: ran `test_partial_batch_watermark_leaves_deferred_suffix_for_next_run` (passes), then independently drove `run_curation` twice against a monkeypatched `lessons_after` and a deciding LLM stub outside the test file — first call processes a prefix, `watermark.json.last_processed_id` lands on the *last retained* id (not the batch's last id), second call's `lessons_after` fake correctly starts after that watermark and receives exactly the deferred suffix, second watermark lands on the true final id. No permanent skip.

Also drove the edge the tests don't cover: a *single* lesson whose own serialized body alone exceeds 48K. `fit_lessons_to_input_budget` returns `retained=[]`, `run_curation` takes the `if not entries:` early-return branch, reports `status: partial, selected: 0, omitted: 1, omitted_ids: ["L00"]`, and — confirmed by reading the code path — never reaches `_atomic_json(wm_path, ...)`, so the watermark file is never written/advanced. That's a stall, not silent loss: the run is honest every time it retries, but it will retry forever on that one oversized lesson with no code path that shrinks or skips it. Worth naming as a residual: not a regression this PR introduces (the old code would have torn it into invalid JSON and advanced past it, which is worse), but not fully closed either — a lesson large enough to individually blow the budget currently blocks the queue rather than being handled.

## 2. Which end is dropped — justified, correctly

Confirmed the new order keeps the **oldest** contiguous prefix and defers the **newest** suffix — the reverse of the old `body[:MAX_INPUT_CHARS]`, which (since lessons are serialized in chronological order) discarded the newest tail by slicing the whole document from the front, but by cutting mid-string discarded whichever field happened to sit at byte 48,000, which in the live failure was inside the newest item's `approach` string. The new ordering is stated and justified in-line (`fit_lessons_to_input_budget`'s docstring and the `run_curation` call-site comment): "the watermark is one scalar cursor, so only a contiguous oldest prefix may be acknowledged safely." That is the correct constraint, not an arbitrary pick — a cursor over a chronologically-sorted stream can only mean "processed through this point," so acknowledging a newest-first subset while an older item sits unprocessed behind it would either advance past the older item (loss) or require a non-scalar/set-based watermark (a bigger change this PR correctly declined to make). The tradeoff — most-recent lessons are the ones delayed under sustained overflow — is named, not hidden, and is the direction implied by the data structure actually in use.

## 3. Honesty of the count — verified by reading the reassignment, not the field name

`entries` is reassigned to the fitted (retained) list before `processed: len(entries)` is computed later in the same function — confirmed by tracing the variable, not assuming the field means what it says. Cross-checked against `test_issue_1216_curator_signal.py`'s updated fixture: a fully-fitting single-lesson batch now asserts the full explicit tuple `input_status: complete, selected: 1, omitted: 0, omitted_ids: [], selection: oldest-prefix-newest-deferred` on the *normal* path, not only the degraded one — so a caller reading `status.json` sees completeness stated positively every run, never only on failure. `processed`, `selected` and `writes` are three different, non-conflatable numbers now (items sent vs. items the model actually acted on vs. items durably staged) — checked they don't silently collapse into each other anywhere in the diff.

## 4. Regression test reproduces the exact live shape — reproduced independently, twice

Ran the PR's own two new tests against `origin/main`'s `knowledge_curator.py` (not against a hypothetical): `test_oversized_lesson_batch_keeps_valid_oldest_prefix_and_defers_newest` and `test_partial_batch_watermark_leaves_deferred_suffix_for_next_run` fail to even *import* against pre-fix code (`ImportError: cannot import name 'fit_lessons_to_input_budget'`) — proof the tests exercise genuinely new code, not a vacuous assertion that happens to pass either way. Separately, outside the test file, reproduced the raw live defect directly: 40 items at the test's sizing serialize to 54,871 chars; `body[:48_000]` (the exact old expression) yields `json.loads` failure `Unterminated string starting at: line 1 column 46698` — the same failure class the issue reports (`Unterminated string at char 47,980` on the real 49,617-char batch; the discrepancy in cut position is only the difference between the live batch and the test fixture's approach-string padding, not a mismatch in mechanism). Confirms the test's 40-items/>48K/newest-crosses-cap shape is not a stand-in for the live failure — it drives the identical code path.

## MAX_INPUT_CHARS — confirmed unmoved

`grep`-verified: `MAX_INPUT_CHARS = 48_000` is unchanged in this diff; the only line touching the name is `fit_lessons_to_input_budget`'s signature (`max_chars: int = MAX_INPUT_CHARS`) and its use as an argument, plus `_messages`' new `if len(body) > MAX_INPUT_CHARS: raise` guard replacing the old slice — the cliff is fitted at, not moved.

## Verification performed

- `python -m pytest tests/test_knowledge_curator.py tests/test_issue_1216_curator_signal.py -v` — 24/24 passed, matching the PR's claim.
- Confirmed non-vacuity: swapped `knowledge_curator.py` for `origin/main`'s copy and re-ran the two new tests — both fail at collection (`ImportError`), confirming they exercise the fix's new code rather than passing coincidentally either way. Restored the fixed file afterward; `git diff --quiet` confirmed a clean restore.
- Reproduced the raw pre-fix defect directly (independent of the test suite): 40-item batch → 54,871-char JSON → `body[:48_000]` → `json.JSONDecodeError: Unterminated string`, matching the issue's failure class.
- Drove `run_curation` end-to-end twice with a scripted LLM and a monkeypatched `lessons_after`, outside the shipped test file, to independently confirm cross-run watermark/deferred-suffix behavior.
- Drove the single-oversized-lesson edge case (not covered by the PR's own tests) to confirm it fails safe (partial status, no crash, no watermark write) rather than silently.
- Full suite: `python -m pytest tests/ -q` — 3087 passed, 17 skipped, 4 failed; the four are exactly the established Windows baseline (`test_bridge_cycle_tags`, three `TestAcquireBridgeLock` cases), matching the PR's own reported numbers exactly.
- `ruff check` on the three touched files — clean. `git diff --check` — clean.
- `grep`-confirmed `MAX_INPUT_CHARS` value is unchanged at 48,000.

## Residual, stated rather than silently accepted

A single lesson whose own serialization exceeds the 48K budget stalls the curator queue indefinitely: `fit_lessons_to_input_budget` returns an empty retained list, the run reports `status: partial, selected: 0` honestly every time, and the watermark correctly never advances past it — but nothing shrinks, splits, or skips that one oversized entry, so it blocks every subsequent lesson behind it in the chronological queue until it is edited externally. Better than the old behavior (which would have torn it and silently advanced past it), but not a closed loop. Worth a follow-up issue, not a blocker here.

PR #1311, opened 2026-09-05
fix(curator): quarantine a lesson larger than the input budget instead of stalling (#1309)

### LINKED ISSUE #1309: A single lesson larger than the curator's input budget blocks the queue forever
## Context (self-contained)

#1307 (PR #1308, merged) fixed the curator tearing its JSON input: `NEW LESSONS` is now trimmed at complete item boundaries, keeping the oldest contiguous prefix and deferring the newest suffix, and the watermark is keyed off the last **retained** item so a partial batch can no longer advance past entries the model never saw.

That is the right shape, and the ordering is correctly justified in-line: the watermark is one scalar cursor over a chronologically-sorted stream, so only a contiguous oldest prefix can be acknowledged safely.

The reviewer drove an edge the tests do not cover, and found a residual the fix does not close.

## The residual

**A single lesson whose own serialized body exceeds `MAX_INPUT_CHARS` (48,000) blocks the queue permanently.**

Traced through the merged code:

- `fit_lessons_to_input_budget` returns `retained=[]` — no prefix fits, because the first item alone is over budget.
- `run_curation` takes the `if not entries:` early return and reports `status: partial, selected: 0, omitted: 1, omitted_ids: ["<id>"]`.
- It never reaches `_atomic_json(wm_path, ...)`, so the watermark is not written and does not advance.

Every subsequent run selects the same oversized lesson first, retains nothing, and stops. There is no code path that shrinks the item, skips it, or escalates.

**This is not a regression.** The old behaviour was worse: it tore the item into invalid JSON and advanced the watermark past it, losing the lesson silently. The new behaviour is honest on every run — `status: partial` with the blocking id named — and loses nothing. But an honest permanent stall is still a stall: the curator stops distilling anything at all until someone notices and intervenes by hand.

## What makes it worth fixing rather than watching

Nothing in the loop currently reads `status: partial` from the curator and acts on it. The run reports the truth into a file, and the same file said `status: ok` for weeks while the input was malformed and nobody read it either. A stall that is only visible to a reader who already suspects it is close to a stall that is invisible.

## What to decide

Pick deliberately and record the reason:

1. **Skip and quarantine.** Advance the watermark past the oversized item, record it to a quarantine list with its id and size, and continue. The lesson is not lost — it is set aside and countable — and the queue moves. Requires deciding who, if anyone, ever revisits the quarantine.
2. **Truncate the item's own fields, not the document.** Trim the largest field of that single lesson to fit, mark the entry as content-truncated in what the model sees, and process it. Keeps the lesson in the stream at reduced fidelity; needs the truncation to be visible to the model, not silent.
3. **Escalate rather than continue.** Leave the stall, but make it loud — a demand item, a scorecard signal, or a failure status that something actually reads. Correct only if a reader exists; if not, this is arm 1 without the benefit.

The measurement that should inform the choice: how large is the largest lesson ever written, and how close does it get to 48,000? If nothing has ever come within half the budget this is theoretical and arm 3 is enough; if items regularly reach 20-30K it is a matter of time and arm 1 or 2 is warranted.

## Acceptance Criteria

- [ ] The maximum serialized lesson size across the retained history is measured and quoted, so the arm is chosen against data rather than intuition.
- [ ] A single over-budget lesson no longer blocks subsequent lessons from being curated — demonstrated by a test that fails on current `main`.
- [ ] Whatever is set aside or truncated is countable from the durable record, with its id, not merely absent.
- [ ] The chosen arm and its reason are recorded next to the code that implements it.

ozand:
Closes #1309.

## The measurement, first

Deployed `knowledge_curator.lessons_after(workspace, "", limit=1_000_000, state_dir)` run read-only on the host as `eeepc-agent` (release `20260904T212321Z-canonical-632bffbb`, 2026-09-05 ~09:00 MSK), each item serialized exactly as `fit_lessons_to_input_budget` does (`json.dumps(item, ensure_ascii=False, separators=(",", ":"))`):

```
items in stream: 993 | max 2113 | p50 738 | p95 1736 | p99 1883 | mean 962
count >=12000: 0 | >=24000: 0 | >=48000: 0
top: LESS-20260904-90931e43 2113 (largest field solution=438) · REFL-ead413e7a8d9-0 2046 (result=790) ·
     LESS-20260904-d44ed220 2031 · REFL-137ea494d7ec-0 1990 · ERR-2026-06-20-001 1981 (fix_applied=653) …
curator status.json now: status ok, processed 40, writes 2 · watermark REFL-53e67c3f1d78 @ 2026-09-04T21:01:48Z
```

The largest item ever written is **4.4% of the 48,000-char budget**; nothing has come within a tenth of it. The stall is real but theoretical today, so the cheapest arm that closes it correctly wins.

## Arm chosen: 1 — skip and quarantine

Reason recorded in the code next to the change (`run_curation`, `_quarantine_oversized_head`). Not **arm 2** (truncate the item's fields): more code, and the model receives a lossy item whose truncation it has to be told about. Not **arm 3** (escalate): nothing reads the curator's `status.json` — the same file said `ok` for weeks over torn input — so an escalation would be arm 1 without the benefit. Who revisits the quarantine: nobody today, by design; `curator/quarantine.jsonl` is the countable record (id, size, budget, reason, timestamp) that the operator or a future reader can consume.

## What changes

- `_quarantine_oversized_head(lessons, max_chars)` splits off **leading** items that exceed the budget on their own. Only the head of the chronological stream can be passed by the scalar watermark without skipping unseen entries, so only leading items are quarantined; an oversized item behind a retained prefix is deferred by #1307's rule and quarantined when it becomes the head next run. A head item with **no id or timestamp** cannot be acknowledged by a scalar cursor at all: it is left in place and named in `status.json` `blocked_reason` — the one honest stall that remains, now with a reason instead of a bare `partial`.
- `run_curation`: for each quarantined item, one row in `curator/quarantine.jsonl` (`timestamp, lesson_id, chars, budget, reason`); the watermark is advanced past them **before** the LLM call (so a model failure cannot re-quarantine them); `status.json`, `watermark.json` (`quarantined: [ids]`) and the run result carry `quarantined` / `quarantined_ids`. If only oversized items were present, `status: quarantined` with no LLM call. Fitting then proceeds exactly as #1307 left it.
- No change to `fit_lessons_to_input_budget`, `_messages`, or the state-path registry (`curator/` is already declared by its live writers; granularity is the segment).

## Tests (`tests/test_knowledge_curator_quarantine.py`, 6 — all six fail on `main`, verified)

- helper takes only the oversized head, records the serialized size, defers an oversized item behind a fitting one
- oversized head quarantined, the two fitting lessons reach the model in one batch, watermark at the last retained, quarantine row has id/chars/budget/reason, status `ok` with `quarantined 1`, second run has nothing to do
- only-oversized stream → watermark advances, quarantine rows, `status: quarantined`, no LLM call, no re-quarantine on the next run
- quarantine row and watermark survive an LLM failure (watermark past the quarantined item, not past the unseen one)
- keyless oversized head → no watermark, no quarantine row, `status: partial` with `blocked_reason`
- oversized item behind a fitting prefix is deferred (#1307), then quarantined on the next run

Existing `tests/test_knowledge_curator.py` unchanged, 18 green. Ruff clean.

Full Windows suite at `7737c94a` (on `1f4ba878`): `4 failed, 3093 passed, 17 skipped` in 21 min — exactly the baseline by name (`test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle` and the three `test_bridge_locking.py::TestAcquireBridgeLock` cases).

None expected to fire: no live item is near the budget. Observable if it ever does: a row in `state/curator/quarantine.jsonl` with the lesson id and size, `status.json.curation.quarantined ≥ 1`, and the watermark past that id.

### DIFF
--- a/nanobot/runtime/knowledge_curator.py
+++ b/nanobot/runtime/knowledge_curator.py
@@ -633,6 +633,38 @@ def fit_lessons_to_input_budget(
     }
 
 
+def _quarantine_oversized_head(
+    lessons: list[dict[str, Any]],
+    max_chars: int = MAX_INPUT_CHARS,
+) -> tuple[list[dict[str, Any]], list[tuple[dict[str, Any], int]], str]:
+    """Split off leading items that exceed ``max_chars`` on their own (#1309).
+
+    Only the head of the chronological stream can be passed over by the scalar
+    watermark without skipping unseen entries, so only leading oversized items
+    are quarantined; an oversized item behind a retained prefix is deferred by
+    :func:`fit_lessons_to_input_budget` and becomes the head next run. An
+    oversized head item with no id or timestamp cannot be acknowledged by a
+    scalar cursor at all; it is left in place and named in ``blocked_reason``
+    — the one honest stall that remains.
+
+    Returns ``(remaining, [(item, serialized_chars), ...], blocked_reason)``.
+    """
+    quarantined: list[tuple[dict[str, Any], int]] = []
+    index = 0
+    while index < len(lessons):
+        size = len(json.dumps([lessons[index]], ensure_ascii=False, separators=(",", ":")))
+        if size <= max_chars:
+            break
+        if not _entry_key(lessons[index]):
+            return lessons[index:], quarantined, (
+                f"head item is {size} chars over a {max_chars}-char budget and has no id or "
+                "timestamp, so the scalar watermark cannot pass it"
+            )
+        quarantined.append((lessons[index], size))
+        index += 1
+    return lessons[index:], quarantined, ""
+
+
 def _messages(lessons: list[dict[str, Any]], index: str, facts: str) -> list[dict[str, str]]:
     body = json.dumps(lessons, ensure_ascii=False, separators=(",", ":"))
     if len(body) > MAX_INPUT_CHARS:
@@ -1722,13 +1754,42 @@ def run_curation(
             "last_success_ts": _now(), **result["stages"],
         })
         return result
+    # #1309: an item that is over the budget on its own can never be the head
+    # of a retained prefix, so under #1307's rule it would stall the stream
+    # forever (honestly, run after run). Quarantine it instead: record it with
+    # its id and size, advance the watermark past it, continue. Arm 1 of the
+    # three in #1309, chosen against the measurement: the largest item in the
+    # 993-item live stream on 2026-09-05 was 2,113 chars (4.4% of 48,000, p99
+    # 1,883), so this is a guard against a theoretical shape and the cheapest
+    # correct one wins. Not arm 2 (field truncation: more code, the model sees
+    # a lossy item) and not arm 3 (escalate: nothing reads the curator's
+    # status file — the same file said `ok` for weeks over torn input — so an
+    # escalation would be arm 1 without the benefit). The quarantine file is
+    # the countable record; nothing revisits it today, by design — the
+    # operator or a future reader can.
+    entries, quarantined, blocked_reason = _quarantine_oversized_head(entries, MAX_INPUT_CHARS)
+    quarantine_report: dict[str, Any] = {"blocked_reason": blocked_reason} if blocked_reason else {}
+    if quarantined:
+        for item, size in quarantined:
+            _append_jsonl(state_dir / "curator" / "quarantine.jsonl", {
+                "timestamp": _now(), "lesson_id": _entry_key(item), "chars": size,
+                "budget": MAX_INPUT_CHARS, "reason": "single item exceeds MAX_INPUT_CHARS",
+            })
+        last_q = _entry_key(quarantined[-1][0])
+        _atomic_json(wm_path, {"last_processed": last_q, "last_processed_id": last_q, "timestamp": _now(),
+                               "quarantined": [_entry_key(item) for item, _ in quarantined]})
+        quarantine_report = {"quarantined": len(quarantined),
+                             "quarantined_ids": [_entry_key(item) for item, _ in quarantined[:10]]}
     # The scalar watermark can safely acknowledge only a contiguous oldest
     # prefix. Newer items that do not fit are explicitly deferred and remain
     # eligible after the watermark advances through the selected prefix.
     entries, input_fit = fit_lessons_to_input_budget(entries, MAX_INPUT_CHARS)
+    input_fit.update(quarantine_report)
     if not entries:
         curation_stage = {
-            "status": "partial",
+            # A head item without any key cannot be passed by a scalar
+            # watermark; that is the one remaining honest stall, and it is named.
+            "status": "quarantined" if quarantined and not input_fit["omitted"] else "partial",
             "processed": 0,
             "writes": 0,
             "staged": [],
--- a/tests/test_knowledge_curator_quarantine.py
+++ b/tests/test_knowledge_curator_quarantine.py
@@ -1,0 +1,144 @@
+"""#1309: a single lesson larger than the curator's input budget must not block
+the queue forever.
+
+Under #1307 the curator keeps the oldest complete prefix that fits and keys the
+watermark off the last retained item. An item that is over the budget on its
+own can never head a retained prefix, so `fit_lessons_to_input_budget` returned
+`[]`, `run_curation` took the empty early return, and the watermark never
+moved: every run picked the same item and stopped, honestly. Measured on the
+live stream on 2026-09-05: 993 items, max 2,113 chars against 48,000 — a guard
+against a theoretical shape, so the cheapest correct arm: quarantine the item
+(durable row with id and size), advance the watermark past it, continue.
+"""
+from __future__ import annotations
+
+import json
+
+from nanobot.runtime import knowledge_curator as curator
+from nanobot.runtime.knowledge_curator import run_curation
+
+BUDGET = curator.MAX_INPUT_CHARS
+
+
+def _lesson(i: int, size: int) -> dict:
+    return {"id": f"L{i:02d}", "timestamp": f"2026-09-05T00:{i:02d}:00Z", "approach": "x" * size}
+
+
+def _wire(monkeypatch, lessons: list[dict]):
+    """A watermark-honouring lessons_after over an in-memory stream, and an LLM
+    that acknowledges every item it is shown and records what it saw."""
+    def fake_lessons_after(_workspace, watermark, *, limit, state_dir):
+        start = 0
+        if watermark:
+            start = next((i + 1 for i, item in enumerate(lessons) if curator._entry_key(item) == watermark), 0)
+        return lessons[start:start + limit]
+
+    seen: list[list[str]] = []
+
+    def llm(messages, _model):
+        body = messages[1]["content"].split("NEW LESSONS:\n", 1)[1].split("\n\nKB INDEXES:", 1)[0]
+        batch = json.loads(body)
+        seen.append([item["id"] for item in batch])
+        return json.dumps([{"action": "unimportant", "lesson_id": item["id"], "reason": "t"} for item in batch])
+
+    monkeypatch.setattr(curator, "lessons_after", fake_lessons_after)
+    monkeypatch.setattr(curator, "promote_reflector_recommendations_to_v2", lambda *_a, **_k: {})
+    return llm, seen
+
+
+def _read(state, name):
+    return json.loads((state / "curator" / name).read_text(encoding="utf-8"))
+
+
+def test_quarantine_helper_takes_only_the_oversized_head():
+    big, small = _lesson(0, BUDGET + 10), _lesson(1, 100)
+    remaining, quarantined, blocked = curator._quarantine_oversized_head([big, big | {"id": "L09"}, small, _lesson(3, BUDGET + 10)], BUDGET)
+    assert [item["id"] for item, _ in quarantined] == ["L00", "L09"]
+    assert all(size > BUDGET for _, size in quarantined)
+    assert [item["id"] for item in remaining] == ["L01", "L03"], "an oversized item behind a fitting one is deferred, not quarantined"
+    assert blocked == ""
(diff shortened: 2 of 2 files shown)


ozand:
Verdict: sound — approve.

Reviewed `7737c94a` on `fix/1309-oversized-lesson` against `origin/main` (`1f4ba878`, which already carries #1307/#1308). Checked in the requested order, driven rather than read.

## 1. Quarantine is a durable, enumerable record — not quiet deletion

`_append_jsonl` fsyncs on every write (`fh.flush(); os.fsync(fh.fileno())`) — the same primitive already used for `curator/decisions.jsonl` and `curator/errors.jsonl`, both of which are precedented as append-only journals nobody rotates or truncates in this codebase; I grepped `host/eeepc/` and `nanobot/` for any rotation/pruning logic touching `curator/*.jsonl` and found none — these files grow forever, same as their two siblings, so `quarantine.jsonl` inherits exactly the durability profile already accepted for the curator's other two audit trails, not a weaker one.

Each row carries `timestamp, lesson_id, chars, budget, reason` — enough to identify *which* lesson was set aside, *how big* it was, and *against what budget*, not just a count. I drove the multi-item case directly (see §2) and confirmed the row for a real oversized item reads `{"lesson_id": "M01", "chars": 48563, "budget": 48000, "reason": "single item exceeds MAX_INPUT_CHARS", "timestamp": "..."}`.

**But you were right to ask "who reads it," and the honest answer is: nobody today.** I grepped the entire repo (`nanobot/`, `scripts/`, `host/eeepc/`) for `quarantine` outside `knowledge_curator.py` and its own test file — zero hits. `scripts/eeebot_dashboard.py` does not read `curator/status.json` or any `curator/*.jsonl` file at all (confirmed by grep, zero matches for `curator` in that script). The PR's own body states this plainly rather than hiding it ("nobody [revisits it] today, by design... the operator or a future reader can") — that is the same disclosure standard #1307/#1308 met, and it is the correct answer to your question, not a defense that dodges it. This is a countable file, not a read one, and the PR says so. I'd flag as a cheap, non-blocking follow-up: a one-line addition to the dashboard or `status.json` summary that surfaces `quarantined_ids` count would close the "nobody reads it" gap for near-zero cost, since the plumbing (the `quarantined`/`quarantined_ids` fields on `status.json`) is already there.

## 2. Watermark advances past exactly the quarantined item — driven independently, mid-batch

Wrote a scenario not in the PR's own test file: five lessons `M00..M04` in one `lessons_after` batch, where `M01` alone exceeds the budget and `M00`, `M02`, `M03`, `M04` are all small and fit together. Ran `run_curation` twice with a real (stubbed) LLM, outside the shipped tests:

- **Run 1**: `M00` is curated alone (`seen[0] == ["M00"]`), watermark lands on `M00`. `M01` is *not yet* quarantined at this point — it's still the deferred tail per #1307's rule (`omitted_ids: ["M01","M02","M03","M04"]`), confirming quarantine only fires once an oversized item becomes the *head* of a batch, not mid-batch, which is the correct and stated scope (`_quarantine_oversized_head` docstring: "an oversized item behind a retained prefix is deferred... and becomes the head next run").
- **Run 2**: `M01` is now the head, gets quarantined (`quarantine.jsonl` row: `lesson_id: M01, chars: 48563, budget: 48000`), and — the check that matters — `M02, M03, M04` (everything *after* the quarantined item) are curated in the same run (`seen[1] == ["M02","M03","M04"]`), watermark lands on `M04`, not stuck at `M01` and not skipping past `M04` into unseen territory. `status.json` correctly reports `quarantined: 1, quarantined_ids: ["M01"]` alongside `processed: 3`.

This confirms the cursor moves exactly one item at a time through quarantine and keeps moving correctly through the subsequent curated batch — no over-advance, no stall transferred downstream.

## 3. The measurement was real, not asserted — reproduced independently on the host

Re-ran the PR's exact stated methodology on the live host, read-only, using the current release (not the one the PR cites — the deploy since then moved `current` to `20260905T042713Z-canonical-ec8721a8`, so this is a fresh independent run, not a repeat of their session): `lessons_after(workspace, "", limit=1_000_000, state_dir)`, each item serialized with the identical `json.dumps(item, ensure_ascii=False, separators=(",", ":"))` used by `fit_lessons_to_input_budget`.

Result: **997 items, max 2,113 chars, p99 1,883, zero items ≥ 12,000/24,000/48,000** — matching the PR's cited `993 items, max 2,113` almost exactly (the 4-item drift is fully explained by ~35 minutes of elapsed time and new lessons landing between their measurement and mine; the max size and top-5 largest items are effectively identical, `LESS-20260904-90931e43` at 2,113 chars in both runs). The largest lesson ever produced is **4.4% of the 48,000-char budget** — the PR's claim is genuine, reproducible, and not stale.

Given that measurement, the issue's own rule ("if nothing has ever reached half the budget, the cheapest arm wins") points at exactly arm 1. The PR names the two arms it rejected and why (arm 2 truncation = more code + a lossy item reaching the model; arm 3 escalation = nothing reads `status.json` today, so escalation would just be arm 1 with extra ceremony) — that reasoning is consistent with what I independently found in §1 (nothing does read it), so the rejection of arm 3 isn't a convenient excuse, it's accurate.

## 4. `MAX_INPUT_CHARS` unchanged; normal path unmodified

`grep`-confirmed: `MAX_INPUT_CHARS = 48_000`, untouched — the diff only *passes* the constant into the new helper, never redefines it. `fit_lessons_to_input_budget` and `_messages` are byte-for-byte unchanged from #1308 (diff confirms zero lines touched in either function). Ran `test_issue_1216_curator_signal.py` (the #1308-era normal-fit fixture) and `test_knowledge_curator.py -k "oversized_lesson_batch or partial_batch_watermark"` (the #1307/#1308 boundary tests) unmodified against this branch — all 8 pass verbatim, confirming a batch that fits behaves identically to how #1308 left it.

## Verification performed

- `pytest tests/test_knowledge_curator_quarantine.py tests/test_knowledge_curator.py -v` — 24/24 passed.
- Non-vacuity: swapped in `origin/main`'s (pre-#1309, post-#1307/#1308) `knowledge_curator.py` and reran the six new quarantine tests — all six fail (one `KeyError: 'blocked_reason'`, one `IndexError` on a second run that should have progressed, four more on missing quarantine behavior), confirming they exercise genuinely new code. Restored the fixed file; `git diff --quiet` confirmed a clean restore.
- Independently drove a 5-item mid-batch scenario (`M00..M04`, `M01` oversized) across two `run_curation` calls, outside the shipped test file, confirming the watermark advances past the quarantined item alone and the items after it are curated in the same run they're first seen as head, not skipped.
- Independently drove the keyless-oversized-head edge case: confirmed no watermark write, no quarantine row, and the same `blocked_reason` string repeats stably on a second run with no data loss — the stall is visible and consistent, not silent.
- Reproduced the host measurement live, read-only, on the current release (not the release the PR measured against) — 997 items, max 2,113 chars, matching the PR's figures closely enough to confirm it is a real, repeatable measurement rather than a one-off or stale number.
- Grepped the full repo for any reader of `curator/quarantine.jsonl` or `curator/status.json`'s `quarantined` field — none exists (same as the precedent `decisions.jsonl`/`errors.jsonl`); the PR states this itself and does not overclaim observability.
- `grep`-confirmed `MAX_INPUT_CHARS` unchanged at 48,000; confirmed `fit_lessons_to_input_budget`/`_messages` diff-empty against #1308.
- Full suite: `pytest tests/ -q` — 3093 passed, 17 skipped, 4 failed, exactly the established Windows baseline (`test_bridge_cycle_tags`, three `TestAcquireBridgeLock` cases), matching the PR's own reported numbers.
- `ruff check` on both touched files — clean.

## Residual, named rather than silently accepted

`curator/quarantine.jsonl` and the `status.json.quarantined` field are currently written but read by nothing — same gap as `decisions.jsonl`/`errors.jsonl` already have, not a new one this PR introduces, and the PR discloses it rather than implying observability that doesn't exist. Worth a cheap follow-up (surface the count somewhere an operator actually looks), not a blocker here.

ozand:
Merging on this review. Three things in it are worth naming, because they are the standard rather than the exception.

**The measurement was reproduced, not accepted.** The reviewer re-ran the PR's methodology on the *current* release rather than the one the PR cited, and got 997 items against the PR's 993 — explaining the drift by the 35 minutes and the deploy in between, and confirming the same largest item (`LESS-20260904-90931e43`, 2,113 chars) in both runs. The largest lesson ever produced is **4.4% of the 48,000-char budget**. That turns the arm choice from a judgement into an application of the issue's own rule: nothing has ever come near half the budget, so the cheapest arm wins, and arm 1 is what the data points at.

**Non-vacuity was demonstrated, not assumed.** Swapping `origin/main`'s pre-#1309 file back in makes all six new quarantine tests fail, then a clean restore confirmed by `git diff --quiet`. A test suite that passes against the unfixed code proves nothing, and this one does not.

**The 'who reads it' question got the honest answer.** Nobody reads `curator/quarantine.jsonl` today — grep across `nanobot/`, `scripts/` and `host/eeepc/` finds no reader, and the dashboard does not touch `curator/` at all. The PR states that itself rather than implying observability it does not have, which is the same disclosure standard #1307 and #1308 met. It inherits the durability profile of `decisions.jsonl` and `errors.jsonl` — append-only, fsynced, never rotated — so it is a countable file, not a read one, and it says so.

The reviewer's suggested follow-up — surfacing the `quarantined_ids` count on the dashboard, since the `status.json` fields already exist — is cheap and correct in principle. I am deliberately not filing it: at a 4.4% high-water mark this path may never fire in production, and an unread record of an event that does not occur is not worth a surface. If a quarantine row ever appears, that is the moment to add the reader, and the row itself will be the trigger.

What this closes: the curator can no longer be stalled indefinitely by one oversized lesson, the watermark advances past exactly the quarantined item and nothing more — driven mid-batch, outside the shipped tests — and the keyless-oversized-head edge stalls visibly and repeatably rather than silently.

Issue #1313, opened 2026-09-05
AGENTS.md is 3,680 chars from stopping every cycle: reflection lane adds ~1,400 bytes/day and #1302 turns overflow into exit 4


ozand:
#1300 was closed as fixed and it is fixed: the executor prompt now drops only sections the operator declared droppable, largest first, and records what went. But the fix converted a silent failure into a loud one, and the loud one is a cycle that spawns nothing. Measuring the remaining headroom today puts it at **3,680 characters**, against a reflection lane that has been adding **~1,400 bytes a day for seven days**.

That is roughly two and a half days of fuse. This issue is to make the budget self-limiting before it burns.

## Measured now (host `eeepc`, 2026-09-05 13:42 MSK)

Instance `AGENTS.md` at `784753fe`:

```
442 lines / 24583 bytes
27 '## ' sections — 14 declared droppable, 13 critical (7602 bytes)
```

Growth, from `git show <rev>:AGENTS.md | wc -c` at each day's last commit:

```
2026-08-29  14718
2026-08-30  18794
2026-08-31  22314
2026-09-01  23449
2026-09-02  23449
2026-09-03  23449
2026-09-04  23449
2026-09-05  24583
```

+9,865 bytes in seven days. The four flat days are the deploy freeze, not a change in lane behaviour.

## The headroom, exactly

Every `phase: system_prompt` ledger row today — six of them, 07:10:08Z through 10:35:10Z — is byte-identical in its totals:

```json
{"phase": "system_prompt", "cycle_id": "cycle-4a577d4482f1", "chars": 23268, "cap": 24000,
 "dropped": [ ...10 sections, 13500 chars total, all "declared-droppable"... ]}
```

So:

- **732 chars** of slack under the cap after the drops that are already happening.
- **2,948 chars** of reserve — four droppable sections the fit loop has not needed to remove yet, because it breaks as soon as the prompt fits:

```
706  ## Prompt budget (droppable sections)
751  ## Turn budget checkpoints (80-iteration cycles)
622  ## Early skip evaluation
869  ## Secret verification scoping (git diff, not full-file grep)
```

Total headroom **3,680 chars**. At 1,409 bytes/day that is **2.6 days**, and it is shorter than that if any single future append is critical and larger than 732 chars — six of the eleven appends measured in #1188 were.

## What happens when it runs out

`nanobot/agent/context.py:226`, strict mode:

```python
sections, dropped = self._drop_droppable_bootstrap_sections(sections, cap)
[...]

Issue #1305, opened 2026-09-05
Audit non-executor role prompt caps and content loss (proposer, curator, strategist, reflector)


ozand:
## Problem

The executor cap is governed by #1300, but it does not establish whether the proposer, knowledge curator, strategist, or reflector loses prompt content through `ContextBuilder._fit_system_prompt` or role-local input/context caps. This leaves a cross-role observability and correctness gap.

## Context

Perform one read-only scope audit of the four named non-executor roles. Inspect code call sites and, where safely available, live-host prompt/state evidence without changing host state. For each role, answer yes/no on whether prompt content can currently be omitted by `ContextBuilder/_fit_system_prompt` or a role-owned cap; identify the cap, omitted section/input, truncation order or selection rule, and whether durable evidence records the omission. #1300 remains executor-only and is related evidence, not a substitute for this audit.

## Scope

- Trace the prompt-construction call paths for proposer, knowledge curator, strategist, and reflector.
- Determine per role whether `ContextBuilder`/`_fit_system_prompt` is used and whether a role-local cap can discard input.
- Collect sanitized, read-only code-call-site and live-host-state evidence sufficient for a per-role yes/no result.
- Compare findings with #1300 and existing reflector issue #1291; report overlap without changing either issue's scope.
- If an affected role is confirmed, identify it for a subsequent, role-specific issue rather than broad remediation here.

## Out of scope

- Code, configuration, test, deployment, host, runtime-state, branch, worktree, or documentation changes.
- Remediation design or implementation.
- Reopening, expanding, or duplicating #1300.
- Creating remediation issues during this audit unless separately authorized.

## Acceptance criteria

- [ ] A concise per-role matrix covers proposer, knowledge curator, strategist, and reflector, each with a yes/no content-loss result.
- [ ] Each result cites the relevant code call site and cap/selection behavior; each uses sanitized live-host evidence where available, or explicitly states why it is unavailable.
- [ ] The report distinguishes `ContextBuilder/_fit_system_prompt` loss from role-local caps and distinguishes prompt-input loss from output limits.
- [ ] The report explicitly compares executor-only #1300 and reflector observability #1291, avoiding duplicate scope.
- [ ] Any confirmed affected role is named as a candidate for one follow-up role-specific issue; no remediation is included here.

## Verification

Read-only source inspection and read-only live-host prompt/state inspection only. Record commands and sanitized findings in the investigation report; no tests, deployments, or mutations are required.

## Dependencies

[...]

PR #1315, opened 2026-09-05
fix(reflector): bound prompt inputs by whole messages, rows and entries (#1314)

### LINKED ISSUE #1314: Reflector slices three serialized JSON inputs at fixed offsets — the same tear that produced #1307 in the curator, currently under budget only by input size
The #1305 scope audit found the reflector clean today and said so precisely: *"the raw-slice mechanism could tear JSON if future inputs cross, but it is not dropping content today."* This issue is the second half of that sentence.

The reflector builds its candidate prompt by slicing three serialized JSON documents at fixed character offsets:

`nanobot/runtime/reflector.py:283-285`

```python
f"CYCLE_ID: {cycle_id}\nTRANSCRIPT:\n{json.dumps(transcript, ensure_ascii=False)[:_MAX_TRANSCRIPT_CHARS]}\n"
f"LEDGER:\n{json.dumps(ledger, ensure_ascii=False)[:_MAX_LEDGER_CHARS]}\n"
f"PRIOR REFLECTIONS:\n{json.dumps(journal, ensure_ascii=False)[:_MAX_JOURNAL_CHARS]}"
```

with

```python
_MAX_TRANSCRIPT_CHARS = 48_000
_MAX_LEDGER_CHARS     = 12_000
_MAX_JOURNAL_CHARS    = 12_000
```

This is character-for-character the mechanism that produced #1307 in the curator: serialize a structure to JSON, then cut the string at a byte offset. The cut lands wherever it lands — mid-key, mid-string-literal, mid-escape-sequence — and the model receives a document that is not JSON, with no marker saying where it stopped being JSON.

## Why it is filed while it is not firing

#1305 measured 96 recent candidate prompt records and **0 of 96** crossed any of the three caps; the largest observed was 32,631 against the 48,000 transcript cap. There is no current loss and this issue does not claim one.

It is filed because the three inputs are exactly the ones that grow:

- `transcript` grows with executor turn count, and the turn budget is 80.
- `ledger` grows with per-cycle phase rows — #1302 and #1303 each added new row types this week, and #1313 proposes another.
- `journal` is the reflector's own prior output, appended per reflection, archived only at `_MAX_JOURNAL_BYTES = 512 * 1024`.

All three move in one direction. The 48,000/12,000/12,000 caps are the only thing between them and a torn document, and crossing a cap produces no error, no log line, and no ledger row — the prompt simply becomes malformed and the reflection built on it is silently lower quality. The failure is indistinguishable from a bad model day.

That is the same reason #1307 went unnoticed in the curator until it was looked for directly.

## The two states are not the same

A cap that has never been crossed and a cap that cannot tear look identical in every artifact the runtime produces. The distinction only exists in the code. This is worth fixing before the inputs cross rather than after, because after, the evidence of the tear is a prompt that was already thrown away.

## Acceptance criteria

- [ ] None of the three inputs is truncated by slicing serialized JSON. Bound the structure before serializing — drop whole transcript turns, whole ledger rows, whole journal entries — so the document handed to the model is always parseable.
- [ ] Truncation is visible when it happens: what was dropped and how much, recorded where the reflector's other telemetry goes, so a future audit can answer "did this prompt lose content" without re-deriving it from input sizes.
- [ ] A test drives each of the three inputs past its cap and asserts the emitted section parses as JSON. It must fail against current `main` — verify that and say so in the PR, because a test that passes before the fix is testing nothing.
- [ ] The bound is chosen per input rather than reused: 48,000 for a transcript and 12,000 for a ledger were picked for the old mechanism, and whatever replaces them should say why each number is what it is.
- [ ] `_MAX_JOURNAL_BYTES = 512 * 1024` archives the live journal but the prompt slice is 12,000 chars — confirm which of the two actually bounds what reaches the model, and state it, since the audit did not separate them.

## Not in scope

The reflector's telemetry blindness — no `component: reflector` rows in `llm_calls` since 2026-07-07 — is a separate defect and is not fixed here. It is worth knowing while working on this file, because it means the reflector's own LLM calls will not appear in the usual place when verifying this change.

ozand:
Closes #1314.

## The measurement first: one of the three caps is being crossed today

Read-only on the host as `eeepc-agent` (2026-09-05, release `632bffbb` state), serializing exactly as `_messages` did (`json.dumps(value, ensure_ascii=False)`):

| input | what reaches the model | observed | old cap | crossed? |
|---|---|---|---|---|
| transcript | latest `llm_calls/prompts` record of the cycle | max 32,652 (443 records, p50 4,370; 54 messages max; largest single message 17,043) | 48,000 | never — and it cannot: the recorder caps a record at 32 KiB (`MAX_LLM_PROMPT_PAYLOAD_BYTES`, #1039) |
| ledger | the cycle's rows, proposed row first | max 3,059 chars / 12 rows over the last 60 cycles; largest row 1,278 (`system_prompt`) | 12,000 | never |
| journal | last 10 reflection rows across archives + live | **today's tail 16,319**; rows p50 266 / p95 2,896 / max 3,690; **349 of 1,031 ten-row windows exceed 12,000**, max window 25,989 | 12,000 | **yes, every run** |

#1305 measured the journal tail at 11,293 on 2026-09-04 and called it under budget. One day of appended rows moved the same window over the cap. Run over today's real inputs (fetched read-only, evaluated locally) the old slice fails on the PRIOR REFLECTIONS section for every cycle:

```
old journal:    len=16319 cap=12000 -> TORN (Unterminated string starting at at char 11481)
```

**This is a live-defect fix, not a hardening.** The `PRIOR REFLECTIONS` section has been arriving as broken JSON on roughly a third of reflector runs (349 of 1,031 ten-row windows; every run today), stopping mid-string with no marker, while the model is asked to fill `followed_previous` from it. Everything built on those reflections — the demand items, the curator's reflector-derived lessons, the `AGENTS.md` additions #1313 is about — degraded silently. #1305's "0 of 96" answered a narrower question: it sampled candidate prompt records, and the journal tail is assembled separately in `_journal_tail`, so it never entered that sample. The issue text has been corrected accordingly.

**And the transcript cap was dead code, not headroom.** `record_llm_prompt` caps every record at 32 KiB upstream (`_cap_payload_record`, #1039) before it is written, so no record the reflector can read was ever within 15,000 chars of the 48,000 slice. The three largest records on the host (32,652 / 32,631 / 32,461) are the recorder's cap showing through, not the transcripts' natural size.

## What changes (`nanobot/runtime/reflector.py`)

Every input is bounded **before** serialization, so the document is always JSON; a section that lost content says so on its label line, and every journal row records what happened.

- `_fit_rows(rows, max_chars, protect_head)` drops whole rows oldest-first after a protected head; the head goes last, so a section can be emptied but never torn. Ledger: protected head is the `proposed` row, so the outcome row survives longest. Journal: plain oldest-first, the newest reflections (the recommendations the model is asked to check) survive.
- `_fit_transcript(record, max_chars)` drops whole executor turns oldest-first, protecting leading `system` messages and the first `user` message (the task). `reasoning_content` / `content` are replaced whole by `_OMITTED` only when the record cannot fit even with zero turns — dropping turns could never rescue such a record, so the field goes and the turns stay.
- `_build_prompt(...) -> (messages, input_fit)` assembles the prompt; labels become e.g. `LEDGER (14 of 23 rows omitted, oldest first, 12684 chars):` when something was dropped, and stay bare (`LEDGER:`) otherwise, so the model knows where the record is incomplete. `_messages` is kept as a thin wrapper for compatibility.
- `run_reflector` journals `input_fit` on **every** reflection row, success and error: `{"status": "complete"|"truncated", "transcript": {budget, total, kept, dropped, chars[, dropped_chars][, fields_omitted][, recorder_truncated]}, "ledger": {...}, "journal": {...}}`, and counts `input_truncated` in the run result. Recorded when complete too, so a row without the key is a row from before this change, not a row that happened not to drop.
- **Upstream truncation is carried through (review finding).** A record the recorder already cut to 32 KiB arrives with `truncated: true`, field values shortened in place — in the worst case the whole message list replaced by `…[payload truncated to fit 32KB budget]`. Such a record fits the budget and drops nothing here, but the prompt *has* lost content. `_fit_transcript` now copies that flag into `transcript_fit.recorder_truncated`, it flips `input_fit.status` to `truncated`, and the label line says `record already shortened by the prompt recorder's 32 KiB cap`. Without this, the row this PR exists for would answer "no" to "did this prompt lose content" exactly on the long cycles where the answer is "yes".

### Budgets, chosen per input against the table above

- **transcript = `MAX_LLM_PROMPT_PAYLOAD_BYTES` (32,768)**. The recorder already caps a record at 32 KiB, so the reflector can never be handed more; 48,000 was dead code. Pinning to the recorder's constant means the message drop acts only if the two caps ever diverge, and `input_fit.transcript.dropped` will say so. The largest record today (32,652 chars) fits whole with 116 to spare; chars ≤ bytes, so a record under the recorder cap always fits.
- **ledger = 12,000**, kept but now derived: ~4× the largest cycle observed (3,059) and nine rows of the largest shape (1,278). Row types were added by #1302 and #1303 this week and #1313 proposes another; the headroom is for that, and the fit makes the cap safe rather than fatal.
- **journal = 30,000** = `_JOURNAL_TAIL` (10) × 3,000, the observed p95 row (2,896) rounded up. Covers every ten-row window observed on the host (max 25,989); ten rows at the maximum observed size (3,690) would drop two, visibly. This is the one budget that changes behaviour, so the rejected arm is recorded here: **keep 12,000 and drop whole entries**, which the new mechanism makes free and labelled. Rejected because it would hand the model 5–7 of the 10 reflections on every run — today's tail would lose 3 — while `_JOURNAL_TAIL = 10` is the declared window `followed_previous` is checked against; dropping a third of it permanently would be #1183's promotion blind-spot in a smaller form (a declared window that the code quietly reads narrower). Cost of raising: +18,000 chars (~4.5k tokens) per reflector call at 9–13 cycles/hour on `cl/gemini-3.5-flash-low`, whose context is not the constraint. If the cost is the wrong side of the trade, the right lever is `_JOURNAL_TAIL`, not a silent char cap: shrink the window explicitly and the budget follows as 10 × 3,000 → N × 3,000.

### `_MAX_JOURNAL_BYTES` vs the prompt slice — settled

`_MAX_JOURNAL_BYTES = 512 KiB` bounds only the live **file**. `_journal_tail` reads `iter_reflection_rows`, which spans the newest two archives plus the live file, so rotation does not change the ten rows the model sees at all. What reaches the model is bounded by `_JOURNAL_TAIL = 10` rows, then by `_MAX_JOURNAL_CHARS`. Today the char cap was the active bound and it tore; after this PR the row count is the active bound and the char cap is a whole-entry guard. Stated in the constants comment.

## Tests (`tests/test_reflector_input_fit.py`, 10) — all fail on `main`, verified

Run against the unmodified module before the fix (log kept): the four structural tests die in `json.loads` on the emitted section —

```
json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 47215 (char 47214)   # transcript
json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 11878 (char 11877)   # ledger
json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 8306 (char 8305)     # journal
```

the rest with `AttributeError: _build_prompt`, `assert 48000 == 32768`, `KeyError: 'input_truncated'`. After the fix: 9 passed.

- transcript over budget → section parses, system + task message kept, newest turn kept, label says `N of M messages omitted`
- oversized `content` field → replaced whole, small messages kept, label says `content omitted`
- recorder-truncated record (`truncated: true`, marker message) → passes through byte-identical, `status: truncated`, `recorder_truncated: true`, `dropped: 0`, label names the recorder; an untruncated twin stays `complete` without the key
- ledger over budget → parses, `proposed` first and `outcome` last, drops oldest-first
- journal over budget → parses, newest entries kept
- under budget → all three sections byte-identical to the inputs, bare labels, `input_fit.status == "complete"`
- budgets pinned: transcript == recorder cap, journal ≥ 10 × 2,900, ledger 12,000
- `run_reflector`: `input_fit` on the success row and on the error row; `input_truncated` counted; complete fit recorded so absence is distinguishable

`tests/test_reflector.py` unchanged, 29 green. Ruff clean.

Full Windows suite at `2fcd07ca` (on `52267426`):

```
4 failed, 3106 passed, 17 skipped, 11 warnings in 1553.39s (0:25:53)
```

Exactly the four Windows-baseline failures, matched by name, not by count.

## Real-input check

Today's newest reflected cycle and the largest retained record, fetched read-only and run through `_build_prompt` locally: all three sections parse, `input_fit.status == "complete"`, nothing dropped (journal 16,319 of 30,000; transcript 32,652 of 32,768; ledger 3,567 of 12,000). The same inputs through the old slice tear the journal at char 11,481.

## Not done, on purpose

- **The duplicated `proposed` row.** `run_reflector` filters the cycle's rows (which already include `proposed`) and then inserts the `proposed` row again at index 0, so the ledger section carries it twice (`total: 23` in the test for 22 rows). Pre-existing, ≤692 chars, not this issue's mechanism; noted, not changed.
- **Projecting journal rows for the model.** `parse_reason`, `response_head`, and now `input_fit` ride along in PRIOR REFLECTIONS. Stripping telemetry keys from the model-facing tail would change what the model sees beyond the ask; left as is, ~220 chars per row.
- **Reflector telemetry blindness** (no `component: reflector` rows since 2026-07-07): separate defect, untouched; verification here used the journal rows, not `llm_calls`.
- `nanobot/agent/context.py` and the AGENTS.md write path: untouched (pi-1 on #1313).

First reflector run: every new row in `state/reflector/reflections.jsonl` carries `input_fit`; expected `status: complete` with `journal.chars` in the 15–20K range. A `truncated` row names the section and count; the label line in the prompt dump says the same.

### DIFF
--- a/nanobot/runtime/reflector.py
+++ b/nanobot/runtime/reflector.py
@@ -355,6 +355,14 @@ def _fit_transcript(record: Any, max_chars: int) -> tuple[Any, dict[str, Any]]:
         fit["dropped_chars"] = dropped_chars
     if fields_omitted:
         fit["fields_omitted"] = fields_omitted
+    if out.get("truncated") is True:
+        # The prompt recorder already cut this record to fit its 32 KiB cap
+        # (#1039): field values shortened in place, in the worst case the whole
+        # message list replaced by one marker. Such a record fits here and drops
+        # nothing, yet the prompt HAS lost content -- the record says so and the
+        # fit must repeat it, or "complete" would be the wrong answer exactly on
+        # the long cycles this issue is about.
+        fit["recorder_truncated"] = True
     return out, fit
 
 
@@ -365,6 +373,8 @@ def _fit_note(fit: dict[str, Any], unit: str) -> str:
         parts.append(f"{fit['dropped']} of {fit['total']} {unit} omitted, oldest first, {fit.get('dropped_chars', 0)} chars")
     if fit.get("fields_omitted"):
         parts.append(", ".join(fit["fields_omitted"]) + " omitted")
+    if fit.get("recorder_truncated"):
+        parts.append("record already shortened by the prompt recorder's 32 KiB cap")
     return f" ({'; '.join(parts)})" if parts else ""
 
 
@@ -397,7 +407,10 @@ def _build_prompt(cycle_id: str, transcript: dict[str, Any], ledger: list[dict[s
         body = _json(value)
         fit["chars"] = len(body)
         sections.append(f"{label}{_fit_note(fit, unit)}:\n{body}")
-    truncated = any(fit["dropped"] or fit.get("fields_omitted") for fit in (transcript_fit, ledger_fit, journal_fit))
+    truncated = any(
+        fit["dropped"] or fit.get("fields_omitted") or fit.get("recorder_truncated")
+        for fit in (transcript_fit, ledger_fit, journal_fit)
+    )
     input_fit = {
         "status": "truncated" if truncated else "complete",
         "transcript": transcript_fit,
--- a/nanobot/runtime/reflector.py
+++ b/nanobot/runtime/reflector.py
@@ -14,7 +14,12 @@ from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import Any, Callable
 
-from nanobot.observability.llm_telemetry import call_context, record_llm_call, record_llm_prompt
+from nanobot.observability.llm_telemetry import (
+    MAX_LLM_PROMPT_PAYLOAD_BYTES,
+    call_context,
+    record_llm_call,
+    record_llm_prompt,
+)
 from nanobot.runtime.model_registry import resolve_model
 
 _MAX_CYCLES = 3
@@ -33,9 +38,31 @@ _MAX_JOURNAL_BYTES = 512 * 1024
 _ARCHIVE_RETENTION_DAYS = 90
 _ARCHIVE_READ_FILES = 2
 _ARCHIVE_RE = re.compile(r"^reflections-(\d{4}-\d{2}-\d{2})(?:-(\d+))?\.jsonl\.gz$")
-_MAX_TRANSCRIPT_CHARS = 48_000
+# #1314: the three prompt inputs are bounded structurally in _build_prompt --
+# whole executor messages, whole ledger rows, whole journal entries -- so the
+# document the model receives always parses. Until #1314 the serialized JSON
+# was sliced at these offsets, the curator tear of #1307 in another file.
+# Each budget is chosen against the host measurement of 2026-09-05:
+#  - transcript: the prompt recorder already caps one record at 32 KiB
+#    (MAX_LLM_PROMPT_PAYLOAD_BYTES, #1039), so the reflector can never be
+#    handed more than that; the old 48,000 was unreachable (largest record
+#    seen: 32,652). Pinning the budget to the recorder cap means the message
+#    drop only acts if the two caps ever diverge.
+#  - ledger: per-cycle context is at most 3,059 chars / 12 rows, largest row
+#    1,278 (a system_prompt row); 12,000 is ~4x the largest cycle and holds
+#    nine rows of the largest shape -- headroom for the row types #1302,
+#    #1303 and #1313 keep adding.
+#  - journal: _JOURNAL_TAIL rows at the observed p95 row size (2,896; max
+#    3,690). The old 12,000 was crossed by 349 of 1,031 ten-row windows on
+#    the host (today's tail: 16,319 chars), so the PRIOR REFLECTIONS section
+#    was already being torn every run.
+# _MAX_JOURNAL_BYTES above bounds only the live file: readers span the newest
+# archives plus the live file, so rotation never changes the tail the model
+# sees. What reaches the model is _JOURNAL_TAIL rows, then _MAX_JOURNAL_CHARS.
+_MAX_TRANSCRIPT_CHARS = MAX_LLM_PROMPT_PAYLOAD_BYTES
 _MAX_LEDGER_CHARS = 12_000
-_MAX_JOURNAL_CHARS = 12_000
+_MAX_JOURNAL_CHARS = 30_000
+_OMITTED = "…[omitted: reflector input fit]"
 _VALID_FINDINGS = {"wasted_steps", "error_pattern", "tool_misuse", "good_practice"}
 _VALID_RECOMMENDATIONS = {"skill_candidate", "instruction_change", "approach_hint"}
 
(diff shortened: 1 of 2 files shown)


ozand:
**Verdict: approve with one required change and one question.** The mechanism is right, the budgets are argued from real measurements rather than asserted, and the fit record answers the audit question the issue asked for. One gap makes that record lie in a specific case.

## Required: the transcript fit reports `complete` for a record the recorder already truncated

`_MAX_TRANSCRIPT_CHARS = MAX_LLM_PROMPT_PAYLOAD_BYTES` is a sound coupling and the safe direction is real — `len()` counts characters, the recorder's cap counts UTF-8 bytes, and characters never exceed bytes, so the reflector's own drop loop can never fire. Your comment says exactly that.

The consequence is the problem. `_cap_payload_record` (`llm_telemetry.py:209`) does not tear JSON — it trims field values in place, re-serializes, and stamps `truncated: True` — but it does lose content, up to and including replacing the whole message list with `…[payload truncated to fit 32KB budget]`. When that has happened, `_fit_transcript` sees a record that already fits, drops nothing, and emits:

```python
{"budget": 32768, "total": N, "kept": N, "dropped": 0}
```

`input_fit["status"]` then reads `complete`. The PR's own stated purpose is that "an audit can answer 'did this prompt lose content' from the row alone", and in this case the row answers no when the answer is yes — and it is the case most likely to matter, since a 32 KB-overflowing transcript is a long cycle.

Propagate it: if `record.get("truncated")` is true, carry that into `transcript_fit` and let it count toward `status`. Cheap, and it closes the hole that the rest of the PR exists to close.

## Question: the journal budget went up 2.5x and the cost is not stated

`_MAX_JOURNAL_CHARS` 12,000 → 30,000. The derivation is fine — `_JOURNAL_TAIL = 10` at p95 row 2,896 is 28,960 — and I accept that keeping ten entries whole is the right target. What is missing is the other side.

The new mechanism gave you a second option the old one did not: keep 12,000 and drop whole entries, which costs nothing extra and now says so on the label line. You chose to raise instead, which adds roughly 18,000 characters — call it 4.5k tokens — to a prompt built once per cycle, at 9-13 cycles an hour.

That may well be correct; ten prior reflections is what `followed_previous` is checked against, and truncating to four would degrade it. But say so in the PR body, with the rejected option and its reason, the way you did for the other two budgets. Right now the transcript and ledger numbers are argued and the one that actually changed behaviour is not.

## Confirmed by reading

- `_fit_rows` protected-head handling is correct: `pop(protect_head if len(kept) > protect_head else 0)` removes the oldest unprotected row while any exist, then falls through to the protected head, so a section empties rather than tears. Termination is guaranteed — `kept` shrinks every iteration.
- `_fit_transcript` tests the field omission against `{**out, "messages": []}`, so `reasoning_content` / `content` are only replaced when no amount of turn-dropping could rescue the record. That matches the docstring and is the right order.
- The `proposed` row is protected at the ledger head, which keeps the row the reflection is about when older rows go.
- Both error and success paths journal `input_fit`, so a torn-input reflection is diagnosable even when parsing failed.

## Correction to the issue I filed

#1314 said the reflector was "currently under budget only by luck of input size" and cited #1305's 0-of-96. Your measurement — 349 of 1,031 ten-row windows over the 12,000 journal cap, today's tail at 16,319 chars — says the PRIOR REFLECTIONS section has been torn on most runs, not none. The issue understated this as pre-emptive when it was live. I will correct it on #1314; the PR body should describe this as a fix to an active defect, not a hardening.

That also means #1305's audit was answering a narrower question than it appeared to. It measured candidate prompt records and reported no cap crossings; the journal tail is assembled separately and was outside what it sampled.

## Merge condition

CI is still pending on 3.11/3.12/3.13 and your full local suite was at ~4% when the PR opened. Post the tail before merge, and confirm the four known Windows failures by name rather than by count — this has been silently five before.

ozand:
**Review addressed at `2fcd07ca`: the mandatory fix is in with a driving test, the journal arm is recorded, the body now states the live defect and the dead 48,000, and the full-suite tail with the four baseline names is in the body.**

**Mandatory — recorder truncation carried through.** `_fit_transcript` copies the record's `truncated: true` into `transcript_fit.recorder_truncated`; `_build_prompt` counts it toward `input_fit.status = "truncated"`; the label line reads `TRANSCRIPT (record already shortened by the prompt recorder's 32 KiB cap):`. Test `test_recorder_truncated_record_is_reported_as_lost_content` feeds the recorder's own worst-case shape (`messages` = the single `…[payload truncated to fit 32KB budget]` marker, `truncated: True`): the section passes through byte-identical, `dropped: 0`, status `truncated`, and an untruncated twin stays `complete` without the key. Fails on `main` (no `_build_prompt`).

**Journal budget — rejected arm now in the body.** Keeping 12,000 and dropping whole entries would hand the model 5–7 of the declared 10 on every run (3 lost on today's tail); a declared window the code reads narrower is #1183 in miniature. Cost of raising stated (+18,000 chars, ~4.5k tokens per call, 9–13 cycles/h, model context not the constraint), and the honest lever if the cost is wrong is `_JOURNAL_TAIL`, not a char cap — shrink the window explicitly and the budget follows as N × 3,000.

**Framing.** Body now opens the fix as a live defect: `PRIOR REFLECTIONS` broken on ~⅓ of runs (349 / 1,031 windows; every run today), downstream demand items / curator lessons / #1313's `AGENTS.md` additions degraded silently, and why #1305's 0/96 missed it (candidate prompt records sampled; `_journal_tail` assembled separately). The 48,000 is called dead code outright: #1039 caps records at 32 KiB upstream, and the three largest host records (32,652 / 32,631 / 32,461) are that cap showing through.

**Full suite at `2fcd07ca`:** `4 failed, 3106 passed, 17 skipped` — `test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle`, `test_bridge_locking.py::TestAcquireBridgeLock::test_acquires_when_free`, `::test_second_acquire_in_same_process_is_contended`, `::test_contended_lock_via_monkeypatched_flock`. Names, not count. Ruff clean. Not merging my own PR.

Issue #1312, opened 2026-09-05
Dashboard globs two artifact families that have no writer, and calls both 'stale' instead of 'retired' (materialized-cycle-* frozen 53d, evolution-* frozen 14d)


ozand:
The semantic status colours from #1277 (PRs #1287/#1301) went live in release `20260905T070336Z-canonical-0f2f87b5`. The first thing the honest rendering shows is that two of the dashboard's artifact families have **no writer at all**, and the dashboard calls both of them `stale` — the one word the ADR-002 contract reserves for a live writer that is late.

## What the live page renders right now

Fetched from `http://127.0.0.1:8080/` on `eeepc` at 2026-09-05 13:42 MSK:

```
Approval Gate:          stale; age=1283.2h (context-only artifact)
Latest Improvement:     stale; age=1283.2h (context-only artifact)
Concrete Statement:     stale; age=1283.2h (context-only artifact)
Latest Report Status:   stale; age=348.1h (context-only artifact)
Artifact Freshness:     materialized=1283.2h ago, report=348.1h ago
```

1283.2h is 53.4 days. 348.1h is 14.5 days.

## Defect 1: `materialized-cycle-*.json` has zero writers

`scripts/eeebot_dashboard.py:1377` globs `IMPROVEMENT_DIR` for `materialized-cycle-*.json`. On the host:

```
newest materialized-cycle-*.json : Jul 14 02:54   (334 files)
newest file in improvements/     : Sep  5 13:31   (llm-proposed-cycle-4a577d4482f1.json)
prefix census, improvements/     : 1752 llm-proposed-cycle
                                    334 materialized-cycle
```

The directory is busy. The globbed prefix is not. `nanobot/runtime/state_paths.py:87` declares exactly one writer for the segment:

```python
"improvements": ("nanobot.runtime.llm_proposer:write_request",),
```

and that writer emits `llm-proposed-{cycle_id}.json` (`llm_proposer.py:2592`). A repo-wide search for a producer of `materialized-cycle-` returns only readers: the dashboard, and `scripts/verify_materialized_improvement.py:17`, which is pinned to a single hardcoded July filename.

This feeds four rendered fields — Approval Gate, Latest Improvement, Concrete Statement, and the materialized half of Artifact Freshness.

## Defect 2: the source selector pins the gate to the dead family forever

`select_artifact_source` (`scripts/eeebot_dashboard.py:1110-1126`) chooses on path presence only, with no age comparison:

```python
if materialized_path is not None:
[...]

PR #1316, opened 2026-09-05
fix(dashboard): retire writerless artifact readers

ozand:
## Scope

Refs #1312. Both writerless artifact families are retired, not replaced with fictional writers. llm-proposed requests are not materialized approval evidence. Dashboard no longer scans frozen evolution/materialized files or returns cached rewards. Dependent fields explicitly say retired. Rollout-preflight does not treat historical reports as partial rollout proof.

## Review corrections (pushed)

- Removed both orphan:#1312 markers AND the ORPHAN_ISSUES entry. reports is outside the registry; improvements names only its live llm_proposer writer. Corrected last-reader history remains in the comment.
- Retired verify_materialized_improvement.py entrypoint: no hardcoded July path or artifact read; emits `retired (#1312): materialized improvement writer removed; verification unavailable`. Exit 0 avoids creating spurious defect demand; this is not a PASS verdict. Historical validation helpers remain for compatibility, without filesystem reads.
- Added explicit retired CSS rule and explanation: offline palette intentionally shared, dashed outline distinguishes retirement.
- Removed dead IMPROVEMENT_DIR / REPORTS_DIR constants and their unused environment configuration.
- Removed the always-True production selector invocation. Both retired families have no selectable evidence. The isolated selector helper and malformed-live provenance test remain; no invented future caller is claimed.

## Test-first evidence

Before the initial production change, tests failed on unchanged main code:
- `AssertionError: retired artifact reader accessed filesystem` (evolution glob).
- `TypeError: select_artifact_source() got an unexpected keyword argument 'materialized_retired'`.

Additional review regressions were run before their production fixes and failed: dead constants still existed; the materialized verifier attempted an artifact read. Logs: red-1312.log and red-review-1312.log in the isolated worktree.

## Verification after review corrections

Focused dashboard/preflight/registry suite: **58 passed in 1.41s**.
Full pytest redirected to full-review-1312.log with explicit worktree cd, no output pipeline:

```
4 failed, 3099 passed, 17 skipped, 11 warnings in 1543.16s (0:25:43)
```

Exactly the named Windows baseline, no fifth failure. Full-file Ruff was not globally green due to pre-existing findings; no claim of global lint cleanliness.

## Review / rollout boundary

Owner review corrections are addressed above. Independent Opus review timed out; no independent verdict is claimed and it does not block preservation/publication. Latest pushed-head CI passed on Python 3.11 (7m10s), 3.12 (3m15s), and 3.13 (2m59s), Actions run 33968607253.

No merge/deploy/restart/host-state mutation. After merge and deployment, read-only curl of the live dashboard must quote Approval Gate, Latest Improvement, Concrete Statement and Artifact Freshness verbatim in the issue closing comment. Before-deploy HTML is baseline only. No /api/cleanup, fabricated state, or secrets read. Issue remains open pending this acceptance.

### DIFF
--- a/nanobot/runtime/state_paths.py
+++ b/nanobot/runtime/state_paths.py
@@ -48,8 +48,8 @@ STATE_PATH_WRITERS: dict[str, tuple[str, ...]] = {
     # control_plane/ and self_evolution/ lost their autoevolve reader (#1224).
     # evolution reports froze with the coordinator, but dashboard/preflight
     # readers survived: #1312 retires them explicitly, not as healthy emptiness.
-    # This orphan names evolution-*.json only; external proof reports stay live.
-    "reports": ("orphan:#1312",),
+    # Their final artifact readers were removed by #1312. No orphan is carried;
+    # reports stays outside this registry (external proof reports remain live).
     "curator": (
         "nanobot.runtime.knowledge_curator:_write_decision",
         "nanobot.runtime.knowledge_curator:_stage_promotions",
@@ -85,7 +85,7 @@ STATE_PATH_WRITERS: dict[str, tuple[str, ...]] = {
         "nanobot.runtime.hypothesis_backlog:append_hypotheses",
     ),
     # llm-proposed requests remain live; materialized-cycle evidence is retired.
-    "improvements": ("nanobot.runtime.llm_proposer:write_request", "orphan:#1312"),
+    "improvements": ("nanobot.runtime.llm_proposer:write_request",),
     "ledger": ("nanobot.runtime.cycle_ledger:append_event",),
     "llm_calls": (
         "nanobot.observability.llm_telemetry:record_llm_call",
@@ -121,5 +121,5 @@ STATE_PATH_WRITERS: dict[str, tuple[str, ...]] = {
 # Every ``orphan:#N`` reference must name an issue listed here, with the
 # one-line reason the orphan is being carried rather than fixed in place.
 ORPHAN_ISSUES: dict[str, str] = {
-    "#1312": "Materialized-cycle and evolution writers are retired; dashboard and preflight retain explicit retired labels, not artifact reads.",
+    # Empty: #1312 removes the artifact readers, rather than carrying orphans.
 }
--- a/nanobot/runtime/state_paths.py
+++ b/nanobot/runtime/state_paths.py
@@ -45,11 +45,11 @@ STATE_PATH_WRITERS: dict[str, tuple[str, ...]] = {
     ),
     # apply.ok is written by the operator (runbook), read by the bridge gate.
     "approvals": ("repo:docs/EEEPC_APPLY_OK_OPERATOR_RUNBOOK.md",),
-    # control_plane/, reports/ and self_evolution/ left this registry with their
-    # last reader, autoevolve (#1224): current_summary.json and evolution-*.json
-    # froze 2026-08-21/22 with the deleted coordinator, self_evolution/ was
-    # never written on the host. proof-*.json in reports/ is live but written
-    # and read outside nanobot/.
+    # control_plane/ and self_evolution/ lost their autoevolve reader (#1224).
+    # evolution reports froze with the coordinator, but dashboard/preflight
+    # readers survived: #1312 retires them explicitly, not as healthy emptiness.
+    # This orphan names evolution-*.json only; external proof reports stay live.
+    "reports": ("orphan:#1312",),
     "curator": (
         "nanobot.runtime.knowledge_curator:_write_decision",
         "nanobot.runtime.knowledge_curator:_stage_promotions",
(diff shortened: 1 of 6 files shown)


ozand:
**Verdict: request changes.** The retirement itself is right and thorough — both scanners gone, the fields say `retired`, the age-blind selector no longer pins the gate to a dead family, and the full suite is at baseline by name. Two things contradict the PR's own claim and two are loose ends.

## Required 1: the orphan markers assert the opposite of what this PR does

`orphan:#N` means a path is **read** with no live writer, carried deliberately rather than fixed. This PR removes every reader. After it, nothing under `nanobot/` or `scripts/` reads `evolution-*.json` or `materialized-cycle-*.json` — I checked the branch: the only remaining hits are prose in `health.py:201`, `state.py:915`/`:1084`, and the new comment in `state_paths.py:51` itself.

So `"reports": ("orphan:#1312",)` and `"improvements": (..., "orphan:#1312")` declare carried orphaned reads that this PR just deleted. Retiring a reader and declaring it an orphan are mutually exclusive states.

Two more reasons this addition is not load-bearing:

- `scan_read_segments` walks `PACKAGE = ROOT / "nanobot"` only. `scripts/` is never scanned, so the dashboard's reads were never in the scan — before or after. The registry entry fixes nothing `test_every_read_segment_declares_a_writer` enforces; it is a voluntary declaration, and it is a false one.
- The registry's own docstring already settles the mixed case: *"a directory with one live file and one orphaned file (`control_plane/`) is declared by its live writer and the orphaned read is recorded in the issue; see #1222."* `improvements` has a live writer. Under the file's stated convention it is declared by that writer alone.

The honest end state is the one the old comment described: `reports` stays out of the registry, `improvements` keeps `llm_proposer:write_request` by itself, and `ORPHAN_ISSUES` goes back to empty. Note `test_registry_entries_are_non_empty_and_orphan_issues_are_explained` asserts `set(ORPHAN_ISSUES) <= used`, so the entry and the markers have to go together.

## Required 2: `verify_materialized_improvement.py` still reads a retired artifact

`scripts/eeepc_privileged_rollout_preflight.py` was handled. `scripts/verify_materialized_improvement.py:17` was not:

```python
"/var/lib/eeepc-agent/self-evolving-agent/state/improvements/materialized-cycle-43b67e3806de.json"
```

#1312 named this as the second reader of the family. It is a hardcoded path to a file frozen on 2026-07-14. Retire it here, or state why it stays — a script pinned to one dead artifact is exactly the shape that made the family look alive in the first place.

## Should fix: `retired` renders identically to `unavailable`, by omission

There is no `.status-badge[data-status="retired"]` rule, so it falls through to the base `.status-badge`, which sets `var(--state-offline)` — pixel-identical to `unavailable` and `offline`.

The line directly above it shows the convention this file already uses:

```css
/* context-only intentionally shares the offline palette: both convey non-authoritative state, distinguished by attributes */
.status-badge[data-status="context-only"] { background: var(--state-offline); ... }
```

An explicit rule and a comment saying the sharing is deliberate. `retired` gets the same palette with neither. Since the thesis of #1312 is that `retired` and `unavailable` send the operator to different places — one says look at the writer, the other says delete the reader — the one word that carries that distinction should not arrive at its styling by fallthrough. Either give it a rule of its own or give it the `context-only` treatment with the comment.

## Should fix: two dead constants left pointing at the retired families

`IMPROVEMENT_DIR` (`eeebot_dashboard.py:129`) and `REPORTS_DIR` (`:135`) are now defined and never read — I grepped the branch, the only hits are the definitions and the env-var name. A configured, documented constant pointing at a retired directory is how the next reader gets written. Delete them with their env vars, or if `EEEBOT_REPORTS_DIR` / `EEEBOT_IMPROVEMENTS_DIR` are part of an operator contract, say so in a comment.

## Question: two parameters that are always `True`

`select_artifact_source` gained `materialized_retired` and `report_retired`, and the one production call site passes `True` to both, unconditionally. The non-retired branch is now unreachable outside tests, and the function's return at that site is always `({}, None, "none")`.

I asked for the malformed-but-live case to be preserved and tested, and I can see why you kept the shape. But `AGENTS.md` says no overengineering, and a parameter whose only possible production value is a constant is configuration that cannot be configured. Either drop the call at that site and let the retirement be visible in one place, or say in the docstring what future caller is expected to pass `False` — if the answer is "a family that gets a writer again", that is worth writing down, and then the parameters earn their keep.

## Confirmed good

- Full suite 4 failed / 3098 passed, and the four are the Windows baseline by name.
- `report_source_status = materialized_source_status = "retired"` and `artifact_freshness` reading `retired (#1312)` satisfy the acceptance criterion that these fields stop claiming staleness.
- `scan_report_artifacts` and `scan_all_report_rewards` return empty rather than replaying cached rewards from frozen files — the right call, and the docstrings say why.
- Not deploying and not claiming the live page as proof was correct. The four verbatim fields are still owed against the deployed release.

ozand:
All five owner review points addressed and pushed: orphan markers and entry removed together; hardcoded verifier retired without artifact reads; retired CSS explicit with dashed outline; dead directory constants removed; always-True production selector call removed. Full local pytest: 3099 passed, 17 skipped, exactly four named Windows baseline failures. Latest-head CI now PASS on Python 3.11/3.12/3.13 (run 33968607253). PR body updated via REST API. Independent Opus timed out; no verdict claimed and no further blocking retry. Deployment and verbatim live fields remain a separate post-merge step.

Issue #996, opened 2026-08-25
Goal-gap futility: suppress a metric direction after N integrated proposals without metric movement


ozand:
## Context (self-contained)

The demand pipeline turns scorecard metric gaps into `goal-gap` demand items (`demand._goal_gap_items` ← `scorecard.goal_gaps`). A completed goal-gap id is only suppressed for a short TTL (#778) because "the condition recurs" — but nothing ever checks whether the work done for a gap actually moved its metric.

Live evidence (ledger, 2026-08-19..25): 151 of 216 written proposals — 70% — served just TWO goal-gap ids:
- `goal-gap-5d4d5a9dc822` = `repeat_failure_rate (V1)` — 84 proposals
- `goal-gap-acb65b1911ab` = `confirmed_integration_ratio (V2)` — 67 proposals

Neither metric improved. The proposals are homogeneous "Add X and Y to scripts/<self-made script>.py" tweaks; for `repeat_failure_rate` the loop is circular — the metric is inflated by the proposer's own self-dedup churn (see #977 analysis), so writing more failure-analysis scripts cannot move it, the gap is re-presented after its TTL, and the monoculture continues. This is the DGM-class "objective chasing" failure: an evolutionary loop pouring effort into a direction whose fitness signal does not respond.

## Task

Deterministic futility detection — no LLM anywhere in this path:

- Track, per goal-gap demand id, the count of INTEGRATED proposals (outcome success + integrated) since the gap was first presented, and snapshots of the gap's metric value at presentation time (scorecard already computes the metrics; persist a small history sidecar under `state/`).
- When a gap has accumulated ≥ N integrated proposals (default 10, env-tunable) while its metric has not improved beyond a small epsilon, mark the gap FUTILE: suppress it from demand collection for a long TTL (default 14 days, env-tunable), emit one `goal_gap_futile` ledger event naming the gap, metric, attempt count, and metric delta, and record the decision in a sidecar (`state/demand/futility.json` or similar) so a future strategist role can read it.
- Futility suppression must be visible: the scorecard/ops surface should expose currently-futile gaps (a line in the existing scorecard output is enough).
- New machinery ships as its own small module (≤300 LOC, stdlib only), registered in the deny-set like other control-plane integrity modules. `demand.py` only gains the read hook.
- Fail-open: any error in futility tracking degrades to "gap presented as today" — never blocks demand collection.

## Acceptance Criteria

[...]

PR #1317, opened 2026-09-05
fix(validator-harness): widen findings parser to read failed_count (#1208 step 2)

ozand:
## #1208 step 2 — the schema decision

Step 1 (PR #1258, merged `8fac1549`, deployed) instrumented every validator-harness run with `findings_parse`, `stdout_keys`, and `stdout_truncated` — turning "the parser probably discards some output" into a measured fact. This PR is step 2: decide the canonical schema and implement it.

### 1. Predicted demand direction — before any code change

The operator's own paragraph is correct, confirmed independently by reading `nanobot/runtime/demand.py:1115-1170` directly: `isinstance(exit_code, int) and exit_code != 0` is checked **unconditionally before** the findings branch — not an if/elif keyed on findings first. A schema fix can therefore only move a candidate from "no verdict" (`findings_count=None`, exit 0) to "findings=0, no demand" (flat) or "findings>0, new demand" (up). **It cannot suppress an already-raised exit-code-driven item**, because exit-code demand is checked first and unconditionally, and it cannot suppress a currently-invisible item either — there is nothing to suppress from `None`.

Concretely: of the population this PR's widening newly classifies as `matched` (see below), one live case at write time flips from invisible to a genuine new candidate — `validate_unused_imports.py` exits 0 under the harness's bare (no-argument) invocation but its own JSON reports `failed_count: 2`, because its `--strict` gate (the thing that would make it exit 1) needs a flag the harness never passes. That is exactly the operator's predicted shape: repairing the parse can raise demand, never lower it. No candidate in the newly-matched population goes from raising to suppressing.

### 2. The canonical schema — decided and recorded

**Direction: the harness widens to read what the scripts already emit, not the reverse.** Live measurement (this session, re-run against the deployed sidecar) found the `unrecognised_keys` population is not uniform:

- **15 of 23** unrecognised paths share one exact shape: `{"success": bool, "checked_count": int, "failed_count": int, "results": [<one row per item CHECKED, pass or fail>]}`. `failed_count` is already exactly a findings count under a different name — verified by reading the deployed sources of `validate_deprecation_status.py` and `validate_unused_imports.py` directly (`failed_count` increments once per item that failed its check, never once per item checked).
- **5 of 23** use five *different* field names for the same concept: `check_git_clean_branch`'s `modified` (list), `check_last_cycle_repeat`'s `is_repeat` (bool), `check_lesson_markdown_sections`'s `violations` (list), `check_stale_feeds`'s `stale` (list), `validate_markdown_format`'s `issues` (list) + `warning_count`. No single shared field name unifies these five, so a harness-side widening cannot reach them without becoming a five-way special case — left unrecognised, deliberately.
- **1 of 23** (`check_disk.py`) was already `matched`.
- **2 of 23** are `not_json` due to output truncation (see #4).

Widening `_classify_findings` to additionally recognise an int `failed_count` field fixes the 15-validator majority with **one read site**, versus rewriting 15 separate script files (several of which have their own tests and callers already depending on the current field names) to a name none of them chose. Rejected direction and reason: making the scripts emit the harness's four names (`findings`/`alerts`/`missing`/`failures`) instead — rejected because it multiplies the edit surface by 15+ files for a single-consumer benefit, and doesn't even unify the other 5 divergent shapes.

**Deliberately not widened: `results`/`issues`/`all_issues` array length as a count.** That array holds one row per item *checked*, not per item *failed* — confirmed live: a 970-item clean run (`validate_filename_compatibility.py`) reports `len(results)==970`, `failed_count==0`. Reading the array length as a findings count would report every clean run's total scanned-file count as N findings — the exact silent mismatch #1208 step 1 was built to catch, reproduced under a different key. This is the trap named in the issue's own "trap" section, caught before it shipped.

A `failed_count` that is a `bool` (a subclass of `int` in Python) is explicitly excluded (`isinstance(value, int) and not isinstance(value, bool)`) so a script that repurposed the name for a flag can't silently become a count of 0 or 1.

### 3. The 23-with-no-`findings_parse`-row population

Re-measured live at write time (not from the issue's stale count): the population with no `findings_parse` on their latest row is **18**, not 23 — the number keeps shrinking as the 4-runs/day rotation catches up since `8fac1549` deployed (2026-09-03T19:07Z). Checked all 18 against `rotation.json`'s `served` map: **every one is present with a `last served` timestamp on or before the deploy** — none is bypassed, skipped, or unreachable. At 5 validators served per run / 4 runs per day against 53 total paths, full coverage takes ~2.6 days; this is a timing fact confirmed by evidence, not a defect, and no code change applies. (The earlier session's 30-vs-330-row coverage confusion, self-corrected on the issue thread, was exactly this distinction — path-has-a-row vs. path-has-a-*classifiable*-row.)

### 4. `stdout_truncated: true` — what the cap protects, and what's behind it

One live case as of this measurement: `validate_filename_compatibility.py`, 168,732 bytes of stdout against the 64 KiB (`_MAX_OUTPUT_BYTES`) cap — reproduced directly: truncating its real output at 65,536 bytes yields `JSONDecodeError: Unterminated string`, confirming the cap genuinely cuts valid JSON mid-document rather than coincidentally aligning with a parse boundary.

The cap protects against a runaway/malicious/OOM-inducing subprocess printer (module docstring #928: "a runaway printer was bounded only by the unit's `MemoryMax`... i.e. an OOM kill, not by this cap"). The thing behind it here is **not** unbounded or malicious — it emits exactly one row per repository file under the same `results`-array convention as the 15 `failed_count` scripts (970 files today), and both historically-cited large emitters (`check_style.py`, 184,676 bytes; `verify_imports.py`, 2,175,307 bytes) reproduce the identical pattern live. This is a **document-size** problem (one bounded-but-growing dimension — repo file count — outgrowing a fixed byte cap), not a **key-naming** problem, and fixing it (a streaming/summarized JSON shape from the scripts, or a per-script output budget scaled to file count) is a different, separable change with its own risk profile. Scoped out of this PR deliberately rather than bundled; named as a follow-up. Practically low-severity today: `failed_count: 0` on the one live truncated case, so the exit-code fallback already agrees with the (unreachable) true count.

## What changed

- `nanobot/runtime/validator_harness.py`: `_classify_findings` now also matches an int `failed_count` field (`_FINDINGS_COUNT_KEYS`), in addition to the four existing list/dict `_FINDINGS_KEYS` fields. Module docstring and INVOCATION CONTRACT section updated to document the widened contract and the rejected `results`-length alternative.
- `tests/test_validator_harness.py`: new `test_failed_count_int_is_now_a_recognised_findings_field` (written first, confirmed to fail against pre-fix code — see below); the one step-1 test whose fixture happened to use `failed_count` was renamed/adjusted to a shape with none of the recognised fields, since its old assertion (`findings_count is None` for that exact shape) is now the opposite of correct behavior.

## Non-vacuity

Ran the new test against pre-fix `_classify_findings` directly (`(None, 'unrecognised_keys', [...])` for the live 15-validator shape) — confirmed it currently discards `failed_count` before writing the test. Full new test then run and confirmed to fail for the right reason (`assert clean_row["findings_count"] == 0` → `AssertionError: None == 0`) before implementing the widening.

## grep every read site of `findings_count` (per the issue's trap)

Only one: `nanobot/runtime/demand.py:1116` (`_validator_defect_items`), which turns a positive count into `defect` demand and a zero into nothing. No dashboard, no other runtime module, no script reads `findings_count`, `findings_parse`, or `stdout_keys` — confirmed by grep across `nanobot/`, `scripts/`, and the instance repo before writing this PR. The widened field feeds the same single consumer step 1 already measured; this PR does not create a new, unread machinery.

## Verification

- `pytest tests/test_validator_harness.py tests/test_demand.py -v` — 291 passed, 5 skipped (pre-existing POSIX-only skips).
- Full suite: `pytest tests/ -q` — 3097 passed, 17 skipped, 4 failed — exactly the established Windows baseline (`test_bridge_cycle_tags`, three `TestAcquireBridgeLock` cases). No fifth failure.
- `ruff check nanobot/runtime/validator_harness.py tests/test_validator_harness.py` — clean.
- Live re-measurement against the deployed `validator_harness/last_runs.jsonl` and the instance repo's actual script sources (read-only, no host mutation) underlies every population count and shape claim above — none of it is inherited unverified from the issue thread.

## Roll-out

Read-only measurement only in this session; the fix itself is a pure parser widening with no schema-breaking change (existing `matched`/`not_json`/`not_object` states unchanged; only a subset of `unrecognised_keys` rows move to `matched`). After deploy, the next several `eeebot-validator-harness.timer` fires (00:00/06:00/12:00/18:00 MSK) should show the 15-validator population's rows carry `findings_parse: matched` with real `findings_count` values, and `demand._validator_defect_items` should surface the `validate_unused_imports.py` candidate predicted above (unless it has since been fixed on its own). Re-checking this against the prediction in §1 is the intended follow-up, not this PR's job to force.

Closes step 2 of #1208; the issue's overall verdict (used/unused, displaced capacity, suppression-worthiness) was already answered in the issue thread before this PR — this PR is the schema-decision implementation the thread called for.

### DIFF
--- a/nanobot/runtime/validator_harness.py
+++ b/nanobot/runtime/validator_harness.py
@@ -136,15 +136,23 @@ defect when a script did not anticipate it:
 - Non-zero exit = a finding worth surfacing: it becomes ``defect`` demand
   with your stderr/stdout as evidence (see
   ``demand._validator_defect_items``).
-- ``--json`` output is read for a findings count ONLY from a top-level
-  object carrying one of :data:`_FINDINGS_KEYS` as a list/dict. Any other
-  shape yields ``findings_count: None`` (exit code decides alone) and the
-  sidecar row says why — ``findings_parse`` is ``matched`` /
-  ``not_json`` / ``not_object`` / ``unrecognised_keys`` (with the
-  ``stdout_keys`` seen), ``json_flag`` says whether the flag was passed,
-  ``stdout_truncated`` whether the :data:`_MAX_OUTPUT_BYTES` cap cut the
-  document (#1208 step 1: measured live, six validators emitted JSON daily
-  under keys this parser never read, invisibly).
+- ``--json`` output is read for a findings count from a top-level object
+  carrying one of :data:`_FINDINGS_KEYS` as a list/dict, OR an int
+  :data:`_FINDINGS_COUNT_KEYS` field (``failed_count`` — #1208 step 2, the
+  schema decision: the harness widens to the scripts' own established
+  convention rather than rewriting 15+ scripts to a name none of them
+  chose). Any other shape yields ``findings_count: None`` (exit code
+  decides alone) and the sidecar row says why — ``findings_parse`` is
+  ``matched`` / ``not_json`` / ``not_object`` / ``unrecognised_keys`` (with
+  the ``stdout_keys`` seen), ``json_flag`` says whether the flag was
+  passed, ``stdout_truncated`` whether the :data:`_MAX_OUTPUT_BYTES` cap
+  cut the document (#1208 step 1: measured live, validators emitted JSON
+  daily under keys this parser never read, invisibly; #1208 step 2: 15 of
+  those share ``failed_count``, an int already meaning exactly the same
+  thing this parser was already counting for the four list/dict fields —
+  widened; a script's ``results``/``issues``/``all_issues`` array of
+  per-item audit rows, whose length equals items CHECKED rather than items
+  FAILED, is deliberately left unrecognised rather than risked as a count).
 - Decay self-declaration is detected from printed output (#936): when a
    script's captured stdout+stderr contains :data:`_DECAY_DECL_RE` AND the
    script's own repo-relative path (``scripts/<filename>``), the run is
@@ -289,6 +297,35 @@ _SANDBOX_DENIAL_RE = re.compile(
 # JSON object with one of these list/dict fields counts; anything else
 # yields None (no findings parsed), never a crash or a guess.
 _FINDINGS_KEYS = ("findings", "alerts", "missing", "failures")
+
+# #1208 step 2 (the schema decision): a second, narrower field the harness
+# reads directly as a COUNT rather than a container length. Live measurement
+# (issue #1208, 2026-09-05) found 15 of 23 previously-``unrecognised_keys``
+# validators share one shape: ``{"success", "checked_count",
+# "failed_count": <int>, "results": [<one row per item CHECKED, pass or
+# fail>]}``. ``failed_count`` is already exactly a findings count under a
+# different name in every script that emits it -- confirmed by reading the
+# deployed sources of ``validate_deprecation_status.py`` and
+# ``validate_unused_imports.py`` directly: it is incremented once per item
+# that FAILED, never once per item checked. ``results`` is deliberately NOT
+# added here: its length equals ``checked_count`` on every script observed,
+# clean or not (confirmed live: a 970-item clean run reports
+# ``len(results)==970``, ``failed_count==0``), so reading it as a findings
+# count would report a clean run's total scanned-file count as N findings --
+# the exact silent-mismatch shape #1208 step 1 was built to catch, not one to
+# reproduce under a new key. The direction of this fix (the harness widens
+# to the scripts' existing convention, not the other way) was chosen because
+# it touches ONE read site here rather than 15 separate script files, each of
+# which would need to rename an established field their own tests and any
+# other caller already depend on -- and because five of the 23 unclassified
+# validators (``check_git_clean_branch``'s ``modified``,
+# ``check_last_cycle_repeat``'s ``is_repeat``,
+# ``check_lesson_markdown_sections``'s ``violations``,
+# ``check_stale_feeds``'s ``stale``, ``validate_markdown_format``'s
+# ``issues``/``warning_count``) use FIVE DIFFERENT field names for the same
+# concept, so no single script-side rename could unify them either -- the
+# harness is the one place all 53 scripts already funnel through.
+_FINDINGS_COUNT_KEYS = ("failed_count",)
 
 
 def _parse_ts(value: Any) -> datetime | None:
@@ -546,23 +583,38 @@ _MAX_RECORDED_KEYS = 24  # top-level keys kept in ``stdout_keys`` for an unrecog
 
 def _classify_findings(stdout: str) -> tuple[int | None, str, list[str]]:
     """Deliberately tiny findings heuristic, now with its verdict spelled
-    out (#1208 step 1). Returns ``(findings_count, findings_parse, stdout_keys)``:
+    out (#1208 step 1) and widened by one exact field (#1208 step 2). Returns
+    ``(findings_count, findings_parse, stdout_keys)``:
 
-    - ``"matched"``: a top-level JSON object with a
+    - ``"matched"``: a top-level JSON object with EITHER a
       ``findings``/``alerts``/``missing``/``failures`` list or dict —
-      ``findings_count`` is that field's length.
+      ``findings_count`` is that field's length — OR an int
+      ``failed_count`` (:data:`_FINDINGS_COUNT_KEYS`), read directly as the
+      count, checked only when none of the container fields matched.
+      ``failed_count`` is checked as ``isinstance(value, int) and not
+      isinstance(value, bool)`` — bool is an int subclass in Python, and a
+      script that repurposed the name for a flag must not silently become a
+      count of 0 or 1.
     - ``"not_json"``: stdout did not parse (plain text, empty, or a JSON
       document truncated at :data:`_MAX_OUTPUT_BYTES`).
     - ``"not_object"``: valid JSON whose top level is not an object.
     - ``"unrecognised_keys"``: a valid object with none of
-      :data:`_FINDINGS_KEYS` — ``stdout_keys`` carries the top-level keys
-      actually seen (sorted, first :data:`_MAX_RECORDED_KEYS`).
+      :data:`_FINDINGS_KEYS` or :data:`_FINDINGS_COUNT_KEYS` — ``stdout_keys``
+      carries the top-level keys actually seen (sorted, first
+      :data:`_MAX_RECORDED_KEYS`).
 
     Every state but ``"matched"`` yields ``findings_count`` ``None``; the
     caller then falls back to the raw exit code alone, never a fabricated
-    count. Before #1208 the three ``None`` states were one value, which is
-    how six live validators emitted JSON daily to keys nobody read without
-    anything recording it."""
+    count. Before #1208 step 1 the three ``None`` states were one value,
+    which is how six live validators emitted JSON daily to keys nobody read
+    without anything recording it. Before #1208 step 2, ``failed_count``
+    (an int already meaning exactly "number of findings" in every script
+    that emits it — confirmed against 15 live validators sharing this
+    shape) fell into ``unrecognised_keys`` alongside the five validators
+    that genuinely use unrelated shapes; deliberately NOT reading
+    ``results`` for the same reason — its length is the count of items
+    CHECKED, not the count that failed, and is identical on a clean run and
+    a failing one."""
     try:
         data = json.loads(stdout)
     except Exception:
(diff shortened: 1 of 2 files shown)


ozand:
**Verdict: approve.** One thing to file before merge, and one question that should be answered in the PR body rather than discovered later. This is the most carefully evidenced PR of the batch — the prediction came first, the trap was caught before it shipped, and every population count was re-measured instead of inherited from the issue thread.

## Required before merge: file the truncation follow-up

Scoping the 64 KiB cap out of this PR is right — it is a document-size problem, not a key-naming one, and bundling it would have mixed two risk profiles. But "named as a follow-up" with no issue number is how a finding evaporates.

The evidence in this PR is already enough to file it, and it is worse than the one live case suggests:

```
_MAX_OUTPUT_BYTES                        65,536
validate_filename_compatibility.py      168,732   (2.6x)
check_style.py                          184,676   (2.8x)
verify_imports.py                     2,175,307   (33x)
```

`verify_imports.py` is not occasionally truncated — at 33x the cap it can never produce parseable JSON under any repo state that has existed. That validator's findings have been structurally invisible since the cap was introduced, and no growth in the repo will change it in either direction. Please open the issue with these three numbers and link it here, so the `stdout_truncated: true` rows have somewhere to land.

## Question: the widening turns one script's declared advisory into demand

`validate_unused_imports.py` on the deployed host:

```python
if args.strict:
    return 0 if result["success"] else 1
return 0
```

with the module docstring saying "the exit code is 0. Pass `--strict` to exit with code 1 when unused imports are…", and a comment at line 36 explaining the choice as "rather than blocking the validator suite".

So exit 0 without `--strict` is not an accident of invocation — it is the script's author stating that these findings are advisory. The harness never passes the flag. After this PR the harness reads `failed_count: 2` from a script that is simultaneously reporting success, and `demand._validator_defect_items` raises a defect item from it.

That may well be the correct call: the harness asks "how many findings", the script answers 2, and the exit code is about invocation mode rather than about whether findings exist. But it is a decision, and right now it arrives as a side effect. You predicted the case precisely in §1 and called it "exactly the operator's predicted shape" — which it is, directionally — without saying whether the resulting demand is *wanted*.

State which reading the PR adopts. If it is "findings exist regardless of the script's exit policy", say so and note that the harness is deliberately overriding a per-script advisory setting. If the intended answer is instead that the harness should pass `--strict`, that is a different and smaller change. Either is defensible; silently taking the first is the thing to avoid, because this is the #996 shape — a correct number wired to a gate that was not designed to receive it.

## What this PR got right, specifically

- **The prediction came first and was correct.** Confirming from `demand.py:1115-1170` that the exit-code branch is unconditional and precedes findings, then deriving that a schema fix can only move a candidate from `None` to flat-or-up, is the analysis I got wrong when I first predicted this work would *reduce* demand. Good that it was re-derived from the source rather than taken from the thread.
- **The trap was caught before it shipped, not after.** `results`/`issues`/`all_issues` hold one row per item *checked*: `validate_filename_compatibility.py` reports `len(results) == 970` with `failed_count == 0`. Reading array length as a count would have announced 970 findings on a clean run — the exact silent mismatch step 1 was built to detect, reproduced under a new key. Rejecting it explicitly, with the live numbers, is worth more than the fix itself.
- **`bool` excluded from `int`.** `isinstance(value, int) and not isinstance(value, bool)` — a script repurposing `failed_count` as a flag cannot become a count of 0 or 1. That is the kind of thing that is invisible until it produces a phantom demand item a year later.
- **My stale count corrected.** 23 → 18, with all 18 present in `rotation.json`'s `served` map at timestamps on or before the `8fac1549` deploy. At 5 served per run × 4 runs/day against 53 paths, ~2.6 days to full coverage. A timing fact, not a defect, and the right conclusion is that no code change applies — which takes more discipline to write than a fix would have.
- **Every read site grepped.** One: `demand.py:1116`. The widened field feeds the consumer step 1 already measured, so this does not create machinery nobody reads. That was the explicit trap in the brief and it was answered with a number, not an assurance.
- **Non-vacuity shown, not claimed.** The pre-fix classifier's actual return (`(None, 'unrecognised_keys', [...])`) and the actual assertion failure (`AssertionError: None == 0`) are both quoted.
- Full suite 3097 passed / 4 failed, and the four named. CI green on 3.11 / 3.12 / 3.13.

ozand:
Спасибо за обзор. По порядку.

## 1. Issue на усечение — заведён

**#1319**, с тремя числами именно в теле: `_MAX_OUTPUT_BYTES=65,536`, `validate_filename_compatibility.py=168,732 (2.6x)`, `check_style.py=184,676 (2.8x)`, `verify_imports.py=2,175,307 (33x)`. Отдельно назван структурный (не «иногда усекается») характер `verify_imports.py`: при 33-кратном превышении разбираемый JSON недостижим ни при каком состоянии репозитория, кроме сильно урезанного. Ссылка добавлена в PR #1317.

## 2. `validate_unused_imports.py` — предсказание оказалось не тем случаем, который я думал

Проверил на хосте перед ответом, а не с ходу принял первое чтение. Копнул глубже: оба из двух `failed_count: 2` элементов — не новые находки, а **дрейф построчного baseline**:

```
BASELINE_UNUSED_IMPORTS (в самом скрипте):
  ("check_mutation_surface.py", 21, "typing.Set")
  ("run_all_tests.py", 920, "pytest")

Реально репортится сегодня:
  check_mutation_surface.py, line 22   <- import Set сдвинулся на 1 строку
  run_all_tests.py, line 962           <- import pytest сдвинулся с 920 на 962
```

`import pytest` в `run_all_tests.py` существует ровно один раз (grep подтверждает), сейчас на строке 962 — не на 920. Baseline матчится по точному номеру строки; файл правился после того, как baseline был зафиксирован (несколько auto-commit'ов между ними: `f64d9d1e --max-seconds`, `4b82ff93 dual verification mode`, добавившие строки выше импорта), сдвинув обе записи и молча выведя оба baseline-исключения из действия.

Это меняет мой ответ на заданный вопрос. Это не «харнесс поднимает советующий сигнал скрипта как обязательный» — это **`validate_unused_imports.py` уже теряет собственный advisory-контракт из-за построчного baseline**, харнесс лишь первый, кто это заметил, потому что раньше `failed_count` никто не читал. Данные до этого PR были невидимы; после — видимы, и то, что стало видно, оказалось настоящим (хоть и мелким, инфраструктурным, не поведенческим) дефектом самого скрипта, а не решением харнесса о режиме.

**Отвечаю на прямой вопрос**: принимаю чтение «находки есть независимо от exit-политики скрипта» — харнесс намеренно не читает флаг `--strict`/`--silent-pass`/аналоги ни у одного из 15 скриптов этой популяции (сверился: `validate_deprecation_status.py`, `validate_documentation_code_blocks.py`, `validate_mutable_defaults.py`, `validate_no_debug_statements.py`, `validate_no_eval_exec.py`, `validate_no_shell_true.py`, `validate_script_entrypoints.py`, `validate_timezone_aware_datetime.py`, `validate_type_comparison.py`, `validate_json_syntax.py` — все имеют одинаковый паттерн `return 0 if result["success"] else 1` БЕЗ advisory-обёртки; `validate_unused_imports.py` — единственный из 15 с явным `if args.strict: ... else: return 0`). Харнесс уже перекрывает per-script advisory для остальных 14 — `validate_unused_imports.py` просто единственный, где это явно замечено, потому что это единственный из 15 с advisory-режимом вообще. Симметрия важнее исключения: раз `demand` уже поднимает defect из `failed_count>0` для 14 скриптов без advisory-обёртки, оставлять именно этот один невидимым out of consistency было бы произвольно.

Передавать `--strict` харнессом не буду предлагать как правку в этом PR — это была бы вторая, отдельная и меньшая правка (один флаг для одного скрипта из 53), и она либо ничего не изменит (baseline и так исключает известный долг), либо изменит поведение ровно этого одного скрипта нерегулярно относительно остальных 14. Комментарий с этим разбором и решением добавлен в тело PR #1317.

Ни один из живых `failed_count: 2` findings сейчас не поднимает demand ошибочно в смысле «выдумано»: оба реальны (не baseline), просто дрейфовавшие относительно собственного baseline скрипта — что само по себе достойный отдельный, самостоятельный найденный дефект в `validate_unused_imports.py`, а не в этом PR или харнессе. Не заводил на него отдельный issue без вашего указания — скажите, если стоит.

ozand:
Correction to my comment above: I cited the truncation follow-up as **#1319** before creating it — that number belongs to a pre-existing, unrelated issue. The actual issue is **#1320**: https://github.com/ozand/eeebot/issues/1320

PR #1318, opened 2026-09-05
fix(runtime): expose and reclaim prompt reserve (#1313)

### LINKED ISSUE #1313: AGENTS.md is 3,680 chars from stopping every cycle: reflection lane adds ~1,400 bytes/day and #1302 turns overflow into exit 4

ozand:
Closes #1313

## What changed

- Records `droppable_reserve_chars` on every successful and overflow `phase: system_prompt` ledger row. The value is the exact character total of declared-droppable bootstrap sections still standing after prompt fitting.
- Covers the real strict boundary: all droppable sections are exhausted, critical content still exceeds the cap, `SystemPromptOverflowError` carries reserve `0`, the bridge spawns nothing and returns exit `4`.
- Preserves #1193 rather than adding a second guard: root `AGENTS.md` remains operator-owned and autonomous proposals/diffs continue to reject as `operator_owned_path`.
- Adds an explicit operator-only removal path, `scripts/agents_md_consolidate.py`. It is dry-run by default, requires named `## ` sections and `--apply`, removes only uniquely matched sections carrying the exact droppable marker, refuses the whole operation for missing/duplicate/unmarked headings, preserves symlinks, and replaces the real file atomically. It is not imported or invoked by the loop.
- Corrects the stale subagent-bridge spec that still said the script tier allowed `AGENTS.md`, and records the decision in ADR-003.

## Why this arm

The autonomous loop must not gain deletion authority over the file containing its charter, staging rules, and forbidden-path guardrails; the prose-blind gate cannot distinguish consolidation from deleting a safety invariant. #1193 already stopped new autonomous growth. Removal is therefore explicit and operator-owned, and is limited to sections the operator has already declared droppable.

Rejected alternatives: automatic age/size deletion, deleting every section omitted from a prompt, and raising `NANOBOT_SYSTEM_PROMPT_MAX_CHARS`. The first two can remove critical meaning; the cap increase only moves the failure date.

## Test-first evidence

The reserve/bridge test was added before production changes and executed against the unmodified implementation. Collection failed as expected:

```text
TypeError: SystemPromptOverflowError.__init__() got an unexpected keyword argument 'reserve_chars'
```

The public field was then deliberately named `droppable_reserve_chars` to keep the ledger contract unambiguous.

## Validation

- Focused acceptance set: `234 passed`.
- Full Windows suite: `4 failed, 3108 passed, 17 skipped`; failures are exactly the established baseline:
- Focused Ruff on new/changed standalone surfaces: passed. Existing whole-file Ruff debt in `bridge.py`/`context.py` is unchanged.
- `git diff --check`: passed.
- Independent Opus review found and blocked a symlink-replacement flaw; the CLI now resolves the real target and has a regression test. Re-review verdict: OK.

## Rollout verification owed

After deployment, quote the latest live `phase: system_prompt` row as two numbers: `cap - chars` slack and `droppable_reserve_chars`. Then run the consolidation CLI first in dry-run mode against the operator-selected instance sections; apply only through the separately reviewed instance-repository change.

No cap was raised. `nanobot/runtime/reflector.py` was not touched. No host state or live symlink was mutated.

### DIFF
--- a/docs/adr/ADR-003-operator-owned-agents-consolidation.md
+++ b/docs/adr/ADR-003-operator-owned-agents-consolidation.md
@@ -1,0 +1,68 @@
+---
+title: Keep AGENTS.md operator-owned and consolidate only declared-droppable sections
+status: proposed
+date: 2026-09-05
+authors: [eeebot maintainers]
+related: ["#1188", "#1193", "#1300", "#1313"]
+tags: [runtime, prompt, operations]
+---
+
+# Status
+
+Proposed — implemented in the change governed by #1313, pending merge and rollout.
+
+# Context
+
+#1188 measured 20 autonomous integrations in five days whose only substantive change was appending to the instance `AGENTS.md`, against six deleted lines. #1193 then made the file operator-owned. #1300/#1302 replaced positional truncation with declared-droppable sections and a loud overflow. On 2026-09-05, #1313 measured only 732 prompt characters of slack plus 2,948 characters of still-droppable reserve.
+
+The remaining decision is who may remove accumulated guidance. Giving the autonomous loop deletion authority over the file would also let it delete charter, staging, and forbidden-path instructions that the prose-blind gate cannot evaluate.
+
+# Decision
+
+`AGENTS.md` remains operator-owned. Autonomous proposals and diffs targeting it remain rejected as `operator_owned_path`.
+
+Removal is an explicit operator action through `scripts/agents_md_consolidate.py`. The operator names each `## ` section. The tool removes a section only when there is exactly one matching heading and that section carries the exact `<!-- prompt-fit: droppable -->` declaration. It is dry-run by default, requires `--apply`, and replaces the file atomically. Missing, duplicate, or unmarked headings reject the whole operation without a partial write.
+
+The loop never imports or invokes this tool, and `agents_md_consolidate.py` is itself in the runtime's `_BLOCKED_EXACT_PATHS`: both proposal sizing and the integration gate reject an autonomous edit to the operator's removal mechanism. Raising the system-prompt cap is not part of the decision.
+
+The product gate's `if f == 'AGENTS.md'` check governs only a repository-root path. The growth measured by #1188 occurred in the separate instance repository; that instance `AGENTS.md` is governed by the instance-side path rules delivered with #1193, not by treating every nested `AGENTS.md` in this product repository as operator-owned.
+
+# Consequences
+
+## What gets easier
+
+The prompt ledger reports the remaining droppable reserve directly, and an operator can convert declared reserve into permanent file reduction without hand-editing section boundaries.
+
+## What gets harder
+
+Consolidation requires an operator to choose sections deliberately. The tool cannot decide semantic redundancy and will not remove a critical section even when the operator believes it is obsolete; the marker must first be reviewed and added in the instance repository.
+
+## What does not change
+
+Critical sections are never dropped by prompt fitting or removed by the consolidation tool. The autonomous loop keeps learning through skills, lessons, and memory, but cannot mutate `AGENTS.md`.
+
+# Alternatives considered
+
+- **Autonomous age- or size-based deletion:** rejected because age and size do not encode importance and the gate cannot distinguish redundant prose from a security invariant.
+- **Automatic deletion of every section dropped from a prompt:** rejected because prompt-fit choice is a runtime budget decision, not authorization to mutate the source file.
+- **Raise `NANOBOT_SYSTEM_PROMPT_MAX_CHARS`:** rejected because it moves the failure date while leaving append-only growth and invisible reserve consumption intact.
+- **Do nothing:** rejected because the measured reserve has a finite exhaustion date.
+
+# Test Contract
+
+| Claim | Test | Status |
+|---|---|---|
+| Remaining declared-droppable characters are recorded after no, partial, and complete exhaustion | `tests/test_context_prompt_fit.py` reserve tests | passing |
+| Both successful and overflow ledger rows carry `droppable_reserve_chars` | `tests/test_bridge_system_prompt_overflow.py` | passing |
+| Dry-run never writes and `--apply` removes only explicitly named marked sections | `tests/test_agents_md_consolidate.py` | passing |
+| Missing, duplicate, or unmarked headings reject without partial mutation | `tests/test_agents_md_consolidate.py` refusal tests | passing |
+| Autonomous root `AGENTS.md` mutation remains rejected | `tests/test_mutation_surfaces.py`, `tests/test_llm_proposer.py` | passing |
+| Autonomous edits to `agents_md_consolidate.py` are rejected by proposal sizing and integration policy | `tests/test_mutation_surfaces.py::test_agents_md_consolidate_script_is_immutable`, `tests/test_llm_proposer.py::TestValidateSizing::test_rejects_agents_md_consolidate_script`, `tests/test_runtime_slice.py::test_bridge_policy_mirrors_stay_synced_with_gate` | passing |
+
+# Rollback
+
+Revert the product commit to remove the metric and operator tool. An applied instance consolidation is separately reversible through the instance repository's git history.
+
+# References
+
+#1188, #1193, #1300, #1313; `docs/specs/subagent-bridge/spec.md`.
--- a/docs/adr/README.md
+++ b/docs/adr/README.md
@@ -18,3 +18,4 @@ Conventions:
 |---|---|---|
 | [ADR-001](ADR-001-agent-architecture.md) | Agent architecture — roles, tools, models, and budgets of the eeebot runtime | accepted |
 | [ADR-002](ADR-002-bounded-state-access.md) | Bounded state access | accepted |
+| [ADR-003](ADR-003-operator-owned-agents-consolidation.md) | Keep AGENTS.md operator-owned and consolidate only declared-droppable sections | proposed |
--- a/docs/specs/subagent-bridge/spec.md
+++ b/docs/specs/subagent-bridge/spec.md
@@ -172,9 +172,19 @@ passes — so a broken or unverified cycle never reaches `main`.
 
 ### Skill mutation and first-hand usage fitness (#939)
 
-The script tier SHALL allow `skills/**` and the instance-root `AGENTS.md` while
-continuing to reject the immutable `goals.md` charter. Proposer and bridge gate
-mirrors SHALL agree on these paths.
+The script tier SHALL allow `skills/**` while continuing to reject the
+immutable `goals.md` charter. Proposer and bridge gate mirrors SHALL agree on
+these paths.
+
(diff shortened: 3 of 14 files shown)


ozand:
**Verdict: approve with one required change.** This is the right shape, and the measurement in it corrected the issue twice — both times against me. One gap: the PR closes the write path and then puts the delete path somewhere the loop can write.

## Required: `scripts/agents_md_consolidate.py` is inside the instance's mutation surface

The reasoning for the operator-only arm is correct — the autonomous loop must not gain deletion authority over the file holding its charter and guardrails, and a prose-blind gate cannot tell consolidation from removing a safety invariant. But the tool that does the deleting landed at a path the loop is allowed to edit:

```python
_ALLOWED_PATH_PREFIXES = ('surfaces/', 'scripts/', 'memory/', 'lessons/', 'docs/', 'tests/', 'skills/')
_GATE_EXT_ALLOWLIST    = frozenset((".py", ".md", ...))
```

`scripts/` is an allowed prefix and `.py` is an allowed extension, so a cycle can propose and auto-integrate a change to this file. The loop cannot *run* it — that part holds — but it can change what it does when the operator runs it. Every guard the PR relies on lives in the script itself: the exact-marker requirement, the refuse-on-unmarked/duplicate/missing rule, the all-or-nothing behaviour, the `--apply` default. All of them are instance-editable.

That is the failure mode from #1188/#1193 pointed the other way. #1193 closed the write path to `AGENTS.md`; this PR must not open a delete path by proxy. It does not need new machinery — one entry:

```python
_BLOCKED_EXACT_PATHS = frozenset({'goals.md', 'IDENTITY.md', 'agents_md_consolidate.py'})
```

`_validate_mutation_surfaces` already checks both basename and full path (`if fname in blocked_exact_paths or f in blocked_exact_paths`), so either spelling works. Add a test that a proposed change to it is rejected, so the protection is observable rather than assumed — the whole reason #1188 is still open at `status:test` is that its guard has never once fired and so cannot be distinguished from an absent guard.

Note also that `gate.py`'s check is `if f == 'AGENTS.md'` — an exact match on the root path only. Worth a sentence in ADR-003 stating that the instance repository's own `AGENTS.md` is governed by the instance-side path rules rather than this one, since the growth this issue is about happened in the instance file, not the root.

## Your measurement corrected mine, twice, and both corrections were right

Confirmed both on the host before writing this:

- **Reserve.** `grep -c -- "<!-- prompt-fit: droppable -->"` gives 13; the loose `grep -c "prompt-fit: droppable"` I used gives 14. The extra hit is line 27, prose *describing* the marker inside the section the loop added this morning. The runtime's 2,242 is right and my 2,948 counted that section's own 706 bytes. `_droppable_reserve_chars` uses `self.DROPPABLE_MARKER`, so it is correct by construction, which is the strongest form of this fix.
- **Growth.** I claimed ~1,400 bytes/day and asserted the flat days were a deploy freeze. The instance log shows the last autonomous `AGENTS.md` append is `4dfc6ee4` on 2026-09-01 19:34, then nothing on the 2nd, 3rd or 4th — across **334 commits** on the branch. That is #1193 working, and it is why "preserve #1193 rather than add a second guard" was the right call. I had built the urgency on the opposite assumption. Both corrections are now on the issue.

## Confirmed by reading

- The ledger row uses `_prompt_fit.get('droppable_reserve_chars')`, so the non-strict path leaving the key unset cannot raise. Deliberate or not, it is safe.
- `SystemPromptOverflowError.droppable_reserve_chars` defaults to 0 and the docstring says why it is always 0 on the real overflow path — the fit loop only gives up after exhausting every droppable section. Recording a known-zero explicitly rather than omitting the key is the right choice: an absent key and a zero mean different things to a later reader.
- Test-first evidence quotes the actual pre-change failure (`TypeError: ... unexpected keyword argument 'reserve_chars'`) rather than asserting it.
- Full suite 4 failed / 3108 passed, four by name.
- Rejected arms recorded: automatic age/size deletion, deleting every prompt-omitted section, and raising the cap. Agreed on all three, and the cap point is the one that most needed writing down.
- The symlink-replacement flaw caught in independent review and fixed with `path.resolve(strict=True)` before the atomic write — that is the bug that would have silently replaced the operator's real file with a regular one.

## On the deployment note

Holding the instance-side consolidation for a separately reviewed change with dry-run evidence is correct and I am not asking you to shorten it. When the rollout check runs, quote the live row's two numbers — `cap - chars` and `droppable_reserve_chars` — rather than a single "it fits", since the whole point of the field is that one number was never enough.

ozand:
Required review change implemented at `2a2bffd1` and force-pushed to the same PR branch.

### Protection added

`agents_md_consolidate.py` is now in `_BLOCKED_EXACT_PATHS` in all three required policy mirrors:

- `nanobot/runtime/gate.py`
- `nanobot/runtime/bridge.py`
- `nanobot/runtime/llm_proposer.py`

Because the existing classifiers check both basename and full relative path, `scripts/agents_md_consolidate.py` is rejected at proposal sizing and again at the integration gate. The operator-only removal mechanism is therefore no longer writable by the autonomous instance.

### Test-first evidence

Before adding the constant, the three focused assertions failed exactly as intended:

```text
FAILED test_agents_md_consolidate_script_is_immutable
  assert [] == ['immutable file blocked from mutation: scripts/agents_md_consolidate.py']
FAILED test_bridge_policy_mirrors_stay_synced_with_gate
  assert 'agents_md_consolidate.py' in frozenset({'IDENTITY.md', 'goals.md'})
FAILED TestValidateSizing::test_rejects_agents_md_consolidate_script
  assert True is False
```

After the fix:

```text
274 passed in 87.25s
```

This set includes mutation surfaces, policy mirror parity, proposer sizing, consolidation CLI, prompt fitting, and bridge overflow tests.

ADR-003 now explicitly states both:

1. the consolidation script itself is immutable to the loop at proposal and integration time;
2. the product gate's exact root `AGENTS.md` check is distinct from the separate instance repository, whose `AGENTS.md` is governed by #1193's instance-side path policy.

Independent follow-up review verdict: **OK; no issues found**.

The separately reviewed instance consolidation with dry-run evidence remains deferred exactly as requested. No host or rollout mutation was made.

Issue #1039, opened 2026-08-26
state retention: purge coordinator-era directories, promotions rotation, subagent archive deletion, llm_calls record cap


ozand:
## Context (self-contained)

`state/` on the eeepc host has grown to 1.1 GB (2GB-RAM host, slow disk). Audit + live inspection (2026-08-26):

| Path | Size | Writer | Pruner |
|---|---|---|---|
| `state/goals/history` | 176 MB / 8,759 files | NONE (froze at coordinator decommission #900/#910 — `state.py:906-913` says so itself) | NONE |
| `state/experiments` | 98 MB / 8,961 files (oldest 2026-04-21, newest 2026-08-22) | NONE (same decommission) | NONE |
| `state/strong_reflection` | 35 MB / 8,629 files (newest 2026-08-21) | NONE (same decommission; its test deleted in commit 72b0b6e0) | NONE |
| `state/promotions` | 106 MB / 17,801 files | `bridge.py:3406-3408` per green runtime-slice cycle + promotion verifier | NONE |
| `state/subagents/archive` | 41 MB / ~5,800 files | three independent "cleanup" scripts, all move-only (`rename()` into archive/), none ever delete | NONE |
| `state/llm_calls/prompts` | 639 MB | `llm_telemetry.py:205-274` — full messages/content per LLM call, no per-record size cap | rotation EXISTS (gzip + 14-day retention, `llm_telemetry.py:167-202`) — the gap is record size, not rotation |

Additional defect: `state.py:1083` prefers `promotions/latest.json`, which is never written outside tests — so every `load_runtime_state_from_root` falls back to a full glob+stat sort over all 17,801 files.

## Task

1. **One-time cleanup (operator-approved before execution):** archive-then-delete `goals/history`, `experiments`, `strong_reflection` (309 MB of coordinator-era fossils). Back up (tar to /opt or offline) before deleting. Remove the now-dead read paths that probe them (`state.py` fallbacks, `scripts/compile_project_lessons.py` input).
2. **promotions/ retention:** mirror `cycle_ledger.py`'s 90-day gzip+prune pattern; either write `promotions/latest.json` from the live writer or delete the dead fast-path branch in `state.py:1083`.
3. **subagents/archive retention:** age-based deletion pass (e.g. 30 days) in the ONE systemd-wired cleaner (`scripts/cleanup_subagent_queue.py`).
[...]

PR #1322, opened 2026-09-05
fix(telemetry): trim a prompt record to the cap instead of clamping every message to 1,000 chars (#1319)

### LINKED ISSUE #1319: The prompt recorder clamps every message to 1,000 chars on any overage: a record one byte over the 32 KiB cap loses ~97% of its content, and the reflector analyses what's left
Surfaced by #1314's new `input_fit` field on its first live rows. The prompt recorder's size cap does not trim a record to fit — it clamps every message to 1,000 characters the moment the record exceeds the cap by any amount.

## What the live rows show

First three reflection rows on the deployed release `20260905T121040Z-canonical-9a67951e`, all on `cycle-59bedf8550bc`:

```json
transcript {"budget": 32768, "total": 2, "kept": 2, "dropped": 0, "recorder_truncated": true, "chars": 2495}
```

The reflector's own bound dropped nothing — `dropped: 0`, both messages kept. The record arrived at 2,495 characters against a 32,768 budget, already carrying `truncated: true`. So the loss happened upstream, in `record_llm_prompt`, before the reflector ever saw it.

## The mechanism

`nanobot/observability/llm_telemetry.py:209`, `_cap_payload_record`, first pass:

```python
serialized = json.dumps(record, ensure_ascii=False, default=str)
if len(serialized.encode("utf-8")) <= max_bytes:
    return record

rec = dict(record)
rec["truncated"] = True

messages = rec.get("messages")
if isinstance(messages, list):
    for msg in messages:
        m_content = m_copy.get("content")
        if isinstance(m_content, str) and len(m_content) > 1000:
            m_copy["content"] = m_content[:1000] + "…[truncated]"
```

The 1,000-character clamp is a fixed constant applied to every message. It is not derived from how far over the cap the record is. A record at 32,769 bytes — one byte over — takes the same clamp as a record at 2 MB.

The live row is that case: two messages, each large, clamped to 1,000 each, producing 2,495 characters total. The record needed to lose a few hundred bytes at most and lost roughly 30 KB.

## Why it matters here specifically

The reflector's system prompt asks the model to "analyze every message, skill, command, tool call/result, error, retry, and detour in this completed cycle." It is being handed 2,495 characters to do that with, and every finding, recommendation and `followed_previous` verdict built on those reflections inherits the gap.

This is upstream of #1314 and was invisible before it: the reflector had no field that distinguished "the transcript was small" from "the transcript was gutted." Now it does, and the first row it wrote says gutted.

## What is not being claimed

Three rows on one cycle is not a distribution. The first task here is to measure how often `recorder_truncated` is true and what `transcript.chars` looks like across a day, using the field #1314 added — not to assume every cycle looks like this one.

Nor is the cap itself wrong. `MAX_LLM_PROMPT_PAYLOAD_BYTES = 32 * 1024` exists because a runaway subprocess printer was previously bounded only by the unit's `MemoryMax`, i.e. by an OOM kill (#1039). The bound should stay. The complaint is about how the bound is met.

## Acceptance criteria

- [ ] Measure first: across at least a day of `llm_calls` prompt records, how many carry `truncated: true`, and what the size distribution looks like before and after capping. State whether the live row above is typical or an outlier. If it is an outlier, say so and close this — a defect that fires on one cycle in a hundred is a different decision than one that fires on all of them.
- [ ] The trim becomes proportional to the overage: a record one byte over the cap loses about one byte, not 97% of its content. Whatever the algorithm is, state the property it guarantees.
- [ ] The 32 KiB bound is preserved. Do not raise it as the fix — the OOM reasoning behind #1039 has not changed.
- [ ] The record says how much it lost, not only that it lost something. `truncated: true` is a boolean where a reader needs a number; the reflector's `input_fit` can then carry it through the way it now carries the flag.
- [ ] A test drives a record just over the cap and asserts the result is close to the cap rather than close to the per-message clamp. It must fail on current `main`.

## Note for whoever takes this

`_cap_payload_record` does not tear JSON — it shortens values in place and re-serializes, so the record stays parseable at every pass. That part is correct and should not change. The five passes are also correctly ordered, from gentlest to the emergency clamp. The defect is entirely in the first pass's constant.

ozand:
Closes #1319.

## Measurement first — the live row is the rule, not an outlier

Read-only on the host as `eeepc-agent`, 2026-09-05, every retained prompt record since the cap existed (2026-08-28 → 2026-09-05, 9 days, 13,924 records):

| what | truncated | share |
|---|---|---|
| all prompt records | 12,566 / 13,924 | **90.2%** |
| `bridge` (executor) records | 12,539 / 12,721 | 98.6% |
| `proposer` records | 0 / 1,168 | 0% |
| **latest record per cycle — what the reflector reads** | **581 / 582** | **99.8%** |

Per day the executor share never dipped below 84% (2026-08-28: 1,860/2,122 … 2026-09-05: 560/661); latest-per-cycle was 100% on eight of nine days and 20/21 on the ninth. Which pass fired: 7,917 records stopped at pass 1 (every message clamped to 1,000 chars), 4,649 (37%) fell through to pass 2 (first + last message only). Size after the cap across those 12,566 records: min 1,532, **median 6,652**, max 32,760 bytes — a median of 20% of the budget. `prompt_tokens` (the only pre-cap figure the record keeps; 4.27 B/token on clean records) puts the median executor record at ~140 KB before the cap, p95 ~300 KB.

Then the before/after comparison the issue asked for, on real pre-cap records: the 2026-08-25 archive predates the cap, so its records are originals. First 354 of them (349 bridge, 5 proposer; pre-cap median 173,404 bytes, max 311,076; 334 over the cap), run through the old and the new `_cap_payload_record` locally:

```
                      old (main)                 new (this PR)
after cap, p50        2,738 bytes  (8% of cap)   32,757 bytes  (100% of cap)
after cap, p10 / p90  1,623 / 24,669             32,697 / 32,764
messages kept, p50    3                          49
whole messages dropped (pass 2+)   —             0 of 334
truncated_chars, p50  (not recorded)             138,152
```

So the record that the issue's live row describes — arriving at the reflector as 2,495 chars with everything kept and nothing dropped by #1314's bound — is what **every** executor cycle has looked like since 2026-08-28. The reflector has been analysing ~2–7 KB of a ~140 KB transcript on 99.8% of cycles. The `prompt_tokens` estimate also corrects the issue's framing in one respect: records are not typically "one byte over", they are typically 4× over. The defect is the same (a constant where a fit was needed) and the fix is the same, but the median record needs to lose 75% of its content no matter how well the cap is met — the proportional trim raises what survives from 8% of the budget to 100% of it, not to 100% of the transcript.

## The fix (`nanobot/observability/llm_telemetry.py`)

**Property, stated where it is implemented (`_level_cut` docstring):** every string leaf under `messages` / `content` / `reasoning_content` longer than a level `L` is cut to `L` and suffixed `…[truncated N chars]`; strings at or below `L` are untouched; `L` is the **largest** level at which the whole record serializes within the cap, found by bisection on the actual serialized size (so bytes-per-character and marker overhead are exact). The record therefore loses only the overage plus the ~70 bytes of bookkeeping that records the loss, taken from its longest strings first. A record one byte over loses ~70 characters, not 97% of its content. If even `L = 32` does not fit, that level is applied and the structural passes do the rest.

- The 32 KiB bound is unchanged. In-place shortening + re-serialize is unchanged (the record stays parseable at every pass). Passes 2–5 are unchanged in order and text; they now fire only when the *structure* alone exceeds the budget (thousands of tiny messages), which happened on 0 of 334 real records.
- The first pass covers every string leaf under the three fields, so a 60 KB `tool_calls[].function.arguments` (a `write_file` body) is levelled with the rest instead of forcing pass 2. The old first pass only saw `content`.
- `messages` / `content` / `reasoning_content` are deep-copied before shortening. The old code shallow-copied each message dict; a nested `tool_calls` list was shared with the executor's live conversation. Nothing was mutated before because only `content` was rewritten, but levelling reaches inside, so the copy is now required and tested.
- **The record carries a number:** `truncated_chars` (content characters removed, marker text excluded) beside the unchanged `truncated: true`. Computed exactly as content-chars-before minus content-chars-after, whichever passes fired. A placeholder as wide as any real count is set before the fit so the key is budgeted for.
- Marker text changes from `…[truncated]` to `…[truncated N chars]`; the one existing assertion on it is updated.

**`nanobot/runtime/reflector.py`:** `_fit_transcript` copies `truncated_chars` into `input_fit.transcript.recorder_truncated_chars` and the label line adds `, N chars removed` — the loop #1315 opened is closed: the row says how much the recorder took, not only that it took.

**Cost, measured on the host** (review question). Production interpreter (`/opt/eeepc-agent/venv/bin/python`, 3.11.2, i686), 120 real pre-cap records from the 2026-08-25 archive, median 101,579 bytes, old and new implementation timed on the same records:

```
one json.dumps + utf-8 encode of the record   p50  9.2 ms   max  32.7 ms
old cap (1,000-char clamp) per record         p50 12.0 ms   p90  22.6   max  47.7 ms
new cap (bisection) per record                p50 77.9 ms   p90 123.6   max 192.3 ms
added per truncated record                    p50 65.8 ms
```

About 17 serializations per record, as the bisection predicts. It is paid once per LLM call, inside `record_llm_prompt`, on ~90% of executor calls; the call it follows waits seconds on the gateway, so this is on the order of 1–3% of call latency. Recorded here and in the `_level_cut` docstring so the next reader does not re-derive it. Not mitigated, deliberately: the measurement says it does not matter yet, and the cheap lever if it ever does is the number of probes, not the property. (Dev machine for comparison: 15.6 ms per record.)

**Double loss at the floor, intentional.** When even `L = _CAP_MIN_KEEP_CHARS = 32` does not fit, the record takes both losses: near-total levelling, then the structural passes drop messages. That happens only when the structure alone exceeds the budget (thousands of tiny messages); on the 334 real over-cap records it happened 0 times. Stated in the constant's comment.

### Rejected arms, recorded

- **Raise the cap.** No — #1039's reason (a runaway printer bounded only by `MemoryMax`) has not changed.
- **Scale every string by the same ratio.** Rejected: it cuts short tool results and role text proportionally with the big ones, so small messages lose their meaning while the large one still dominates. Levelling protects everything below `L` entirely and takes only from the longest — the reflector needs to see *every* step, and steps are mostly small.
- **Drop whole oldest messages** (the reflector's own #1314 rule). Rejected for the recorder: it is the only durable copy of the conversation, and "every message present, long ones shortened" is the better archival property than "some messages missing". Pass 2 already exists for the emergency and now fires on 0 of 334 real records.
- **Estimate the level from character counts instead of bisecting on bytes.** Tried first; multibyte text and marker overhead made the estimate collapse to the floor and fall through to pass 2 on the test data. Replaced with exact bisection.

## Tests — fail on `main`, verified

`tests/test_llm_prompt_cap_proportional.py` (7): run against the unmodified module first —

```
FAILED test_record_well_over_the_cap_lands_near_the_cap_not_near_the_clamp   assert 2364 >= (32768 - 2048)
FAILED test_record_one_byte_over_loses_about_one_byte                        KeyError: 'truncated_chars'
FAILED test_overage_is_taken_from_the_longest_strings_and_short_ones_stay_whole   endswith(' chars]') False
FAILED test_multibyte_content_still_fits_and_still_uses_most_of_the_budget   assert 4364 >= (32768 // 2)
FAILED test_structural_overhead_beyond_the_budget_falls_through_to_the_old_passes  KeyError: 'truncated_chars'
5 failed, 2 passed
```

The two that pass on `main` are guard tests and are labelled as such: the caller's messages are not mutated (true before, must stay true now that levelling reaches nested lists) and an under-cap record is returned untouched.

- two 20 K messages → lands within 2 KiB of the cap, not at 2 × 1,000; `truncated_chars` equals content-before minus content-after
- exactly one byte over → loses 1–96 chars (the ~70 of bookkeeping), the shorter message untouched
- 30 K system + short task + small `tool_calls` + 8 K tool result → short ones byte-identical, all four roles present, pool levelled
- Cyrillic (2 B/char) → fits and uses ≥ half the budget (was 4,364 bytes)
- 3,000 tiny messages → structure alone exceeds the cap → falls through to the old passes, bound and count hold
- `tests/test_reflector_input_fit.py` +1: `recorder_truncated_chars` carried into the fit and the label

`tests/test_llm_prompt_recording.py` / `test_llm_telemetry.py` / `test_reflector*.py`: 73 passed together. Ruff clean on the changed files.

Full Windows suite at `002c1a32` (on `9a67951e`); the follow-up `72d597dc` is comment-only (13 lines):

```
4 failed, 3114 passed, 17 skipped, 11 warnings in 1370.00s (0:22:50)
```

Exactly the four Windows-baseline failures, matched by name.

## Not done, on purpose

- **`prompt_tokens` × 4.27 as the pre-cap size.** It is a proxy; the pre-cap byte size is not recorded and adding it would be a second number on the record for a one-time measurement. The 2026-08-25 archive gave real originals instead.
- **Reflector telemetry blindness** (`component: reflector` in `llm_calls`): untouched, separate defect.
- `nanobot/agent/context.py`, `gate.py`, `state_paths.py`, `validator_harness.py`, `scripts/eeebot_dashboard.py`: untouched.

First reflector run after the deploy: `input_fit.transcript.chars` on new `reflections.jsonl` rows should sit near 32,000 (was 2,495 on the three pre-fix rows), `recorder_truncated: true` with `recorder_truncated_chars` in the ~100,000 range on a normal executor cycle. A row with `chars` in the low thousands and `recorder_truncated_chars` absent means the deployed release predates this PR.

### DIFF
--- a/nanobot/observability/llm_telemetry.py
+++ b/nanobot/observability/llm_telemetry.py
@@ -27,6 +27,7 @@ actually be inspected (see ``scripts/llm_prompt_inspect.py``).
 from __future__ import annotations
 
 import contextlib
+import copy
 import gzip
 import json
 import os
@@ -206,46 +207,116 @@ def _rotate_and_prune(prompts_dir: Path, today: str, retention_days: int) -> Non
 MAX_LLM_PROMPT_PAYLOAD_BYTES = 32 * 1024  # 32KB per-record cap (#1039)
 
 
+# #1319 -- how the cap is met. Until #1319 the first pass clamped every message
+# to a fixed 1,000 chars on any overage, so a record one byte over lost ~97% of
+# its content; on the host that left a median 6.6 KB of the 32 KiB budget and
+# 581 of 582 reflector inputs were cut that way. Passes 2-5 below are unchanged
+# and now fire only when the record's structure alone (message count, keys)
+# exceeds the budget.
+_CAP_FIELDS = ("messages", "content", "reasoning_content")
+_CAP_MIN_KEEP_CHARS = 32  # a string levelled below this is noise; fall through to pass 2 instead
+_CAP_MARKER_RE = re.compile(
+    r"…\[truncated \d+ chars\]|…\[truncated\]|…\[intermediate messages omitted\]|…\[payload truncated to fit 32KB budget\]"
+)
+
+
+def _string_leaves(value: Any, out: list[tuple[Any, Any, str]], container: Any = None, key: Any = None) -> None:
+    """Collect ``(container, key, string)`` for every string leaf so it can be replaced in place."""
+    if isinstance(value, str):
+        if container is not None:
+            out.append((container, key, value))
+    elif isinstance(value, dict):
+        for k, v in value.items():
+            _string_leaves(v, out, value, k)
+    elif isinstance(value, list):
+        for i, v in enumerate(value):
+            _string_leaves(v, out, value, i)
+
+
+def _content_chars(value: Any) -> int:
+    """Characters of string content in ``value``, not counting the recorder's own markers."""
+    if isinstance(value, str):
+        return len(_CAP_MARKER_RE.sub("", value))
+    if isinstance(value, dict):
+        return sum(_content_chars(v) for v in value.values())
+    if isinstance(value, list):
+        return sum(_content_chars(v) for v in value)
+    return 0
+
+
+def _level_cut(rec: dict[str, Any], max_bytes: int) -> None:
+    """Pass 1 of :func:`_cap_payload_record`: level the string leaves under :data:`_CAP_FIELDS`.
+
+    Property guaranteed: every string longer than a level ``L`` is cut to ``L``
+    and suffixed with ``…[truncated N chars]``; strings at or below ``L`` are
+    untouched; ``L`` is the LARGEST level at which the whole record serializes
+    within ``max_bytes`` (found by bisection on the actual serialized size, so
+    bytes-per-character and marker overhead are accounted for exactly). Hence
+    the record loses only the overage plus the bookkeeping that records it
+    (the ``truncated`` / ``truncated_chars`` keys and one marker per cut
+    string, ~70 bytes), taken from its longest strings first: a record one
+    byte over loses ~70 characters, never a fixed fraction. If even
+    ``L = _CAP_MIN_KEEP_CHARS`` does not fit, that level is applied and the
+    structural passes that follow do the rest.
+    """
+    leaves: list[tuple[Any, Any, str]] = []
+    for field in _CAP_FIELDS:
+        if field in rec:
+            _string_leaves(rec[field], leaves, rec, field)
+    lengths = [len(s) for _, _, s in leaves if len(s) > _CAP_MIN_KEEP_CHARS]
+    if not lengths:
+        return
+
+    def apply(level: int) -> None:
+        for container, key, s in leaves:
+            container[key] = s[:level] + f"…[truncated {len(s) - level} chars]" if len(s) > level else s
+
+    def fits(level: int) -> bool:
+        apply(level)
+        return len(json.dumps(rec, ensure_ascii=False, default=str).encode("utf-8")) <= max_bytes
+
+    low, high = _CAP_MIN_KEEP_CHARS, max(lengths)  # the record is known not to fit at `high` (nothing cut)
+    if fits(low):
+        while high - low > 1:  # largest level that fits; fits() is monotone in the level
+            mid = (low + high) // 2
+            if fits(mid):
+                low = mid
+            else:
+                high = mid
+    apply(low)
+
+
 def _cap_payload_record(record: dict[str, Any], max_bytes: int = MAX_LLM_PROMPT_PAYLOAD_BYTES) -> dict[str, Any]:
     """Ensure record JSON serialization does not exceed max_bytes.
 
-    If the serialized record exceeds max_bytes, trims heavy string/list fields
-    (messages, content, reasoning_content) and adds ``truncated: True`` to the record (#1039).
+    If the serialized record exceeds max_bytes, shortens string values in place
+    under ``messages`` / ``content`` / ``reasoning_content`` -- never slicing
+    the JSON text, so the record stays parseable at every pass (#1039) -- and
+    adds ``truncated: True`` plus ``truncated_chars`` (characters of content
+    removed, markers excluded) so a reader learns how much was lost, not only
+    that something was (#1319). Pass 1 (:func:`_level_cut`) removes only the
+    overage, from the longest strings first; passes 2-5 are the structural
+    fallbacks, gentlest first, and fire only when levelling alone cannot fit
+    the record.
     """
     serialized = json.dumps(record, ensure_ascii=False, default=str)
     if len(serialized.encode("utf-8")) <= max_bytes:
         return record
 
     rec = dict(record)
+    for field in _CAP_FIELDS:
+        if field in rec:
+            rec[field] = copy.deepcopy(rec[field])  # the caller's live messages must not be shortened
     rec["truncated"] = True
+    # Placeholder as wide as any real count, so the fit below budgets for the key; the true value is shorter.
+    rec["truncated_chars"] = 999_999_999
 
-    # First pass: trim message content
-    messages = rec.get("messages")
-    if isinstance(messages, list):
-        trimmed_msgs = []
-        for msg in messages:
-            if isinstance(msg, dict):
-                m_copy = dict(msg)
-                m_content = m_copy.get("content")
-                if isinstance(m_content, str) and len(m_content) > 1000:
-                    m_copy["content"] = m_content[:1000] + "…[truncated]"
-                trimmed_msgs.append(m_copy)
-            else:
-                trimmed_msgs.append(msg)
-        rec["messages"] = trimmed_msgs
-
-    # Trim reasoning_content if large
-    reasoning = rec.get("reasoning_content")
-    if isinstance(reasoning, str) and len(reasoning) > 1000:
-        rec["reasoning_content"] = reasoning[:1000] + "…[truncated]"
-
-    # Trim content if large
-    content = rec.get("content")
-    if isinstance(content, str) and len(content) > 2000:
-        rec["content"] = content[:2000] + "…[truncated]"
+    # First pass: level the string content down to the budget (#1319)
+    _level_cut(rec, max_bytes)
 
     serialized = json.dumps(rec, ensure_ascii=False, default=str)
     if len(serialized.encode("utf-8")) <= max_bytes:
+        rec["truncated_chars"] = _content_chars(record) - _content_chars(rec)
         return rec
 
     # Second pass: aggressively trim intermediate messages
@@ -282,6 +353,7 @@ def _cap_payload_record(record: dict[str, Any], max_bytes: int = MAX_LLM_PROMPT_
                 elif isinstance(rec[k], (list, dict)):
                     rec[k] = "…[truncated]"
 
+    rec["truncated_chars"] = max(0, _content_chars(record) - _content_chars(rec))
     return rec
 
 
--- a/nanobot/observability/llm_telemetry.py
+++ b/nanobot/observability/llm_telemetry.py
@@ -214,7 +214,12 @@ MAX_LLM_PROMPT_PAYLOAD_BYTES = 32 * 1024  # 32KB per-record cap (#1039)
 # and now fire only when the record's structure alone (message count, keys)
 # exceeds the budget.
 _CAP_FIELDS = ("messages", "content", "reasoning_content")
-_CAP_MIN_KEEP_CHARS = 32  # a string levelled below this is noise; fall through to pass 2 instead
+# A string levelled below this is noise, so when even this level does not fit
+# the record takes BOTH losses on purpose: near-total levelling here, then the
+# structural passes drop messages. That only happens when the structure alone
+# (message count, keys) exceeds the budget; on 334 real over-cap records from
+# the 2026-08-25 archive it happened 0 times.
+_CAP_MIN_KEEP_CHARS = 32
 _CAP_MARKER_RE = re.compile(
     r"…\[truncated \d+ chars\]|…\[truncated\]|…\[intermediate messages omitted\]|…\[payload truncated to fit 32KB budget\]"
 )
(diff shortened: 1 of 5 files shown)


ozand:
**Verdict: approve, pending 3.11 CI and one answer about cost.** The measurement is the best part and it corrected my framing of the issue. The algorithm is right and its property is stated where a reader will find it.

## The measurement answered the question I actually asked

I asked for frequency before a fix and said an outlier would be a complete delivery. It is not an outlier:

```
latest record per cycle — what the reflector reads    581 / 582    99.8%
bridge (executor) records                          12,539 / 12,721  98.6%
proposer records                                        0 / 1,168     0%
```

Nine days, 13,924 records, never below 84% on any single day. The live row in #1319 is what every executor cycle has looked like since the cap existed.

And the before/after on real pre-cap records from the 2026-08-25 archive is the right way to show a fix works — original inputs, both implementations, same population:

```
after cap, p50     old  2,738 bytes (8% of cap)    new  32,757 bytes (100% of cap)
messages kept p50  old  3                          new  49
whole messages dropped (pass 2+)                   new  0 of 334
```

## And it corrected my framing, which is worth saying plainly

I wrote #1319 around "a record one byte over loses ~97% of its content." That is the pathological illustration, not the typical case. Your `prompt_tokens` reconstruction puts the median executor record at ~140 KB before the cap — **4× over**, not one byte over. So the median record must lose ~75% of its content no matter how perfectly the cap is met, and this PR does not restore the transcript; it fills the budget instead of leaving 92% of it empty.

That is a smaller claim than my issue implied and a more accurate one. Good that the PR states it rather than letting the issue's framing stand.

## Read and confirmed correct

- `apply(low)` is called unconditionally after the bisection loop. I went looking for the bug where the last `fits(mid)` leaves the record clamped at a non-fitting level; the final `apply(low)` closes it. `fits()` is monotone in the level and `apply` always rebuilds from the originals captured in `leaves`, so it is idempotent across probes.
- `copy.deepcopy` on the three cap fields before mutating. Necessary once leaves are replaced in place, and the comment says why — the caller's live messages must not be shortened.
- `rec["truncated_chars"] = 999_999_999` set as a placeholder *before* the fit, so the bisection budgets for the key at its widest and the real value can only be shorter. That is the kind of detail that would otherwise produce a record one digit over the cap.
- `_content_chars` strips the recorder's own markers via `_CAP_MARKER_RE` before counting, so `truncated_chars` measures content lost rather than bookkeeping added.
- The reflector carries it through: `recorder_truncated_chars` in `input_fit`, and the label line now says how much was removed. That closes the loop #1314 opened — the flag became a number at both layers.

## The one question: bisection cost on the host

`fits()` serializes the entire record each probe, and the loop runs about log2(max string length) times — call it 18 probes for a 300 KB record. That is up to ~18 full `json.dumps` plus UTF-8 encodes of a large structure, per capped record, and 90.2% of records are capped.

This runs on an i386 host with 2 GB of RAM, inside the recording path of every LLM call. Did you measure the added wall time per record? A number from the host, even a rough one, is what I need — if it is single-digit milliseconds this is a non-issue and should be said so in the PR body; if it is tens of milliseconds per record it is still probably fine but should be recorded next to the cap's own rationale, because the next person to touch this will otherwise have to re-derive it.

If it turns out to matter, the cheap mitigation is bounding the probe count rather than abandoning the property — but do not do that pre-emptively. Measure first, exactly as you did for the defect itself.

## Smaller note, not blocking

`_CAP_MIN_KEEP_CHARS = 32`: when even that level does not fit, the docstring says the level is applied and the structural passes do the rest — so such a record takes both a near-total levelling and message drops. Your own measurement says this never happened across 334 real records (0 dropped at pass 2+), so it is a cold path and the choice is defensible. Worth one sentence saying that the double loss is deliberate and how often it fired in the sample, so a future reader does not read it as an oversight.

ozand:
## Cost answered — accepted, and here is the arithmetic so nobody re-derives it

Measured on the host: **p50 78 ms, p90 124 ms, max 192 ms** per truncated record, against 12 ms for the old fixed clamp.

That is 6.5× and it is fine. The number that settles it is throughput, not the ratio:

```
13,924 records / 9 days          ≈ 1,547 per day ≈ 64 per hour
90.2% truncated                  ≈ 58 capped records per hour
58 × 78 ms                       ≈ 4.5 seconds of extra CPU per hour
```

Under two minutes a day on a host that runs 9–13 cycles an hour. And per call it sits inside an LLM round trip measured in seconds, so 78 ms is low single-digit percent of the operation it is attached to — it cannot become the bottleneck without the LLM already having stopped being one.

Not mitigating pre-emptively was the right call, and naming the lever without pulling it is exactly what I asked for. If this ever does need bounding, the probe count is the knob and the property survives a bounded bisection — it just lands on a slightly smaller level than the true maximum, which loses a little content rather than the guarantee.

Worth keeping in the docstring next to the cap's own rationale, which you have done. The reason to write it down is that 6.5× reads alarming out of context and 4.5 seconds an hour does not; a future reader will meet the ratio first.

Approved on cost. Merge once the full suite tail is in the body with the four baseline failures by name.

ozand:
**Cost measured on the host, written down, not mitigated; the floor's double loss is now stated as intentional with its count; full-suite tail and the four baseline names are in the body. Follow-up `72d597dc` is comment-only (13 lines, `git diff --stat`), so the suite at `002c1a32` covers the code.**

**Bisection cost.** Production interpreter (`/opt/eeepc-agent/venv/bin/python` 3.11.2, i686), 120 real pre-cap records from the 2026-08-25 archive (median 101,579 bytes), both implementations on the same records:

```
one json.dumps + utf-8 encode of the record   p50  9.2 ms   max  32.7 ms
old cap (1,000-char clamp)                    p50 12.0 ms   p90  22.6   max  47.7 ms
new cap (bisection)                           p50 77.9 ms   p90 123.6   max 192.3 ms
added per truncated record                    p50 65.8 ms
```

~17 serializations, as predicted. Paid once per LLM call inside `record_llm_prompt`, on ~90% of executor calls, after a call that waits seconds on the gateway — order of 1–3% of call latency. Tens of milliseconds, so: recorded in the `_level_cut` docstring and in the body beside the cap's own justification, and the lever named (number of probes, not the property). Not applied, per your instruction to measure first.

**Floor.** Comment on `_CAP_MIN_KEEP_CHARS` now says the double loss (near-total levelling, then structural drop) is deliberate, fires only when the structure alone exceeds the budget, and happened 0 times on the 334 real over-cap records. Same sentence in the body.

**Framing.** Agreed and kept in the body as the smaller, truer claim: the median record is ~4× over, must lose ~75% under any perfect cap, and the PR fills the budget instead of leaving 92% of it empty — it does not restore the transcript.

Full Windows suite at `002c1a32`: `4 failed, 3114 passed, 17 skipped` in 22:50 — `test_bridge_cycle_tags.py::TestBridgeCycleTagIntegration::test_fail_open_tag_failure_never_breaks_a_green_cycle`, `test_bridge_locking.py::TestAcquireBridgeLock::test_acquires_when_free`, `::test_second_acquire_in_same_process_is_contended`, `::test_contended_lock_via_monkeypatched_flock`. Names, not count. Not merging my own PR.

